PR #20, opened 2023-09-07
Add sample method to data_pipeline

najielhachem:
Add sample data source operator that can sample from multiple data sources provided weights.

**Check list:**


tests/unit/data/data_pipeline/test_sample.py
```diff
@@ -0,0 +1,204 @@
...
+from fairseq2.data import DataPipeline, read_sequence
+from fairseq2.data.text.text_reader import read_text
+from fairseq2.utils.version import is_pt2_or_greater
+
+
+@pytest.mark.skipif(
+    not is_pt2_or_greater(),
+    reason="different sampling results with versions lower than PyTorch 2.0.0",
+)
+class TestSampleOp:
```
gwenzek, line 19: @cbalioglu what's up with test classes ?
pytest doesn't need those. And with one file per op like we do, filename + function name is enough to identify the test case.

(less indent == nicer black formatting)

src/fairseq2/data/data_pipeline.py
```diff
@@ -88,6 +88,18 @@ def round_robin(pipelines: Sequence["DataPipeline"]) -> "DataPipelineBuilder":
                 The data pipelines to round robin.
             """
 
+        @staticmethod
+        def sample(
+            pipelines: Sequence["DataPipeline"],
+            weights: Optional[Sequence[float]] = None,
+        ) -> "DataPipelineBuilder":
+            """Extract examples from ``pipelines`` by sampling based on ``weights``.
```
gwenzek, line 96: could you also document the behavior around the end of the pipeline ?
when is this pipeline considered exhausted ?

gwenzek (commented):
The code LGTM, but I wanted to talk about the use case and the handling of end of pipeline.

In this implementation, as soon as one child data source is done, the `sampled_data_source stops`.
In `round_robin_data_source` on the contrary I ensure that all data source have been exhausted  at least once before stopping.

* do we need the two logics ?
* if yes, I think they should be in the same data source with a `stop_on_shortest` flag
* if yes, I think we could merge `sampled_data_source` and `round_robin_data_source` by adding a `weights` parameter to `round_robin`

cbalioglu:
@gwenzek although they are technically similar in that they both iterate through various data pipelines in each iteration, I think semantically they are fundamentally two different operations. For sampling you never want to keep reading from a data pipeline once it is exhausted, doing that would defeat the whole purpose of sampling. So to me `stop_on_shortest` feels like a redundant/confusing parameter for sampling.

Maybe they can internally share some implementation logic, but I think they should be exposed as two separate operators.

tests/unit/data/data_pipeline/test_sample.py
cbalioglu: Old habits die hard, so used to having classes in `unittest`, and `googletest`, just adding them without giving thought. Feel free to remove them. I don't really have any strong preference.

fairseq2n/src/fairseq2n/data/sample_data_source.cc
```diff
@@ -0,0 +1,83 @@
...
+
+    return output;
+}
+
+void
+sample_data_source::reset()
+{
+    auto seed = generator_.current_seed();
+    generator_ = at::globalContext().defaultGenerator(c10::DeviceType::CPU);
+    generator_.set_current_seed(seed);
```
cbalioglu, line 46: Please don't do this. As a library we should never ever modify the default RNG. This can mess up the entire training flow. If the user wants repeatable reads from the pipeline, they should call `torch.manual_seed()`.

fairseq2n/src/fairseq2n/data/sample_data_source.cc
```diff
@@ -0,0 +1,83 @@
...
+    t.record(eod_);
+    for (const data_pipeline &p : pipelines_)
+        p.record_position(t);
+}
+
+void
+sample_data_source::reload_position(tape &t)
+{
+    auto state = t.read<at::Tensor>();
+    generator_.set_state(state);
```
cbalioglu, line 67: Again, please remove this. The checkpoint handling should be responsible for saving/restoring RNG states. It is not our job.

fairseq2n/src/fairseq2n/data/sample_data_source.cc
```diff
@@ -0,0 +1,83 @@
...
+
+    eod_ = t.read<bool>();
+    for (data_pipeline &p : pipelines_)
+        p.reload_position(t);
+}
+
+std::size_t
+sample_data_source::next_index()
+{
+    auto result = at::multinomial(weights_, 1, false, generator_)
```
cbalioglu, line 77: Nice, this reduced a lot of complexity from your previous revisions :)

fairseq2n/src/fairseq2n/utils/tensor.h
```diff
@@ -0,0 +1,31 @@
...
+#include <vector>
+
+#include <ATen/Tensor.h>
+
+namespace fairseq2n::detail {
+
+template <typename T>
+inline at::Tensor
+make_tensor_from_vector(
+    const std::vector<T>& src,
```
cbalioglu, line 19: Nit: in the rest of the code we put `&` next to the variable name.

fairseq2n/src/fairseq2n/data/sample_data_source.h
```diff
@@ -0,0 +1,51 @@
...
+#include "fairseq2n/data/data_source.h"
+
+namespace fairseq2n::detail {
+
+/// @brief sample from a list of datasources
+class sample_data_source final : public data_source {
+public:
+    explicit
+    sample_data_source(std::vector<data_pipeline> &&pipelines,
+     std::vector<float> &&weights);
```
cbalioglu, line 25: Nit: no need for a line break here. You can put this parameter next to `pipelines`.

fairseq2n/src/fairseq2n/data/data_pipeline.cc
```diff
@@ -224,6 +225,39 @@ data_pipeline::round_robin(std::vector<data_pipeline> pipelines)
...
+        });
+    if (is_broken)
+        throw_<std::invalid_argument>(
+            "At least one of the specified data pipelines is broken and cannot be used in sample.");
+
+    if (!weights)
+        weights = std::vector<float>(pipelines.size(), 1.0F / static_cast<float>(pipelines.size()));
+    else if (weights.value().size() != pipelines.size())
+        throw_<std::invalid_argument>(
+            "`weights` should have the same size as `pipelines`.");
```
cbalioglu, line 250: Nit: Can you also include their sizes in the error message. You can check out other op functions as an example. Note that `throw_<>()` handles message formatting, so it should be relatively easy.

fairseq2n/src/fairseq2n/data/data_pipeline.cc
```diff
@@ -224,6 +225,39 @@ data_pipeline::round_robin(std::vector<data_pipeline> pipelines)
     return data_pipeline_builder{std::move(factory)};
 }
 
+data_pipeline_builder
+data_pipeline::sample(
+    std::vector<data_pipeline> pipelines,
+    std::optional<std::vector<float>> weights)
```
cbalioglu, line 231: Nit: could you please use the `float32` alias here to be consistent with the rest of the code. It is defined in [float.h](https://github.com/facebookresearch/fairseq2/blob/main/fairseq2n/src/fairseq2n/float.h).

fairseq2n/src/fairseq2n/data/sample_data_source.cc
```diff
@@ -0,0 +1,83 @@
...
+#include "fairseq2n/utils/tensor.h"
+
+namespace fairseq2n::detail {
+
+sample_data_source::sample_data_source(std::vector<data_pipeline> &&pipelines, std::vector<float> &&weights)
+    : pipelines_(std::move(pipelines))
+{
+    auto pipelines_count = pipelines_.size();
+
+    weights_ = make_tensor_from_vector(weights, { static_cast<std::int64_t>(pipelines_count) });
```
cbalioglu, line 24: Nit: you can also pass `weights_.size()` and get rid of line 22.

cbalioglu (commented):
Thanks again for your work Naji! And sorry for the late review. I just left an important comment that should be easy to fix, otherwise looks good to me.

gwenzek:
> For sampling you never want to keep reading from a data pipeline once it is exhausted, doing that would defeat the whole purpose of sampling

imagine you're training a multilingual model. You have English, French and Zulu data. Volume is 100Gb of English, 10Gb of French, 1Gb of Zulu.
You want to up sample the Zulu and down sample English. So that you're model is trained on 50% English, 40% French and 10% Zulu. If you use `sample([0.5, 0.4, 0.1])` you're epoch will stop after having read the first  5Gb of English, which is only 5% of the original training set. If you then reset the dataloader and start a new epoch you'll restart reading the first 5Gb of English over and over.

fairseq2n/src/fairseq2n/data/sample_data_source.cc
gwenzek: can you explain this in bit more details, it's not really clear to me.
I thought that this would create a new generator just for this pipeline operator. Are you saying that `defaultGenerator` is just returning a pointer to a shared RNG ? Should we use a shared RNG, one for the op, or one for the full pipeline ?

cbalioglu:
> You want to up sample the Zulu and down sample English. So that you're model is trained on 50% English, 40% French and 10% Zulu. If you use sample([0.5, 0.4, 0.1]) you're epoch will stop after having read the first 5Gb of English, which is only 5% of the original training set. If you then reset the dataloader and start a new epoch you'll restart reading the first 5Gb of English over and over.

Yeah that makes sense. I haven't been focusing on up/down sampling really, but when you consider that, `stop_on_shortest` sounds reasonable. Thanks for the example.

I still think that round-robin and sampling should be two separate ops though. Round-robin by its nature does not really have any randomness, it is all about iterating each pipeline one-by-one. It feels weird to me to use an op called `round_robin()` for sampling or the other way round.

fairseq2n/src/fairseq2n/data/sample_data_source.cc
cbalioglu: Yes, this just returns a handle to the shared (i.e. default) CPU RNG of the entire process. We definitely should not modify it as that will likely mess up other parts of users' code. In fact, I am not even entirely convinced that we should have a "seed" or RNG in the data pipeline logic at all. The typical use case is for users is to save/restore default RNGs (for CPU and CUDA) as part of checkpoint logic and that should be sufficient to restore randomness in data pipeline ops as well. I might be overseeing something though, let me know if you can think of a specific case where an explicit seed/RNG would be needed.

tests/unit/data/data_pipeline/test_sample.py
najielhachem: I am closing this since it's not related particularly to this PR.

fairseq2n/src/fairseq2n/data/sample_data_source.cc
najielhachem: In that case I agree that we should not have a seed in our pipeline. We can let the `at::Generator` handle it when the user calls `torch.manual_seed`

najielhachem:
I agree with @cbalioglu about having two operators since semantically we don't expect a similar behaviour from round_robin and sample. But as mentioned by @gwenzek, the buffering seems useful so I will extract this part into a separate code and use it in both of our operators.

fairseq2n/src/fairseq2n/data/sample_data_source.cc
```diff
@@ -0,0 +1,79 @@
...
+
+    eod_ = false;
+    for (data_pipeline &p : pipelines_)
+        p.reset();
+}
+
+void
+sample_data_source::record_position(tape &t) const
+{
+    t.record(generator_.get_state());
```
cbalioglu, line 52: You don't have to record the state of RNG. That is again the job of the user.

fairseq2n/src/fairseq2n/data/sample_data_source.cc
```diff
@@ -0,0 +1,79 @@
...
+    t.record(eod_);
+    for (const data_pipeline &p : pipelines_)
+        p.record_position(t);
+}
+
+void
+sample_data_source::reload_position(tape &t)
+{
+    auto state = t.read<at::Tensor>();
+    generator_.set_state(state);
```
cbalioglu, line 63: Again, please remove this line. The state handling for RNG should be the responsibility of the user. Just retrieve a handle to the generator in the constructor and use it in `multinomial` call.

fairseq2n/src/fairseq2n/data/sample_data_source.cc
```diff
@@ -0,0 +1,79 @@
...
+    if (!output)
+        eod_ = true;
+
+    return output;
+}
+
+void
+sample_data_source::reset()
+{
+    generator_ = at::globalContext().defaultGenerator(c10::DeviceType::CPU);
```
cbalioglu, line 42: `defaultGenerator` won't change throughout the lifetime of the process. You can move this call to constructor.

najielhachem: True, I missed that. All generator state handling will be done outside of this operator

cbalioglu (approved):
LGTM!

PR #54, opened 2023-09-18
Expose DynamicLossScaler

cbalioglu:
A nit PR that exports `DynamicLossScaler` in the `fairseq2.optim` package.

### DIFF
--- a/src/fairseq2/optim/__init__.py
+++ b/src/fairseq2/optim/__init__.py
@@ -5,4 +5,5 @@
 # LICENSE file in the root directory of this source tree.
 
 from fairseq2.optim.adamw import AdamW as AdamW
+from fairseq2.optim.dynamic_loss_scaler import DynamicLossScaler as DynamicLossScaler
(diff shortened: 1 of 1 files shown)


gwenzek:
what's the difference with Pythorch `GradScaler` ?

https://pytorch.org/docs/stable/notes/amp_examples.html#gradient-accumulation
https://pytorch.org/docs/stable/amp.html#gradient-scaling

cbalioglu:
In fact internally it uses GradScaler or DistributedGradScaler depending on the provided arguments. It just makes them much more usable for real world use cases like providing an additional `min_scale` as a threshold so that you can figure out whether your loss is exploding, various log statements so that you can track your scale, also an relatively easier API and couple fixes for bugs in DistributedGradScaler.

You can check out the implementation [here](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/optim/dynamic_loss_scaler.py). It is relatively short.

PR #67, opened 2023-09-26
add `.collate` for `.map(Collater)`

gwenzek:
As disscussed in #64 I propose to add a shortcut `.collate` to replace `.map(Collater)`.
This simplifies the API for the user, because they don't have to learn about the `Collater` class, and it's more discoverable because the `.collate` function will be documented next to the other data pipeline ops.

### DIFF
--- a/fairseq2n/python/src/fairseq2n/bindings/data/data_pipeline.cc
+++ b/fairseq2n/python/src/fairseq2n/bindings/data/data_pipeline.cc
@@ -404,6 +404,33 @@ def_data_pipeline(py::module_ &data_module)
             py::arg("selector") = std::nullopt,
             py::arg("drop_remainder") = false)
         .def(
+            "collate",
+            [](
+                data_pipeline_builder &self,
+                std::optional<std::int64_t> maybe_pad_idx,
+                std::int64_t pad_to_multiple,
+                std::optional<std::vector<collate_options_override>> maybe_opt_overrides,
+                std::size_t num_parallel_calls) -> data_pipeline_builder &
+            {
+                auto opts = collate_options()
+                    .maybe_pad_idx(maybe_pad_idx).pad_to_multiple(pad_to_multiple);
+
+                std::vector<collate_options_override> opt_overrides{};
+                if (maybe_opt_overrides)
+                    opt_overrides = *std::move(maybe_opt_overrides);
+
+                map_fn f = collater(opts, std::move(opt_overrides));
+                element_mapper mapper{f, std::nullopt};
+
+                self = std::move(self).map(std::move(mapper), num_parallel_calls);
+
+                return self;
+            },
+            py::arg("pad_idx") = std::nullopt,
+            py::arg("pad_to_multiple") = 1,
+            py::arg("overrides") = std::nullopt,
+            py::arg("num_parallel_calls") = 1)
+        .def(
             "filter",
             [](data_pipeline_builder &self, predicate_fn fn) -> data_pipeline_builder &
             {
--- a/fairseq2n/python/src/fairseq2n/bindings/data/data_pipeline.cc
+++ b/fairseq2n/python/src/fairseq2n/bindings/data/data_pipeline.cc
@@ -420,9 +420,8 @@ def_data_pipeline(py::module_ &data_module)
                     opt_overrides = *std::move(maybe_opt_overrides);
 
                 map_fn f = collater(opts, std::move(opt_overrides));
-                element_mapper mapper{f, std::nullopt};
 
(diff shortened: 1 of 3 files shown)


fairseq2n/python/src/fairseq2n/bindings/data/data_pipeline.cc
```diff
@@ -378,6 +378,33 @@ def_data_pipeline(py::module_ &data_module)
...
+            {
+                auto opts = collate_options()
+                    .maybe_pad_idx(maybe_pad_idx).pad_to_multiple(pad_to_multiple);
+
+                std::vector<collate_options_override> opt_overrides{};
+                if (maybe_opt_overrides)
+                    opt_overrides = *std::move(maybe_opt_overrides);
+
+                map_fn f = collater(opts, std::move(opt_overrides));
+                element_mapper mapper{f, std::nullopt};
```
cbalioglu, line 398: Do we need an `element_mapper` here? Since we are explicitly passing `nullopt`, it feels superfluous. It should be possible to pass `f` directly to `map()`.

src/fairseq2/data/data_pipeline.py
```diff
@@ -115,6 +130,17 @@ def bucket_by_length(
...
 
+        def collate(
+            self,
+            pad_idx: Optional[int] = None,
+            pad_to_multiple: int = 1,
+            overrides: Optional[Sequence["CollateOptionsOverride"]] = None,
+        ) -> Self:
+            """Concatenate a list of inputs into a single inputs.
+
+            See :py:class:`fairseq2.data.Collater` for details.
```
cbalioglu, line 141: Nit: Maybe also note that this is a convenience function over `map` + `Collater`?

cbalioglu (commented):
Looks good to me. Just left two minor comments. I think you should rebase it though. It seems you also have the doc changes included in this PR.

src/fairseq2/data/data_pipeline.py
```diff
@@ -27,9 +28,21 @@
...
 if TYPE_CHECKING or _DOC_MODE:
 
-    class DataPipeline:
+    class DataPipeline(Iterable[Any]):
+        """fairseq2 native data pipeline.
+
+        The pipeline state can be persisted to the disk, allowing it to be resumed later.
+        It is a Python Iterable, but it also contains the iterator states.
+        Calling `iter` a second time while the first iterator is still being used
+        will segfault or worse.
```
najielhachem, line 37: the `or worse` sounds a bit ambiguous to put in a doc. Can you elaborate ?

gwenzek: Initially I just wrote "will cause Undefined Beahvior", but I find that a bit unclear. "or worse" here means data corruption if you have two threads writing and reading from shared buffer concurrently.

cbalioglu: I am not sure when exactly you observe this kind of behavior. Considering that the iterator is always called under GIL, from data pipeline's point of view, they are just a sequence of next() calls. It shouldn't cause any race condition or segfault. If you have a use case that we can reproduce, please share it. It should be treated as bug and we should fix it.

gwenzek: I wasn't able to reproduce segfault anymore, but it is still a footgun to call `__iter__` twice on the same object.
See eg the following test:

```py
def test_two_iterators_interfere_with_each_others() -> None:
    dataloader = (
        fairseq2.data.text.read_text(FILE, rtrim=True)
        .map(lambda line: torch.tensor([c for c in bytes(line)]))
        .bucket_by_length([(10, 10), (5, 20), (1, 100)])
        .prefetch(5)
        .and_return()
    )
    it1 = iter(dataloader)
    it2 = iter(dataloader)
    l1 = list(itertools.islice(it1, 10))
    l2 = list(itertools.islice(it2, 10))

    # it1 and it2 are reading from the same dataloader,
    # so they interfere with each others.
    assert l1 != FILE_LINES[:10]
    assert l2 != FILE_LINES[:10]
    assert l1 != l2

```

fairseq2n/python/src/fairseq2n/bindings/data/data_pipeline.cc
gwenzek: done

src/fairseq2/data/data_pipeline.py
gwenzek: done

gwenzek:
can we merge this ? I've addressed all the feedback.

cbalioglu (approved):

PR #76, opened 2023-10-02
Introduce stop_at_shortest in sample and round_robin

najielhachem:

Following this [PR](https://github.com/facebookresearch/fairseq2/pull/20) review we decided to do the following:

* Extract common parts of `sample` and `round_robin` into separate class
* Introduce circular logic for `sample` operator useful for up_sampling
* Add `stop_at_shortest` flag to both operator to give user more flexibility on how the operator works.

**Check list:**


tests/unit/data/data_pipeline/test_sample.py
```diff
@@ -115,6 +119,19 @@ def test_op_works_when_seed_is_set_manually(self) -> None:
...
+    def test_op_works_when_up_sampling(self) -> None:
+        dp1 = read_sequence([1, 2, 3, 4, 5]).and_return()
+        dp2 = read_sequence([11, 12]).and_return()
+
+        rdp = DataPipeline.sample(
+            [dp1, dp2], [0.5, 0.5], stop_at_shortest=False
+        ).and_return()
+        for _ in range(2):
+            with tmp_rng_seed(cpu_device, seed=1234):
+                assert list(rdp) == [11, 1, 12, 2, 3, 11, 4, 12, 11, 12, 5]
```
gwenzek, line 131: great

tests/unit/data/data_pipeline/test_round_robin.py
```diff
@@ -76,6 +77,28 @@ def test_op_works_when_pipelines_have_different_lengths(self) -> None:
 
             pipeline.reset()
 
+    @pytest.mark.skipif(
+        python_devel_only(),
+        reason="New fairseq2n API in Python-only installation. Skipping till v0.2.",
```
gwenzek, line 82: Not for this PR: 

I'm not fond of the behavior of this `python devel only`
First it seems to be about not disturbing Python only developer that didn't build fairseq2n locally. So `fairseq2n_dev_only` would be more clear.
Second I'd prefer if it was annotation `@fairseq2n_dev_only` with  a standard reason.
Third when are we supposed to removed those annotations ? Will we have to do this manually ? 

I don't really see who would want to run our unit tests from github main branch but can't build fairseq2n.

gwenzek (approved):
LGTM.

If I had to nit, I'd says that I feed it weird to have that little logic spread out in 6 files.
sampled_data_source and round_robin_data_source, may not need to be their own class anymore.
But it's also consistent with the rest of the codebase, where we have two files per operator.

tests/unit/data/data_pipeline/test_round_robin.py
cbalioglu: > First it seems to be about not disturbing Python only developer that didn't build fairseq2n locally. So fairseq2n_dev_only would be more clear.

I think if you convert it to a standalone annotation, a name like `@fairseq2n_dev_only` makes sense, but as a parameter to `skip_if`, `fairseq2n_dev_only` does not sound right, because it is just the opposite: you want to skip the test *if* this is a python-only installation. I am open to having a standalone annotation though.

> Third when are we supposed to removed those annotations ? Will we have to do this manually ?

There are two ways to resolve this issue. We can either have a version range check, like `@skipif(fairseq2n.__version__< 0.2)` which would not require us to remove the annotation in the future. The disadvantage is, those checks would not be relevant after a stable release and will start to pile up with time for no good reason. The other alternative is what we have right now. Just checking whether the versions do not match `@skipif(fairseq2.__version__ != fairseq2n.__version__)`. The advantage is that this forces you to remove the annotations before a stable release (otherwise the tests won't run), the disadvantage is that it is a manual work that needs to be done as part of the version bump. Having said that I think we won't need this annotation that often in the future once the native library becomes more mature. The manual removal is not ideal, but we might add some extra measures like forcibly fail any test that has this annotation and if the latest version is already past the one in the annotation.

> Can someone explain me the use case ?

Pretty much anyone who wants to contribute to the Python code base of fairseq2, but does not want to deal (or have experience) with C++. The whole reason why we ended up with fairseq2n in last minute was because people had quite a bit of trouble with building the library (e.g. while working on seq-gen and UnitY upstreaming). The original feedback was to completely separate the C++ parts to a separate repo, but we then decided to keep it here and offer pre-built binaries so that people who want to contribute to fairseq2 won't have to build the native parts if they don't want to. By the way, I got this approach pretty much from jax where they similarly separate the library into two parts: jax (python only) and jaxlib (pre-built).

cbalioglu: Ok, just gave it a try in #83. Let me know what you think.

gwenzek: thanks, let's follow on that PR, and merge this one.

fairseq2n/src/fairseq2n/data/data_pipeline.h
```diff
@@ -108,6 +111,8 @@ using map_fn = std::function<data(data &&)>;
 
 using predicate_fn = std::function<bool(const data &)>;
 
+using index_generator_fn = std::function<std::size_t()>;
```
cbalioglu, line 114: Ideally this alias should reside in `multi_data_source.h`. It is not used by any of the public data pipeline ops and is mostly an implementation detail used by `sample` and `round_robin`.

fairseq2n/src/fairseq2n/data/multi_data_source.cc
```diff
@@ -0,0 +1,107 @@
...
+// All rights reserved.
+//
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+
+#include "fairseq2n/data/multi_data_source.h"
+
+namespace fairseq2n::detail {
+
+multi_data_source::multi_data_source(std::vector<data_pipeline> &&pipelines, index_generator_fn &&index_gen_fn, bool stop_at_shortest)
```
cbalioglu, line 11: Nit: I haven't counted it, but if this line exceeds ~100 chars, please break it.

fairseq2n/src/fairseq2n/data/round_robin_data_source.cc
```diff
@@ -8,110 +8,53 @@
...
-                return b;
-            });
+    pipelines_count_ = pipelines.size();
+    pipeline_idx_ = 0;
 
-        if (all_done) {
-            is_eod_ = true;
+    auto gen = [this]()
+    {
+        this->pipeline_idx_ %= pipelines_count_;
```
cbalioglu, line 18: Nit: this code should compile without `this->` if I am not wrong.

tests/unit/data/data_pipeline/test_round_robin.py
```diff
@@ -76,6 +77,28 @@ def test_op_works_when_pipelines_have_different_lengths(self) -> None:
...
+    )
+    def test_op_works_when_pipelines_have_different_lengths_stop_on_shortest(
```
cbalioglu, line 84: Should we name this parameter `stop_on_shortest` or `stop_at_shortest`?

fairseq2n/src/fairseq2n/data/sample_data_source.cc
```diff
@@ -10,63 +10,52 @@
...
-    : pipelines_(std::move(pipelines))
+sample_data_source::sample_data_source(std::vector<data_pipeline> &&pipelines, std::vector<float32> &&weights, bool stop_at_shortest)
 {
-    weights_ = make_tensor_from_vector(weights, { static_cast<std::int64_t>(pipelines_.size()) });
+    weights_ = make_tensor_from_vector(weights, { static_cast<std::int64_t>(pipelines.size()) });
     generator_ = at::globalContext().defaultGenerator(c10::DeviceType::CPU);
+
+    auto gen = [this]()
+    {
+        auto result = at::multinomial(this->weights_, 1, false, this->generator_)
```
cbalioglu, line 24: Nit: again, not sure whether you need `this->` here.

fairseq2n/src/fairseq2n/data/multi_data_source.h
```diff
@@ -0,0 +1,48 @@
...
+
+#pragma once
+
+#include <vector>
+
+#include "fairseq2n/data/data_pipeline.h"
+
+namespace fairseq2n::detail {
+
+class multi_data_source final : public data_source {
```
cbalioglu, line 15: It is subjective, but I think `composite_data_source` would sound more accurate. "Composition" has a more clear meaning in software engineering. `multi` sounds a bit ambiguous to me.

fairseq2n/src/fairseq2n/data/multi_data_source.cc
```diff
@@ -0,0 +1,107 @@
...
+
+    // One or more data pipelines might be empty, so we have to keep looping
+    std::optional<data> output{};
+    while (!output && !eod()) {
+        auto pipeline_idx = next_index_gen_();
+
+        if (!buffer_[pipeline_idx]) // init buffer at index
+            buffer_[pipeline_idx] = next_in_pipeline(pipeline_idx);
+
+        output = buffer_[pipeline_idx];
```
cbalioglu, line 32: You should ideally use move semantics here; otherwise, this will cause a copy for composite objects like lists and maps: 

```
output = std::exchange(buffer_[pipeline_idx], next_in_pipeline(pipeline_idx))
```

fairseq2n/src/fairseq2n/data/multi_data_source.cc
```diff
@@ -0,0 +1,107 @@
...
+{
+    if (stop_at_shortest_) // with this flag on, the operator is a simple iterator
+        return pipelines_[next_index_gen_()].next();
+
+    // One or more data pipelines might be empty, so we have to keep looping
+    std::optional<data> output{};
+    while (!output && !eod()) {
+        auto pipeline_idx = next_index_gen_();
+
+        if (!buffer_[pipeline_idx]) // init buffer at index
```
cbalioglu, line 29: You can reduce the repetitive `buffer_[pipeline_idx]` expressions:

```cpp
// Note that this is a reference to the element.
std::optional<data> &maybe_example = buffer_[pipeline_idx];

// Now use `maybe_example` instead.
```

fairseq2n/src/fairseq2n/data/multi_data_source.cc
```diff
@@ -0,0 +1,107 @@
...
+
+std::optional<data>
+multi_data_source::next()
+{
+    if (stop_at_shortest_) // with this flag on, the operator is a simple iterator
+        return pipelines_[next_index_gen_()].next();
+
+    // One or more data pipelines might be empty, so we have to keep looping
+    std::optional<data> output{};
+    while (!output && !eod()) {
```
cbalioglu, line 26: Neat! This is definitely a much cleaner implementation than mine in `round_robin`.

cbalioglu (commented):
Solid work! Just left a few nit comments. There is only one line where we should leverage move-semantics for performance, but otherwise most of the my comments are optional suggestions. Let me know once you at least address that performance related issue, and I can stamp it right away.

fairseq2n/src/fairseq2n/data/multi_data_source.cc
najielhachem: nice catch, I didn't think about that.

najielhachem:
@gwenzek I agree that there is too many files now but we must have a class for round_robin and sample since the save/reload and reset have different logics.

najielhachem:
@cbalioglu Thanks for the detailed review! I did all the changes requested and the PR should be ready by now.

src/fairseq2/data/data_pipeline.py
```diff
@@ -94,24 +94,36 @@ def zip(
...
+        def round_robin(
+            pipelines: Sequence["DataPipeline"],
+            stop_at_shortest: bool = False,
+        ) -> "DataPipelineBuilder":
             """Extract examples from ``pipelines`` in round robin.
 
             :param pipelines:
                 The data pipelines to round robin.
+            :param stop_at_shortest:
+                Flag to stop round_robin when first pipeline reaches its end.
```
cbalioglu, line 106: Nit: As a convention for flag parameters, I use the template "If ``True``, blab bla". In this case, that would look like: "If ``True``, stops round robin when first pipeline reaches its end."

src/fairseq2/data/data_pipeline.py
```diff
@@ -94,24 +94,36 @@ def zip(
...
+            stop_at_shortest: bool = False,
         ) -> "DataPipelineBuilder":
             """Extract examples from ``pipelines`` by sampling based on ``weights``.
 
             :param data_pipelines:
                 The data pipelines to sample from.
             :param weights:
                 Desired distribution of pipelines. If None, use uniform distribution.
+            :param stop_at_shortest:
+                Flag to stop sampling when first pipeline reaches its end.
```
cbalioglu, line 124: Same as above.

fairseq2n/src/fairseq2n/data/composite_data_source.cc
```diff
@@ -0,0 +1,114 @@
...
+#include "fairseq2n/data/composite_data_source.h"
+
+namespace fairseq2n::detail {
+
+composite_data_source::composite_data_source(
+    std::vector<data_pipeline> &&pipelines,
+    index_generator_fn &&index_gen_fn,
+    bool stop_at_shortest)
+  : pipelines_(std::move(pipelines)),
+    next_index_gen_(std::move(index_gen_fn)),
```
cbalioglu, line 16: Nit: for this assignment and the next one for `stop_at_shortest_`, ideally use brace initialization (i.e. `next_index_gen_{std::move(..)}, stop_at_shorted_{stop_at_shortest}` to follow the convention in the rest of the code base.

cbalioglu (approved):
I left two very nit comments. Thanks again for this work! The implementation has become much cleaner! Feel free to merge it whenever you want.

cbalioglu (approved):

PR #83, opened 2023-10-04
Introduce fairseq2n pytest marker

cbalioglu:
Based on the discussion in #76, this PR introduces a pytest marker named `fairseq2n` to indicate which versions of fairseq2n are compatible with a marked test.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,21 +4,16 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import warnings
 from argparse import ArgumentTypeError
 from pathlib import Path
 from typing import cast
 
 import pytest
+from packaging.version import Version
 
 import tests.common
 from fairseq2.typing import Device
-
-
-def parse_device_arg(value: str) -> Device:
-    try:
-        return Device(value)
-    except RuntimeError:
-        raise ArgumentTypeError(f"'{value}' is not a valid device name.")
 
 
 def pytest_addoption(parser: pytest.Parser) -> None:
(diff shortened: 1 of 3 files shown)


tests/unit/data/data_pipeline/test_sample.py
```diff
@@ -10,15 +10,12 @@
...
+from tests.common import tmp_rng_seed
 
 cpu_device = torch.device("cpu")
 
 
-@pytest.mark.skipif(
-    python_devel_only(),
-    reason="New fairseq2n API in Python-only installation. Skipping till v0.2.",
-)
+@pytest.mark.fairseq2n("0.2a0")
```
gwenzek, line 18: do we need to have a custom pytest marks? 
why not using `skip_if_fairseq2n_newer` as an annotation ?

```
def skip_if_fairseq2n_newer(version):
  ...
  if ...:
     warnings.warn(...)
  return pytest.skip(installed_version < annotated_version, reason=f"The test requires fairseq2n v{annotated_version} or greater.")

```

tests/conftest.py
```diff
@@ -46,3 +55,28 @@ def pytest_ignore_collect(
         return not cast(bool, config.getoption("integration"))
 
     return False
+
+
+def pytest_runtest_setup(item: pytest.Function) -> None:
+    marker = item.get_closest_marker(name="fairseq2n")
+    if marker is not None:
+        skip_if_fairseq2n_newer(*marker.args, **marker.kwargs)
```
gwenzek, line 63: if we end up using custom marks can we have a link to documentation ? https://docs.pytest.org/en/stable/reference/reference.html#custom-marks

tests/unit/data/data_pipeline/test_sample.py
cbalioglu: TBH no particular reason. I just followed the instructions here: https://docs.pytest.org/en/7.1.x/example/markers.html. Looks like one advantage of using a pytest marker is that it integrates your marker with the pytest CLI (described [here](https://docs.pytest.org/en/7.1.x/example/markers.html)). It also makes it possible to use a marker in parametrized tests. Let me know if you have any objection on using a pytest marker though. I don't have a strong opinion on this.

tests/conftest.py
cbalioglu: Yep. I just realized that I should also update the contribution guidelines and mention this annotation there.

cbalioglu:
Not needed anymore after #124

PR #84, opened 2023-10-04
Concatenate method for DataPipeline class

am831:
Implements ```cat``` method in the ```DataPipeline``` class which concatenates two vectors of data pipelines. Another solution would be to take one vector of data pipelines and concatenate all pipelines in the vector. 

**Check list:**


fairseq2n/python/src/fairseq2n/bindings/data/data_pipeline.cc
```diff
@@ -358,8 +358,35 @@ def_data_pipeline(py::module_ &data_module)
                 return data_pipeline::count(start, std::move(key));
             },
             py::arg("start") = 0,
-            py::arg("key") = std::nullopt);
+            py::arg("key") = std::nullopt)
+        .def_static(
+            "cat",
```
cbalioglu, line 363: I would prefer to name this op `concat` to make it a bit more descriptive.

fairseq2n/python/src/fairseq2n/bindings/data/data_pipeline.cc
```diff
@@ -358,8 +358,35 @@ def_data_pipeline(py::module_ &data_module)
...
+            [](
+                std::vector<std::reference_wrapper<data_pipeline>> &refs1,
```
cbalioglu, line 365: Is there a specific reason why we take two lists of data pipelines (i.e. `refs1` and `refs2`) instead of just one?

cbalioglu (commented):
Thanks for your first PR! My main feedback/question is why we need to have two lists of pipelines. I think the ideal API should look like `.concat(pipelines: Sequence[DataPipeline])` (in C++ `.concat(std::vector<data_pipeline> pipelines)`, so I can't see the reason for the second "pipelines2".

Looks like there are also some linting and build issues that you can check at the bottom of the PR conversation pane.

fairseq2n/python/src/fairseq2n/bindings/data/data_pipeline.cc
am831: Ok I will change it

fairseq2n/python/src/fairseq2n/bindings/data/data_pipeline.cc
am831: I was unsure how it should be implemented, but I will change it to just one

fairseq2n/src/fairseq2n/data/cat_data_source.cc
```diff
@@ -0,0 +1,76 @@
...
+
+void cat_data_source::reload_position(tape &t)
+{
+    for (auto &pipeline : pipeline1_)
+        pipeline.reload_position(t);
+    for (auto &pipeline : pipeline2_)
+        pipeline.reload_position(t);
+}
+
+std::vector<data_pipeline> cat_data_source::concatenate(
```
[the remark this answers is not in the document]
am831: After I change the function to ```.concat(std::vector<data_pipeline> pipelines)``` this is where concatenating pipelines will be implemented. Should it be implemented elsewhere?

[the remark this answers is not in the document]
am831: We discussed in our meeting that we just need next(), I will remove this, thanks

fairseq2n/src/fairseq2n/data/concat_data_source.cc
```diff
@@ -0,0 +1,49 @@
...
+//
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+
+#include "fairseq2n/data/concat_data_source.h"
+#include <vector>
+
+namespace fairseq2n::detail {
+
+concat_data_source::concat_data_source(
```
cbalioglu, line 12: Nit: In C++ files, we typically break lines at 100 chars. I haven't counted, but looks like the `pipelines` parameter can be on the same line here.

fairseq2n/src/fairseq2n/data/concat_data_source.cc
```diff
@@ -0,0 +1,49 @@
...
+    : pipelines_{std::move(pipelines)}
+{}
+
+std::optional<data>
+concat_data_source::next()
+{
+    std::optional<data> d;
+    for (auto &p : pipelines_) {
+        d = p.next();
+        if (d) {
```
cbalioglu, line 23: Nit: for single line statement blocks, we do not use opening/closing braces. You can write this as:

```cpp
if (d)
  return d;
```

fairseq2n/src/fairseq2n/data/data_pipeline.cc
```diff
@@ -283,6 +285,34 @@ data_pipeline::count(std::int64_t start, std::optional<std::string> key)
...
+
+    bool is_broken = std::any_of(
+        pipelines.begin(), pipelines.end(), [](const data_pipeline &pipeline)
+        {
+            return pipeline.is_broken();
+        });
+
+    if (is_broken)
+        throw_<std::invalid_argument>(
+            "At least one of the specified data pipelines is broken and cannot be used in concat.");
```
cbalioglu, line 304: Nit: I would end this sentence with "cannot be concatenated."

fairseq2n/src/fairseq2n/data/concat_data_source.cc
```diff
@@ -0,0 +1,49 @@
...
+// LICENSE file in the root directory of this source tree.
+
+#include "fairseq2n/data/concat_data_source.h"
+#include <vector>
+
+namespace fairseq2n::detail {
+
+concat_data_source::concat_data_source(
+    std::vector<data_pipeline> &&pipelines)
+    : pipelines_{std::move(pipelines)}
```
cbalioglu, line 14: There is a subtle bug here. This is one of the complicated features of C++, but here you should use parentheses instead of brace initialization. Otherwise, the compiler will pick the initializer_list constructor instead of move constructor.

```cpp
: pipelines_(std::move(pipelines))
```

fairseq2n/src/fairseq2n/data/data_pipeline.cc
am831: Fixed

fairseq2n/src/fairseq2n/data/concat_data_source.cc
am831: Fixed

fairseq2n/src/fairseq2n/data/concat_data_source.cc
am831: Fixed

cbalioglu (approved):
LGTM!

cbalioglu (approved):

PR #88, opened 2023-10-05
Introduce LoRA layers and wrappers

chaoweihuang:
This PR introduces [Low-Rank Adaptation (LoRA)](https://arxiv.org/abs/2106.09685) modules as well as wrappers for enabling/disabling models with LoRA. This is a joint contribution with @light1726 .

**Notes**
- Currently only `Projection` and `Embedding` are supported. `Conv` layers are not supported yet.
- Currently `load_lora` is not implemented.
- For usage examples, please check `tests/unit/nn/test_lora.py` and `tests/integration/models/test_llama_lora.py`.

None

**Check list:**

### DIFF
--- a/src/fairseq2/models/llama/builder.py
+++ b/src/fairseq2/models/llama/builder.py
@@ -365,5 +365,5 @@ def get_llama_lora_config() -> LoRAConfig:
         r=8,
         alpha=16.0,
         dropout_p=0.05,
-        keys=[".*decoder.layers.*.self_attn.*(q_proj|v_proj)$"]
+        keys=[".*decoder.layers.*.self_attn.*(q_proj|v_proj)$"],
     )
--- a/src/fairseq2/models/llama/builder.py
+++ b/src/fairseq2/models/llama/builder.py
@@ -15,6 +15,7 @@ from fairseq2.models.transformer import (
 )
 from fairseq2.models.utils.arch_registry import ArchitectureRegistry
 from fairseq2.nn.embedding import Embedding
+from fairseq2.nn.lora import LoRAConfig
 from fairseq2.nn.normalization import LayerNorm, RMSNorm
 from fairseq2.nn.position_encoder import RotaryEncoder
 from fairseq2.nn.transformer import (
@@ -357,3 +358,12 @@ def create_llama_model(
         The data type of module parameters and buffers.
     """
     return LLaMABuilder(config, device=device, dtype=dtype).build_model()
+
+
+def get_llama_lora_config() -> LoRAConfig:
+    return LoRAConfig(
+        r=8,
+        alpha=16.0,
+        dropout_p=0.05,
+        keys=[".*decoder.layers.*.self_attn.*(q_proj|v_proj)$"]
+    )
--- a/src/fairseq2/nn/lora.py
+++ b/src/fairseq2/nn/lora.py
@@ -52,6 +52,7 @@ class LoRALayer(ABC):
 
 @final
 class LoRAEmbedding(Embedding, LoRALayer):
+    wrapped: Embedding
     weight: Parameter
     lora_A: Parameter
     lora_B: Parameter
@@ -78,9 +79,11 @@ class LoRAEmbedding(Embedding, LoRALayer):
         )
         LoRALayer.__init__(self, config)
 
-        self.weight = Parameter(
-            wrapped.weight, requires_grad=False  # type: ignore[arg-type]
-        )
+        self.wrapped = wrapped
+
+        self.wrapped.weight.requires_grad_(False)
+
+        self.weight = self.wrapped.weight  # type: ignore[assignment]
 
         self.lora_A = Parameter(
             torch.empty((self.r, self.num_embeddings), device=device, dtype=dtype)
@@ -110,26 +113,31 @@ class LoRAEmbedding(Embedding, LoRALayer):
         nn.init.normal_(self.lora_B)
 
     @property
-    def wrapped_module(self) -> nn.Module:
-        return self
+    def wrapped_module(self) -> Embedding:
+        return self.wrapped
 
     def merge(self) -> None:
         if self.merged:
             return
 
-        self.weight.data += (self.lora_B.data @ self.lora_A.data).T * self.scaling
+        with torch.no_grad():
+            self.weight += (self.lora_B @ self.lora_A).T * self.scaling  # type: ignore[misc]
+
         self.merged = True
 
     def unmerge(self) -> None:
         if not self.merged:
             return
 
-        self.weight.data -= (self.lora_B.data @ self.lora_A.data).T * self.scaling
+        with torch.no_grad():
+            self.weight -= (self.lora_B @ self.lora_A).T * self.scaling  # type: ignore[misc]
+
         self.merged = False
 
 
 @final
 class LoRALinear(Projection, LoRALayer):
+    wrapped: Projection
     weight: Parameter
     bias: Optional[Parameter]
     lora_A: Parameter
@@ -156,17 +164,17 @@ class LoRALinear(Projection, LoRALayer):
             :meth:`reset_lora_parameters` will become noop.
         """
         Projection.__init__(self, wrapped.input_dim, wrapped.output_dim)
-
         LoRALayer.__init__(self, config)
 
-        self.weight = Parameter(
-            wrapped.weight, requires_grad=False  # type: ignore[arg-type]
-        )
+        self.wrapped = wrapped
 
-        if wrapped.bias is not None:
-            self.bias = Parameter(
-                wrapped.bias, requires_grad=False  # type: ignore[arg-type]
-            )
+        self.wrapped.weight.requires_grad_(False)
+
+        self.weight = self.wrapped.weight  # type: ignore[assignment]
+
+        if self.wrapped.bias is not None:
+            self.wrapped.bias.requires_grad_(False)
+            self.bias = self.wrapped.bias  # type: ignore[assignment]
         else:
             self.register_module("bias", None)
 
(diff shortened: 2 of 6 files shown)


src/fairseq2/nn/lora.py
```diff
@@ -0,0 +1,373 @@
...
+
+    @property
+    def wrapped_module(self) -> nn.Module:
+        return self
+
+    def merge(self) -> None:
+        if self.merged:
+            return
+
+        self.weight.data += (self.lora_B.data @ self.lora_A.data).T * self.scaling
```
cbalioglu, line 120: Please use just `self.weight +=` and refrain from using the `data` attribute, which exists mostly for legacy reasons.

cbalioglu, line 120: Should this not be performed under `torch.no_grad()`?

src/fairseq2/nn/lora.py
```diff
@@ -0,0 +1,373 @@
...
+        self.merged = True
+
+    def unmerge(self) -> None:
+        if not self.merged:
+            return
+
+        self.weight.data -= (self.lora_B.data @ self.lora_A.data).T * self.scaling
```
cbalioglu, line 127: Same comments as above.

src/fairseq2/nn/lora.py
```diff
@@ -0,0 +1,373 @@
...
+            )
+
+    def reset_lora_parameters(self) -> None:
+        """Reset the parameters and buffers of the module."""
+        nn.init.zeros_(self.lora_A)
+        nn.init.normal_(self.lora_B)
+
+    @property
+    def wrapped_module(self) -> nn.Module:
+        return self
```
cbalioglu, line 114: You should ideally store the `wrapped` argument passed to the constructor as a state variable and return it here.

```python
class LoRAEmbedding(Embedding, LoRALayer):
  def __init__(self, wrapped: Embedding, ...) -> None:
    self.wrapped = wrapped

  @property
   def wrapped_module(self) -> Embedding:
     return self.wrapped
```

cbalioglu: I think this also addresses your question in the PR summary about wrap/unwrap discrepancy about the model types. Ideally we should always preserve the original wrapped module within the LoRA module.

src/fairseq2/nn/lora.py
```diff
@@ -0,0 +1,373 @@
...
+        """
+        Embedding.__init__(
+            self,
+            wrapped.num_embeddings,
+            wrapped.embedding_dim,
+            wrapped.pad_idx,
+        )
+        LoRALayer.__init__(self, config)
+
+        self.weight = Parameter(
```
cbalioglu, line 81: How about:

```python
wrapped.weight.requires_grad_(False)

self.weight = wrapped.weight
```

This would preserve the object identity and would not wrap a Parameter within another Parameter.

src/fairseq2/nn/lora.py
```diff
@@ -0,0 +1,373 @@
...
+            torch.empty((self.r, self.input_dim), device=device, dtype=dtype)
+        )
+
+        self.lora_B = Parameter(
+            torch.empty((self.output_dim, self.r), device=device, dtype=dtype)
+        )
+
+        if self.dropout_p > 0.0:
+            self.dropout = Dropout(self.dropout_p)
+
```
cbalioglu, line 183: Nit: please remove the blank line here.

src/fairseq2/nn/lora.py
```diff
@@ -0,0 +1,373 @@
...
+        self.reset_lora_parameters()
+
+    def forward(self, x: Tensor) -> Tensor:
+        if self.merged:
+            return linear(x, self.weight, self.bias)
+        else:
+            h1 = linear(x, self.weight, self.bias)
+
+            if self.dropout is not None:
```
cbalioglu, line 199: To avoid repetitiveness, you can say:

```python
...
if self.dropout is not None:
  x = self.dropout(x)

h2 = linear(x, self.lora_B @ self.lora_A * self.scaling)
...

File: src/fairseq2/nn/lora.py [hunk 7]
```diff
@@ -0,0 +1,373 @@
...
+        if not isinstance(submodule, LoRALayer):
+            continue
+
+        submodule_path = name.split(".")
+        parent = module.get_submodule(".".join(submodule_path[:-1]))
+        submodule_name = submodule_path[-1]
+
+        unwrapped_layer: Optional[nn.Module] = None
+        if isinstance(submodule, LoRALinear):
+            # TODO: currently there's no way to distinguish which type the
```
cbalioglu, line 291: Please see my comment above regarding this.

src/fairseq2/nn/lora.py
```diff
@@ -0,0 +1,373 @@
...
+
+def unmerge_lora(module: nn.Module) -> None:
+    for submodule in module.modules():
+        if isinstance(submodule, LoRALayer):
+            submodule.unmerge()
+
+
+def lora_state_dict(module: nn.Module) -> Dict[str, Any]:
+    state_dict = module.state_dict()
+    lora_states = {name: state for name, state in state_dict.items() if "lora_" in name}
```
cbalioglu, line 344: I think just checking the "lora_" within the parameter name is error prone. How about retrieving the full names of LoRA layers by iterating the model (e.g. `if isinstance(submodule, LoRALayer): names.append(submodule_name)`) and then using them to fetch their state_dicts?

src/fairseq2/nn/lora.py
```diff
@@ -0,0 +1,373 @@
...
+    return lora_states
+
+
+def freeze_non_lora(
+    module: nn.Module, unfreeze_bias: Literal["none", "all", "lora_only"] = "none"
+) -> None:
+    # Set requires_grad to False for all parameters in the module except
+    # lora layers
+    for name, param in module.named_parameters():
```
cbalioglu, line 353: Similarly, this is also error prone. I would do it in semi-pseudo code:

```python
for name, submodule in module.named_modules():
   if not isinstance(submodule, LoRALayer):
      for param_name, param in submodule.named_parameters(recurse=False):
         if param_name == "bias":
            ...
         else:
           param.requires_grad = False
```

src/fairseq2/nn/lora.py
```diff
@@ -0,0 +1,373 @@
...
+            if (
+                isinstance(submodule, LoRALayer)
+                and getattr(submodule, "bias", None) is not None
+            ):
+                submodule.bias.requires_grad = True
+
+
+def _is_target_module(name: str, target_keys: List[str]) -> bool:
+    # Check if the `name` matches any of the `target_keys``.
+    return any(re.match(key, name) for key in target_keys) or any(
```
cbalioglu, line 371: Nit: instead of iterating twice you can do:

```python
return any(name == key or re.match(key, name) for key in target_keys)
```

src/fairseq2/nn/lora.py
chaoweihuang: Updated in the new version

src/fairseq2/nn/lora.py
chaoweihuang: Thanks for the suggestion! Updated in the new version.

src/fairseq2/nn/lora.py
chaoweihuang: Updated

src/fairseq2/models/llama/builder.py
```diff
@@ -357,3 +358,12 @@ def create_llama_model(
...
     """
     return LLaMABuilder(config, device=device, dtype=dtype).build_model()
+
+
+def get_llama_lora_config() -> LoRAConfig:
+    return LoRAConfig(
+        r=8,
+        alpha=16.0,
+        dropout_p=0.05,
+        keys=[".*decoder.layers.*.self_attn.*(q_proj|v_proj)$"],
```
cbalioglu, line 368: Just to double check, we do not fine-tune K projection, is that right?

src/fairseq2/nn/lora.py
```diff
@@ -0,0 +1,361 @@
...
+
+        self.wrapped.weight.requires_grad_(False)
+
+        self.weight = self.wrapped.weight  # type: ignore[assignment]
+
+        if self.wrapped.bias is not None:
+            self.wrapped.bias.requires_grad_(False)
+            self.bias = self.wrapped.bias  # type: ignore[assignment]
+        else:
+            self.register_module("bias", None)
```
cbalioglu, line 179: Small bug, this should be `register_parameter`.

tests/unit/nn/test_lora.py
```diff
@@ -0,0 +1,174 @@
...
+
+import torch
+from torch.nn.functional import embedding, linear
+
+from fairseq2.nn import Linear
+from fairseq2.nn.embedding import StandardEmbedding
+from fairseq2.nn.lora import LoRAConfig, LoRAEmbedding, LoRALinear
+
+
+def test_lora_liner_works() -> None:
```
cbalioglu, line 17: Nit: s/liner/linear/

cbalioglu (commented):
Nice! Left two minor comments, but otherwise looks good to me.

src/fairseq2/models/llama/builder.py
chaoweihuang: Yes, only q and v projections. This is the default setting from the original LoRA paper (https://arxiv.org/abs/2106.09685).

Though it might be worth noting that, recent papers have shown that fine-tuning only q&v projections might not be sufficient for larger models. The recent QLoRA paper (https://arxiv.org/abs/2305.14314) suggested fine-tuning all projection layers.

cbalioglu (approved):
LGTM!

PR #94, opened 2023-10-06
Adding `self_attn_mask` to the Transformer Encoder Layer API.

kauterry:

- Adds `self_attn_mask `to the Transformer Encoder Layer API.
- Adds `conv_norm_type` as an explicit config parameter for the Conformer convolution module.

Tested with m4t_predict for S2ST and confirm that the behavior is as expected.

**Check list:**

### DIFF
--- a/src/fairseq2/models/s2t_transformer/builder.py
+++ b/src/fairseq2/models/s2t_transformer/builder.py
@@ -5,7 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 from dataclasses import dataclass
-from typing import Literal, Optional
+from typing import Optional
 
 from torch.nn import SiLU
 
@@ -90,9 +90,6 @@ class S2TTransformerConfig:
     depthwise_conv_kernel_size: int
     """The kernel size of depthwise convolutions in Conformer blocks."""
 
-    conv_norm_type: Literal["batch_norm", "layer_norm"]
-    """The type of norm layer in the Conformer convolution module."""
-
     def update_target_vocabulary(self, info: VocabularyInfo) -> None:
         """Update target vocabulary configuration from ``info``."""
         self.target_vocabulary_size, self.target_pad_idx = info.size, info.pad_idx
@@ -121,7 +118,6 @@ def _tiny() -> S2TTransformerConfig:
         ffn_inner_dim=256 * 4,
         dropout_p=0.3,
         depthwise_conv_kernel_size=0,
-        conv_norm_type="batch_norm",
     )
 
 
@@ -142,7 +138,6 @@ def _small() -> S2TTransformerConfig:
         ffn_inner_dim=256 * 8,
         dropout_p=0.1,
         depthwise_conv_kernel_size=0,
-        conv_norm_type="batch_norm",
     )
 
 
@@ -163,7 +158,6 @@ def _medium() -> S2TTransformerConfig:
         ffn_inner_dim=512 * 4,
         dropout_p=0.15,
         depthwise_conv_kernel_size=0,
-        conv_norm_type="batch_norm",
     )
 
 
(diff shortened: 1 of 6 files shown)


src/fairseq2/models/s2t_transformer/builder.py
```diff
@@ -90,6 +90,9 @@ class S2TTransformerConfig:
     depthwise_conv_kernel_size: int
     """The kernel size of depthwise convolutions in Conformer blocks."""
 
+    conv_norm_type: Literal["batch_norm", "layer_norm"]
```
cbalioglu, line 93: I don't think we should expose this as a configuration parameter for S2T Transformer. That model architecture has never used Conformer with LayerNorm and will like never use it.

src/fairseq2/models/s2t_transformer/builder.py
```diff
@@ -380,6 +388,7 @@ def build_conformer_block(self) -> TransformerEncoderLayer:
         conv = ConformerConvolution(
             self.config.model_dim,
             self.config.depthwise_conv_kernel_size,
+            norm_type=self.config.conv_norm_type,
```
cbalioglu, line 391: As I mentioned above, S2T Transformer never had an architecture using "layer_norm" Conformer blocks. We can remove this "conv_norm_type" parameter from its configuration set.

src/fairseq2/models/w2vbert/builder.py
```diff
@@ -43,6 +43,7 @@ def _encoder_600m() -> Wav2Vec2EncoderConfig:
         layer_drop_p=0.0,
         norm_order=TransformerNormOrder.POST,
         depthwise_conv_kernel_size=31,
+        conv_norm_type="batch_norm",
```
cbalioglu, line 46: Since "batch_norm" is the default anyways, I don't think that this extra argument pass is needed.

src/fairseq2/models/w2vbert/builder.py
```diff
@@ -74,6 +75,7 @@ def _encoder_300m() -> Wav2Vec2EncoderConfig:
...
```
cbalioglu, line 78: Same as above.

src/fairseq2/models/wav2vec2/builder.py
```diff
@@ -139,6 +139,9 @@ class Wav2Vec2EncoderConfig:
     depthwise_conv_kernel_size: int
     """The kernel size of depthwise convolutions in Conformer blocks."""
 
+    conv_norm_type: Literal["batch_norm", "layer_norm"]
```
cbalioglu, line 142: Unlike S2T Transformer and w2v-BERT, I assume you need "conv_norm_type" here for the new wav2vec2 architecture in SC. If that is the case, we can keep this, but otherwise, you can remove it similar to other models.

src/fairseq2/models/wav2vec2/builder.py
```diff
@@ -139,6 +139,9 @@ class Wav2Vec2EncoderConfig:
...
+    """The type of norm layer in the Conformer convolution module."""
```
cbalioglu, line 143: Very nit: I would remove the term "layer" from this docstr to avoid confusion between batch and layer normalization. I would just write "The type of normalization to use in Conformer convolution modules".

cbalioglu (commented):
Nice! Just left a few nit comments, but nothing major. Let me know once you address my feedback and I can stamp/merge it right away.

src/fairseq2/models/wav2vec2/builder.py
kauterry: Yes we do need it for the new wav2vec2 model.

cbalioglu (approved):
LGTM!

PR #95, opened 2023-10-07
Introduce `causal_depthwise_conv` in the Conformer convolution module.

kauterry:

Pad the input sequence entirely on the left for the Conv1d making it causal.

**Check list:**

### DIFF
--- a/src/fairseq2/models/conformer/convolution.py
+++ b/src/fairseq2/models/conformer/convolution.py
@@ -8,6 +8,7 @@ from typing import Optional
 
 from torch import Tensor
 from torch.nn import GLU, BatchNorm1d, Conv1d, Module, SiLU
+from torch.nn.functional import pad
 
 from fairseq2.nn.normalization import LayerNorm, StandardLayerNorm
 from fairseq2.nn.utils.mask import apply_padding_mask
@@ -22,6 +23,8 @@ class ConformerConvolution(Module):
     pointwise_conv1: Conv1d
     pointwise_conv1_activation: GLU
     depthwise_conv: Conv1d
+    depthwise_kernel_size: int
+    causal_depthwise_conv: bool
     batch_norm: Optional[BatchNorm1d]
     layer_norm: Optional[LayerNorm]
     depthwise_activation: Module
@@ -32,6 +35,7 @@ class ConformerConvolution(Module):
         model_dim: int,
         depthwise_kernel_size: int,
         *,
+        causal_depthwise_conv: bool = False,
         norm_type: str = "batch_norm",
         depthwise_activation: Optional[Module] = None,
         device: Optional[Device] = None,
@@ -42,6 +46,9 @@ class ConformerConvolution(Module):
             The dimensionality of the model.
         :param depthwise_kernel_size:
             The kernel size of the depthwise convolution.
+        :param causal_depthwise_conv:
+            If True, uses a causal depthwise convolution similar to that described in
+            Section 2.1 of :cite:t:`https://doi.org/10.48550/arxiv.2006.11477`.
         :param norm_type:
             The type of norm layer applied after the depthwise convolution.
         :param depthwise_activation:
@@ -71,13 +78,15 @@ class ConformerConvolution(Module):
             model_dim,
             depthwise_kernel_size,
             # We preserve the sequence length regardless of the kernel size.
-            padding="same",
+            padding="same" if not causal_depthwise_conv else 0,
             # We want to perform depthwise convolution.
             groups=model_dim,
             bias=False,
             device=device,
             dtype=dtype,
         )
+        self.depthwise_kernel_size = depthwise_kernel_size
+        self.causal_depthwise_conv = causal_depthwise_conv
 
         if norm_type not in ("batch_norm", "layer_norm"):
             raise ValueError(
@@ -135,6 +144,10 @@ class ConformerConvolution(Module):
 
         # (N, 2 * M, S) -> (N, M, S)
         seqs = self.pointwise_conv1_activation(seqs)
+
+        # Pad the sequence entirely on the left in the case of a causal convolution.
+        if self.causal_depthwise_conv:
+            seqs = pad(seqs, (self.depthwise_kernel_size - 1, 0))
 
         # (N, M, S) -> (N, M, S)
         seqs = self.depthwise_conv(seqs)
--- a/src/fairseq2/models/conformer/convolution.py
+++ b/src/fairseq2/models/conformer/convolution.py
@@ -48,7 +48,7 @@ class ConformerConvolution(Module):
             The kernel size of the depthwise convolution.
         :param causal_depthwise_conv:
             If True, uses a causal depthwise convolution similar to that described in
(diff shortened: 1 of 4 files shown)


src/fairseq2/models/conformer/convolution.py
```diff
@@ -14,6 +15,49 @@
...
+            bias=bias,
+            padding_mode=padding_mode,
+            device=device,
+            dtype=dtype,
+        )
+        # Pad the last dimension entirely on the left to make the convolution causal.
+        self.pad = (kernel_size - 1, 0)
+
+    def forward(self, x: Tensor) -> Tensor:
+        x = F.pad(x, self.pad)
```
cbalioglu, line 57: Do we use this `CausalConv1d` anywhere else beyond a `ConformerConvolution`? If not, I don't think that it is necessary to have a module for this functionality. We can achieve the same within `ConformerConvolution`'s forward method like:

```python
def forward(...):
   ...

   if self.depthwise_conv_is_causal:
       seqs = F.pad(seqs, (self.depthwise_conv.kernel_size[0] - 1, 0))
   
   seqs = self.depthwise_conv(seqs)
```

cbalioglu: Also, do we have any literature reference for this padding technique?

src/fairseq2/models/conformer/convolution.py
```diff
@@ -42,6 +87,8 @@ def __init__(
             The dimensionality of the model.
         :param depthwise_kernel_size:
             The kernel size of the depthwise convolution.
+        :param depthwise_conv_is_causal:
```
cbalioglu, line 90: Nit: I would prefer this to be named `causal_depthwise_conv` which is a bit more aligned with the naming convention we use in other places in the code base.

src/fairseq2/models/conformer/convolution.py
```diff
@@ -42,6 +87,8 @@ def __init__(
...
+            Is the depthwise convolution causal or not.
```
cbalioglu, line 91: Nit: For boolean parameters, we typically use the convention "If ``True``, uses a causal depthwise convolution." or even better if we have a literature reference "If ``True``, uses a causal depthwise convolution as described in :cite:t`xyz`".

src/fairseq2/models/conformer/convolution.py
```diff
@@ -66,18 +113,32 @@ def __init__(
...
-            depthwise_kernel_size,
-            # We preserve the sequence length regardless of the kernel size.
-            padding="same",
-            # We want to perform depthwise convolution.
-            groups=model_dim,
-            bias=False,
-            device=device,
-            dtype=dtype,
-        )
+        if depthwise_conv_is_causal:
```
cbalioglu, line 116: Again, as I mentioned above. To me it looks like the only difference between these two modules is the extra pad statement in `CausalConv1d`. If that is true, we can certainly move it to `ConformerConvolution` and avoid this extra convolution module.

src/fairseq2/models/s2t_transformer/builder.py
```diff
@@ -90,6 +90,9 @@ class S2TTransformerConfig:
     depthwise_conv_kernel_size: int
     """The kernel size of depthwise convolutions in Conformer blocks."""
 
+    depthwise_conv_is_causal: bool
```
cbalioglu, line 93: S2T Transformer architecture has never supported causal depthwise convolution. We should not expose this as a configurable parameter in S2T Transformer.

src/fairseq2/models/w2vbert/builder.py
```diff
@@ -43,6 +43,7 @@ def _encoder_600m() -> Wav2Vec2EncoderConfig:
         layer_drop_p=0.0,
         norm_order=TransformerNormOrder.POST,
         depthwise_conv_kernel_size=31,
+        depthwise_conv_is_causal=False,
```
cbalioglu, line 46: Since the default is already `False` and this parameter is not relevant for w2v-BERT please remove this explicit argument pass.

src/fairseq2/models/w2vbert/builder.py
```diff
@@ -75,6 +76,7 @@ def _encoder_300m() -> Wav2Vec2EncoderConfig:
...
```
cbalioglu, line 79: Same as above.

src/fairseq2/models/wav2vec2/builder.py
```diff
@@ -139,6 +139,9 @@ class Wav2Vec2EncoderConfig:
     depthwise_conv_kernel_size: int
     """The kernel size of depthwise convolutions in Conformer blocks."""
 
+    depthwise_conv_is_causal: bool
```
cbalioglu, line 142: Same parameter name and docstr (one line below) suggestion as in conformer/convolution.py.

src/fairseq2/models/conformer/convolution.py
```diff
@@ -6,6 +6,7 @@
 
 from typing import Optional
 
+import torch.nn.functional as F
```
cbalioglu, line 9: Recently I have started using explicit imports (and updated the code base to follow that convention), so in the rest of the code base we use `from torch.nn.functional import pad`. I personally prefer this style since it gives you immediately the overview of that features/ops we use at the top of the file.

cbalioglu (commented):
Mostly looks good. I have one feedback about the existence of the `CausalConv1d` module though. Let me know what you think. Otherwise, it is pretty close to be merged.

src/fairseq2/models/conformer/convolution.py
kauterry: I believe it's used in WaveNet and similar architectures: https://www.deepmind.com/blog/wavenet-a-generative-model-for-raw-audio

It's a pretty common idea, so I wondered if we should have it as a module for anyone to leverage.

src/fairseq2/models/w2vbert/builder.py
cbalioglu: Nevermind. Ignore my comment. This is `Wav2Vec2EncoderConfig` so this parameter is required to be passed.

src/fairseq2/models/conformer/convolution.py
kauterry: Alrighty.

src/fairseq2/models/conformer/convolution.py
```diff
@@ -42,6 +46,9 @@ def __init__(
             The dimensionality of the model.
         :param depthwise_kernel_size:
             The kernel size of the depthwise convolution.
+        :param causal_depthwise_conv:
+            If True, uses a causal depthwise convolution similar to that described in
+            Section 2.1 of :cite:t:`https://doi.org/10.48550/arxiv.2006.11477`.
```
cbalioglu, line 51: This is wav2vec 2.0 paper right? I don't see a Section 2.1 or a reference to causality there. Are you sure that this is the right reference. Also I forgot to mention it in the first review, but you should also put the bibtex of the paper to bibliography.bib at the root of the repository.

kauterry: Oh no, wrong link sorry!

cbalioglu (approved):

PR #104, opened 2023-10-16
Improve padding and attention mask handling

cbalioglu:
This PR includes several closely-coupled changes:

1) The major change is the introduction of the new `PaddingMask` and `AttentionMask` types that replace all uses of `seq_lens` and tensor-based `padding_mask` and `self_attn_mask` parameters. This is also the main reason why this PR touches many files since those parameters are used in many places in the code base. The main benefit of these two new types is the lazy initialization of masks and not materializing them at all whenever possible (e.g. when a global causal attention mask is used with PyTorch SDPA or xformers, we entirely skip constructing the mask). This can significantly improve the performance in the right settings and also reduces memory use (again, sometimes significantly when used with very large context lengths).
2) `MultiheadAttention` also needed to be refactored. Mask handling is moved down to `SDPA` implementations since each SDPA has varying support for different mask types. As part of this work `add_bias_kv` and `add_zero_attn` are also removed. Initially we introduced them to have feature parity with fairseq MHA, but turns out that these two parameters are not used anywhere. We can consider introducing them in the future again if requested, but they add needless complexity to mask handling and both parameters are superseded quite a while ago by superior techniques (e.g. pre-LN).
3) Various nit updates at several places in the code base (like renaming some local variables, updating docstrs) mainly related to the first two points.
4) Fixes the causal attention mask to support bfloat16.

Tested with LLaMA 2 and SeamlessM4T models in various settings and verified bitwise output parity.

### DIFF
--- a/doc/reference/abc.rst
+++ b/doc/reference/abc.rst
@@ -9,14 +9,3 @@ ABCs and Protocols
     :nosignatures:
 
     gang.Gang
-    nn.PositionEncoder
-    nn.Projection
-    nn.transformer.AttentionMaskGenerator
-    nn.transformer.AttentionWeightHook
-    nn.transformer.FeedForwardNetwork
-    nn.transformer.MultiheadAttention
-    nn.transformer.SDPA
-    nn.transformer.TransformerDecoder
-    nn.transformer.TransformerDecoderLayer
-    nn.transformer.TransformerEncoder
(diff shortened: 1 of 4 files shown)


cbalioglu:
Manually verified clang-tidy check. GitHub runner is (again) out of disk space.

PR #146, opened 2023-11-09
enabling GELU for nllb text decoder (for Expressivity model compatibi…

yilinyang7:
This PR is to help enabling GELU activation for the text decoder (of Expressivity model).

### DIFF
--- a/src/fairseq2/models/nllb/builder.py
+++ b/src/fairseq2/models/nllb/builder.py
@@ -7,6 +7,8 @@
 from dataclasses import dataclass
 from typing import Optional
 
+from torch.nn import Module, ReLU
+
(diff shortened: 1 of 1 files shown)


src/fairseq2/models/nllb/builder.py
```diff
@@ -67,6 +69,9 @@ class NllbConfig:
     dropout_p: float
     """The dropout probability in Transformer layers."""
 
+    decoder_activate_fn: Module = ReLU()
```
kauterry, line 72: nit: rename this to `decoder_activation_fn`

yilinyang7: done, thanks

cbalioglu:
resolved internally.

PR #157, opened 2023-11-15
Nit update to BannedSequenceProcessor

cbalioglu:
A nit update to `BlockSequenceProcessor` for the (very) edge case when the `sum()` op happens to output 0 for non matching token sequences (e.g. `sum((4, 5) - (5, 4)) == 0`).


shagunsodhani:
Ha! this is a great catch!

PR #164, opened 2023-11-17
Revise AssetDownloadManager

cbalioglu:
This PR revises the implementation of `AssetDownloadManager`.


kauterry:
> This PR revises the implementation of `AssetDownloadManager`.

Could you add some details about the revision of this class? What are we able to achieve now which wasn't possible before?

PR #179, opened 2023-11-27
Extend mypy coverage

cbalioglu:
This PR fixes some of the subtle mypy package resolving issues by explicitly listing package directories.


.github/workflows/_lint_py.yaml
```diff
@@ -99,4 +99,6 @@ jobs:
 
           trap remove_matcher EXIT
 
-          mypy .
+          mypy
```
shagunsodhani, line 102: I dont know what this do. I have only seen mypy used along with some path (like in the next line). I googled but did not find anything.

cbalioglu: When you don't provide a path/package/module, it checks the configuration file (in fairseq2 it is pyproject.toml) to figure out the default setting. I updated pyproject.toml in this PR to explicitly list src, fairseq2n/python, and tests as the paths it should check.

shagunsodhani: Thank you :)

PR #182, opened 2023-11-28
Reimplement beam search

cbalioglu:
The existing beam search implementation in fairseq2 was mostly a cleaned-up copy/paste version of fairseq due to a very tight release deadline. This PR completely rewrites it. The implementation is much cleaner and simpler while having bitwise/runtime performance parity with the fairseq version (in particular the 2xbeam_size trick is used here as well) while having slightly less memory overhead. We also now have abstractions for both Decoder and EncoderDecoder models so that beam search can be used both with regular MT models as well as language models.


Mortimerp9:
this breaks the previous API it seems (for the demos)

lifeiteng:
so many break changes. do you plan to update https://github.com/facebookresearch/seamless_communication ?

Mortimerp9:
@lifeiteng seamless_communication should still work with fairseq2 0.1.*
We are working on an updated version to follow the fairseq2 updates.

PR #185, opened 2023-11-28
Use fixed _legacy_pad_idx for fairseq reproducibility

cbalioglu:
This PR uses "1" as `_legacy_pad_idx` argument for NLLB and S2T Transformer instead of pad_idx to ensure full parity with fairseq model outputs.


yilinyang7 (approved):
Thanks for the quick fix!

Issue #170, opened 2023-11-20
fairseq2n is not compatible with fmt in PyTorch nightlies


ezyang:
```
ome/ezyang/local/fairseq2/fairseq2n/python/src/fairseq2n/bindings/data/string.cc:7:
/home/ezyang/local/pytorch-venv/torch/include/fmt/core.h: In instantiation of ‘constexpr fmt::v10::detail::value<Context> fmt::v10::detail::make_arg(T&) [with bool PACKED = true; Context = fmt::v10::basic_format_context<fmt::v10::appender, char>; T = const fairseq2n::immutable_string; typename std::enable_if<PACKED, int>::type <anonymous> = 0]’:
/home/ezyang/local/pytorch-venv/torch/include/fmt/core.h:1808:51:   required from ‘constexpr fmt::v10::format_arg_store<Context, Args>::format_arg_store(T& ...) [with T = {const fairseq2n::immutable_string}; Context = fmt::v10::basic_format_context<fmt::v10::appender, char>; Args = {fairseq2n::immutable_string}]’
/home/ezyang/local/pytorch-venv/torch/include/fmt/core.h:1826:18:   required from ‘constexpr fmt::v10::format_arg_store<Context, typename std::remove_cv<typename std::remove_reference<T>::type>::type ...> fmt::v10::make_format_args(T& ...) [with Context = fmt::v10::basic_format_context<fmt::v10::appender, char>; T = {const fairseq2n::immutable_string}]’
/home/ezyang/local/pytorch-venv/torch/include/fmt/core.h:2788:44:   required from ‘std::string fmt::v10::format(fmt::v10::format_string<T ...>, T&& ...) [with T = {const fairseq2n::immutable_string&}; std::string = std::__cxx11::basic_string<char>; fmt::v10::format_string<T ...> = fmt::v10::basic_format_string<char, const fairseq2n::immutable_string&>]’
/home/ezyang/local/fairseq2/fairseq2n/python/src/fairseq2n/bindings/data/string.cc:68:35:   required from here
/home/ezyang/local/pytorch-venv/torch/include/fmt/core.h:1576:63: error: ‘fmt::v10::detail::type_is_unformattable_for<const fairseq2n::immutable_string, char> _’ has incomplete type
 1576 |     type_is_unformattable_for<T, typename Context::char_type> _;
      |                                                               ^
[...]

Issue #152, opened 2023-11-14
Cannot build with GCC 11.4.1: error: ‘fmt::v10::detail::type_is_unformattable_for<const fairseq2n::immutable_string, char> _’ has incomplete type


ezyang:
**Describe the bug:**

```
FAILED: python/src/fairseq2n/bindings/CMakeFiles/py_bindings.dir/data/string.cc.o 
/usr/bin/c++ -D_FORTIFY_SOURCE=2 -D_GLIBCXX_USE_CXX11_ABI=1 -Dpy_bindings_EXPORTS -I/home/ezyang/local/d/fairseq2/fairseq2n/python/src -I/home/ezyang/local/d/fairseq2/fairseq2n/src -I/home/ezyang/local/d/fairseq2/fairseq2n/build/src -isystem /home/ezyang/local/d/pytorch-env/include/python3.10 -isystem /home/ezyang/local/d/fairseq2/fairseq2n/third-party/pybind11/include -isystem /home/ezyang/local/d/pytorch/torch/include -isystem /home/ezyang/local/d/pytorch/torch/include/torch/csrc/api/include -isystem /home/ezyang/local/d/fairseq2/fairseq2n/third-party/fmt/include -isystem /home/ezyang/local/d/fairseq2/fairseq2n/third-party/zip/src -O2 -g -DNDEBUG -std=c++17 -fPIC -fvisibility=hidden -Wall -Wcast-align -Wconversion -Wdouble-promotion -Wextra -Wfloat-equal -Wformat=2 -Winit-self -Wlogical-op -Wno-unknown-pragmas -Wpointer-arith -Wshadow=compatible-local -Wsign-conversion -Wswitch -Wunused -Wnon-virtual-dtor -Woverloaded-virtual -Wuseless-cast -fasynchronous-unwind-tables -fstack-protector-strong -MD -MT python/src/fairseq2n/bindings/CMakeFiles/py_bindings.dir/data/string.cc.o -MF python/src/fairseq2n/bindings/CMakeFiles/py_bindings.dir/data/string.cc.o.d -o python/src/fairseq2n/bindings/CMakeFiles/py_bindings.dir/data/string.cc.o -c /home/ezyang/local/d/fairseq2/fairseq2n/python/src/fairseq2n/bindings/data/string.cc
In file included from /home/ezyang/local/d/fairseq2/fairseq2n/src/fairseq2n/fmt.h:23,
[...]

PR #192, opened 2023-11-28
Reorder CMake Python binding dependencies

cbalioglu:
This PR changes the CMake dependency order for the Python C extension library to avoid header file conflicts with PyTorch. See #170, #152.


cbalioglu:
cc @ezyang

PR #203, opened 2023-12-10
Revise third-party builds

cbalioglu:
This PR revises the handling of third-party lib builds for fairseq2n including migrating to static libjpeg-turbo and libpng. It also temporarily disables JPEG decoding tests due to a segfault on Macs and DataPipeline sampling tests due to issues with RNG on Macs.


lifeiteng:
@cbalioglu Why do directory names change so frequently?
I like developing and testing my codebase(base on fairseq2) on multiple platforms, and on mac I have to build fairseq2 from source, but it changes too frequently.

PR #205, opened 2023-12-11
Add runtime torch version check

cbalioglu:
This PR attempts to improve the user experience at runtime by reporting a version mismatch between PyTorch and fairseq2 instead of failing later by a non-descriptive error.

### DIFF
--- a/native/python/setup.py
+++ b/native/python/setup.py
@@ -8,7 +8,6 @@ from os import path
 from typing import Final, List, Optional
 
 import torch
-from packaging import version
 from setuptools import Command, find_packages, setup
 from setuptools.command.install import install as install_base
 from setuptools.dist import Distribution as DistributionBase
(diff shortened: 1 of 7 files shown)


native/python/setup.py
```diff
@@ -167,7 +166,7 @@ def get_inputs(self) -> List[str]:
         "tbb>=2021.8;platform_machine=='x86_64'",
         # PyTorch has no ABI compatibility between releases; this means we have
         # to ensure that we depend on the exact same version that we used to
-        # build our extension module.
-        "torch==" + version.parse(torch.__version__).public,
+        # build fairseq2n.
+        "torch==" + torch.__version__,
```
antoine-tran, line 170: why did we abandon the version.parse() way ? At least for me the old code works and the new one doesn't. I installed torch  via pip and get the version listed as '2.0.1' and not '2.0.1+cu117',

cbalioglu: Hey @antoine-tran, thanks for the report! Working on a fix now. With pip-based Python virtual environments it is the opposite case. So looks like I have to have handle conda and pip packages of PyTorch separately to have a full working solution.

PR #221, opened 2023-12-17
Introduce FSDP helper APIs

cbalioglu:
This PR introduces a few small FSDP helper functions, makes some nit improvements in docstrs, and updates the required isort version.


src/fairseq2/nn/utils/module.py
```diff
@@ -97,20 +130,20 @@ def empty_like(src: Tensor) -> Tensor:
...
+        for name, param in m.named_parameters(recurse=False):
+            if param is None:
                 continue
 
             with torch.no_grad():
-                new_prm = Parameter(empty_like(prm), prm.requires_grad)
+                new_param = Parameter(empty_like(param), param.requires_grad)
 
-            setattr(m, name, new_prm)
+            setattr(m, name, new_param)
```
shagunsodhani, line 140: Curious what is the benefit of using `setattr(m, name, new_param)` over `m.name = new_param`

cbalioglu: `name` here is of type string. So syntactically you cannot write a valid Python statement like `m."foo" = "xyz"`.

shagunsodhani: got it - thank you :)

cbalioglu: no problem! :)

PR #228, opened 2023-12-19
Introduce GeneratorBag

cbalioglu:
This PR introduces the utility `GeneratorBag` class.

### DIFF
--- a/src/fairseq2/utils/rng.py
+++ b/src/fairseq2/utils/rng.py
@@ -4,28 +4,14 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-import random
+from __future__ import annotations
 
-import numpy as np
+from typing import Any, Dict, List, Mapping
+
 import torch
+from torch import Generator, Tensor
 
-
-def seed(value: int) -> None:
-    """Set RNG seed for ``random``, ``np.random``, and ``torch``.
-
-    :param value:
-        The new seed.
-    """
-    if value >= 1 << 32:
-        raise ValueError(
-            f"`value` must be greater than or equal to 0 and less than 2^32, but is {value} instead."
-        )
-
-    random.seed(value)
-
-    np.random.seed(value)
-
-    torch.manual_seed(value)
+from fairseq2.typing import Device
 
 
 def use_deterministic(value: bool, warn_only: bool = False) -> None:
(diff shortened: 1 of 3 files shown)


src/fairseq2/utils/rng.py
```diff
@@ -40,3 +26,90 @@ def use_deterministic(value: bool, warn_only: bool = False) -> None:
...
+        """Create a :class:`GeneratorBag` instance holding the default random
+        number generators of ``devices``."""
+        generators = []
+
+        for device in set(devices):
+            if device.type == "cpu":
+                generators.append(torch.default_generator)
+            elif device.type == "cuda":
+                # Ensure that the default CUDA generators are initialized.
+                torch.cuda.init()
```
shagunsodhani, line 54: Curious why do we need this - doesnt this happen automatically ? Or is it related to the use of fairseqn

src/fairseq2/utils/rng.py
```diff
@@ -40,3 +26,90 @@ def use_deterministic(value: bool, warn_only: bool = False) -> None:
...
+
+                idx = device.index
+                if idx is None:
+                    idx = torch.cuda.current_device()
+
+                generators.append(torch.cuda.default_generators[idx])
```
shagunsodhani, line 60: Ha! I did not know about this!

src/fairseq2/utils/rng.py
```diff
@@ -40,3 +26,90 @@ def use_deterministic(value: bool, warn_only: bool = False) -> None:
...
+        self.generators[0].seed()
+
+        random_seed = self.generators[0].initial_seed()
+
+        for g in self.generators[1:]:
+            g.manual_seed(random_seed)
+
+    def manual_seed(self, seed: int) -> None:
+        """Set the seed of the random number generators."""
+        if seed >= 1 << 32:
```
shagunsodhani, line 82: Does this condition check if seed is negative ?

src/fairseq2/utils/rng.py
```diff
@@ -40,3 +26,90 @@ def use_deterministic(value: bool, warn_only: bool = False) -> None:
     torch.backends.cudnn.benchmark = not value
 
     torch.use_deterministic_algorithms(value, warn_only=warn_only)
+
+
+class GeneratorBag:
```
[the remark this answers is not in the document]
cbalioglu: Good to hear that. I was going back and forth between `RNGBag` and `GeneratorBag`. Decided to go with `GeneratorBag` to align with the `Generator` type in PyTorch (which I have never liked TBH. Too generic, also gets easily confused with the "generator expressions" in Python). Now hearing you also prefer `RNGBag`, let me rename this.

src/fairseq2/utils/rng.py
```diff
@@ -40,3 +26,90 @@ def use_deterministic(value: bool, warn_only: bool = False) -> None:
...
+        """
+        self.generators = list(generators)
+
+    @staticmethod
+    def from_device_defaults(*devices: Device) -> GeneratorBag:
+        """Create a :class:`GeneratorBag` instance holding the default random
+        number generators of ``devices``."""
+        generators = []
+
+        for device in set(devices):
```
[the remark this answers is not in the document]
cbalioglu: Once stored internally, the mapping is only relevant for `state_dict`/`load_state_dict`, but you are right. This might become confusing if people try to manually access `RNGBag.generators`. Let me fix this.

src/fairseq2/utils/rng.py
cbalioglu: The initialization happens automatically once you call a CUDA op, but, for instance, does not happen if your Python script starts with:

```
d = torch.device("cuda:0")

bag = RNGBag.from_default_devices(d)  # raises IndexError
```

Since we haven't called any CUDA op till this point, `torch.cuda.default_generators` will be empty. In vast majority of the use cases in the real world, CUDA will be already initialized by the time a function such as this will be called, but there is no guarantee. So I included this call to ensure it.

src/fairseq2/utils/rng.py
cbalioglu: Not really. Let me add that check as well.

src/fairseq2/utils/rng.py
shagunsodhani: Thank you for the explanation!

PR #229, opened 2023-12-19
Introduce StatefulObjectBag

cbalioglu:
This PR introduces the utility `StatefulObjectBag` type.

### DIFF
--- a/src/fairseq2/utils/state.py
+++ b/src/fairseq2/utils/state.py
@@ -1,0 +1,156 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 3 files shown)


src/fairseq2/utils/state.py
```diff
@@ -0,0 +1,156 @@
...
+    """Represents an object that follows the ``state_dict`` convention."""
+
+    def state_dict(self) -> Dict[str, Any]:
+        ...
+
+    def load_state_dict(self, state_dict: Mapping[str, Any]) -> None:
+        ...
+
+
+class StatefulObjectBag:
```
shagunsodhani, line 39: Is this object intended to be used as a singleton ?

cbalioglu: No. Let me share the training recipe tomorrow. There you can see how it is used.

PR #232, opened 2023-12-20
Fix `extract_masked_elements` unflatten bug

jc-audet:

The function `extract_masked_elements` in `models/wav2vec2/masker.py` is not behaving as expected because it is breaking down the pipeline on my setup and the current behavior of `seqs` does not seems to align with the commented shapes.

Specifically the N dimension in the unflatten operation is referring to the `size(0)` of a (N x T, M) tensor which leads to a tensor of shape (N x T, 1, M), while the intended output is (N, T, M). This PR fixes the problem on my setup.

**Check list:**


src/fairseq2/models/wav2vec2/masker.py
```diff
@@ -149,8 +149,11 @@ def extract_masked_elements(seqs: Tensor, temporal_mask: Tensor) -> Tensor:
         The temporal mask. *Shape:* :math:`(N,S)`, where :math:`N` is the batch
         size and :math`S` is the sequence length.
     """
+    
+    N=seqs.size(0)
```
cbalioglu, line 153: Nit: do you mind naming this variable `batch_size`? I know it is more verbose, but, as a convention, we use that variable name in several places in the code base.

cbalioglu (commented):
Just left a nit comment. And looks like black complains about formatting issues. Let me know once you address them, I can stamp and merge it right away.

src/fairseq2/models/wav2vec2/masker.py
jc-audet: Done!

cbalioglu (approved):
LGTM!

PR #233, opened 2023-12-20
Introduce CheckpointManager

cbalioglu:
This PR introduces `CheckpointManager`, and performs some nit docstr updates and variable renaming.

### DIFF
--- a/src/fairseq2/checkpoint.py
+++ b/src/fairseq2/checkpoint.py
@@ -1,0 +1,329 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from abc import ABC, abstractmethod
+from pathlib import Path
+from pickle import PickleError
+from typing import Any, Dict, Optional, Tuple, cast, final
+
+import torch
+from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
+from torch.distributed.fsdp.api import FullStateDictConfig, StateDictType
+from torch.nn import Module
+
+from fairseq2.gang import Gang
+from fairseq2.models.utils.checkpoint import load_checkpoint
+from fairseq2.nn.utils.module import (
+    infer_device,
+    reset_non_persistent_buffers,
+    to_empty,
+)
+from fairseq2.typing import CPU, META, Device, finaloverride
+
+
+class CheckpointManager(ABC):
+    """Saves and loads training checkpoints."""
+
+    @abstractmethod
+    def save_checkpoint(self, step_nr: int, state: Dict[str, Any]) -> None:
+        """Save a checkpoint.
+
+        :param step_nr:
+            The training step number of the checkpoint.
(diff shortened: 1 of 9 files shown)


src/fairseq2/checkpoint.py
```diff
@@ -0,0 +1,329 @@
...
+                raise RuntimeError(
+                    f"The checkpoint directory for step {step_nr} cannot be created. See inner exception for details."
+                ) from ex
+
+        self.gang.barrier()
+
+        checkpoint_file = step_dir.joinpath(f"rank_{self.gang.rank}.pt")
+
+        try:
+            torch.save(checkpoint, checkpoint_file)
```
shagunsodhani, line 137: For DDP, wouldn't we end up saving redundant copies of the model ?

src/fairseq2/checkpoint.py
```diff
@@ -0,0 +1,329 @@
...
+
+    @finaloverride
+    def save_checkpoint(self, step_nr: int, state: Dict[str, Any]) -> None:
+        checkpoint = {"step_nr": step_nr, "data": state}
+
+        step_dir = self.checkpoint_dir.joinpath(f"step_{step_nr}")
+
+        if self.gang.rank == 0:
+            try:
+                step_dir.mkdir(parents=True, exist_ok=True)
```
shagunsodhani, line 126: When checkpointing FSDP models on non-nfs / local partitions, we may want to execute this step on all the nodes.

src/fairseq2/checkpoint.py
cbalioglu: For `CheckpointManager` the provided `state` is an opaque dict. It does not really care what it contains. In a DDP setting, the trainer would do something like:

```
state_dict = {...}

if rank == 0:
    state_dict["model"] = model.state_dict()

checkpoint_manager.save_checkpoint(1234, state_dict)
```

src/fairseq2/checkpoint.py
cbalioglu: What do you mean by "non-nfs/local" partitions?

src/fairseq2/checkpoint.py
shagunsodhani: Okay so it is the responsibility of the caller to make sure they do not call `checkpoint_manager.save_checkpoint(1234, state_dict)` across multiple ranks when training with ddp.

src/fairseq2/checkpoint.py
shagunsodhani: I was thinking of the case where we are checkpointing shards when using FSDP. In that case, all ranks will checkpoint their shard to the disk. We are creating the directory only in rank=0 process. Now if model was saved to a partition that is local to each node (i.e. physical machine), the checkpoint directory would only exist for the node with rank=0 process.

cbalioglu: Oh I see what you mean. In majority of cases, nodes share a common distributed file system (e.g NFS, fsx), but I think it is a fair ask to support local disk setup. Let me revise this implementation to address your feedback. Thanks!

PR #237, opened 2023-12-22
gradio interface for mistral chatbot

uralik:
**What does this PR do?**

Adds a gradio interface for mistral instruct 7b chatbot

@cbalioglu  minor observation is that the nucleus param change doesnt influence the output as much as I would expect. Worth checking this further.

### DIFF
--- a/recipes/mistral/gradio_chatbot.py
+++ b/recipes/mistral/gradio_chatbot.py
@@ -1,0 +1,67 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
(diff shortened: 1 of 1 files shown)


uralik:
Example: 
image

recipes/mistral/gradio_chatbot.py
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+try:
+    import gradio as gr
+except ImportError:
+    print("Install gradio: pip install gradio")
```
shagunsodhani, line 10: Should we specify the version of gradio ?

PR #245, opened 2023-12-28
Revise setup_logging

cbalioglu:
This PR revises `setup_logging` utility function so that only rank 0 (i.e. coordinator rank/process) writes to a real file, while all ranks log to stderr.

### DIFF
--- a/src/fairseq2/utils/logging.py
+++ b/src/fairseq2/utils/logging.py
@@ -8,54 +8,45 @@ import logging
 import time
 from logging import DEBUG, INFO, FileHandler, Formatter, Handler, StreamHandler
 from pathlib import Path
-from typing import List
+from typing import List, Optional
 
-from fairseq2.gang import Gang
+from fairseq2.gang import is_coordinator_process
 
 
 def setup_logging(
-    log_dir: Path, gang: Gang, debug: bool = False, utc: bool = False
+    *,
+    log_file: Optional[Path] = None,
+    debug: bool = False,
+    utc_time: bool = False,
 ) -> None:
     """Set up logging for a training or eval job.
 
-    :param log_dir:
-        The log output directory.
-    :param gang:
-        The gang of the job.
+    :param log_file:
+        The file to which logs will be written.
     :param debug:
         If ``True``, sets the log level to `DEBUG`; otherwise, to `INFO`.
-    :param utc:
+    :param utc_time:
         If ``True``, logs dates and times in UTC.
     """
-    if gang.rank == 0:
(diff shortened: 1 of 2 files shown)


src/fairseq2/utils/logging.py
```diff
@@ -8,54 +8,45 @@
...
     :param debug:
         If ``True``, sets the log level to `DEBUG`; otherwise, to `INFO`.
-    :param utc:
+    :param utc_time:
         If ``True``, logs dates and times in UTC.
     """
-    if gang.rank == 0:
+    handlers: List[Handler] = [StreamHandler()]  # Log to stderr.
+
+    if log_file is not None and is_coordinator_process():
```
shagunsodhani, line 33: Could we add a flag to disable this behavior ? Having all ranks write to a (different) file is helpful when debugging errors. In the case of slurm jobs, we already have per-node error and log files so its not an issue but when debugging single node distributed jobs, it will be helpful to have the ability to log output from all ranks to files.

cbalioglu: @shagunsodhani How/with which "launcher" do you run your single-node distributed jobs? I typically use torchrun or submitit for local runs and both can capture and write stderr/stdout of running processes to a file. If there is a need beyond those two, I can certainly add it as an extra feature.

shagunsodhani: For the debug runs, the workflow is to request an interactive job and launch a job using torchrun. Maybe you are right that in this case, the output would be logged to the right place anyways (since the launched is still torchrun)

cbalioglu: Yes, with torchrun, you can start the ranks with `torchrun --log-dir /path/to/logs` and it will write stdout/stderr of all ranks to individual files within that directory. Similarly submitit with the "local" cluster writes stdout/stderr to the executor directory.

Note that the log file mentioned in this PR is only about recording the log statements, so it is a subset of stdout/stderr streams. It is just for convenience to do a "tail -f /path/to/train.log" to monitor the progress of the job. Any troubleshooting requires you to then delve in to the actual stdout/stderr outputs of the ranks.

shagunsodhani: > Yes, with torchrun, you can start the ranks with torchrun --log-dir /path/to/logs and it will write stdout/stderr of all ranks to individual files within that directory.

Thanks for the clarification. 

> Note that the log file mentioned in this PR is only about recording the log statements, so it is a subset of stdout/stderr streams

But this setting does affect all the logged statements correct ? (including the ones coming from 3rd party libs). So I felt an opt-out maybe useful but I do not have a usecase for that for now.

PR #246, opened 2023-12-29
Revise log_file handling in setup_logging

cbalioglu:
This PR revises (again) the `log_file` handling in `setup_logging()`. After @shagunsodhani's feedback and giving it more thought, I decided to roll back the last commit that used the log file only for rank 0. In this revision, `log_file` is expected to include a "rank" replacement field (e.g. /path/to/train_{rank}.log) which will then create an individual log file per rank (like before). Note that for most real-world use cases where a launcher is used (e.g. torchrun, submitit), `log_file` is not really needed since that functionality is already provided by the launchers. It is mostly meant to be used for jobs that are manually launched (e.g. for local debugs).

### DIFF
--- a/src/fairseq2/gang.py
+++ b/src/fairseq2/gang.py
@@ -384,8 +384,6 @@ def _get_int_from_env(var_name: str) -> Optional[int]:
         )
 
 
-def is_coordinator_process() -> int:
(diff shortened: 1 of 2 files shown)


shagunsodhani (approved):

shagunsodhani:
Thank you for incorporating the feedback!

PR #247, opened 2024-01-02
Improve the implementation of CheckpointManager

cbalioglu:
This PR improves the implementation of `CheckpointManager`:

1) Makes `save_checkpoint` idempotent by using a marker file instead of a global symbolic link
2) Related to the first point, supports non-distributed file systems (e.g. local debug runs, or AWS SageMaker S3-backed checkpointing)
3) Ensures that LOOKUP cache is handled correctly on NFS-backed checkpoint directories


cbalioglu:
cc @shagunsodhani

PR #254, opened 2024-01-06
Reformat log statements

cbalioglu:
A nit PR that reformats log statements in the code base. No behavioral change introduced.

### DIFF
--- a/src/fairseq2/optim/dynamic_loss_scaler.py
+++ b/src/fairseq2/optim/dynamic_loss_scaler.py
@@ -148,11 +148,8 @@ class DynamicLossScaler:
         if self._are_close(old_scale, new_scale):
             return LossScaleResult(old_scale, new_scale)
 
-        # fmt: off
         if new_scale > old_scale:
-            self._log(logging.INFO,
-                "No gradient overflow detected in the last %s step(s), increasing loss scale from %s to %s.", self.scale_window, old_scale, new_scale
-            )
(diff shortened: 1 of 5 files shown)


src/fairseq2/optim/dynamic_loss_scaler.py
```diff
@@ -148,36 +148,26 @@ def _update_scale(self) -> LossScaleResult:
         if self._are_close(old_scale, new_scale):
             return LossScaleResult(old_scale, new_scale)
 
-        # fmt: off
         if new_scale > old_scale:
-            self._log(logging.INFO,
-                "No gradient overflow detected in the last %s step(s), increasing loss scale from %s to %s.", self.scale_window, old_scale, new_scale
-            )
+            self._log_info("No gradient overflow detected in the last %s step(s), increasing loss scale from %s to %s.", self.scale_window, old_scale, new_scale)  # fmt: skip
```
shagunsodhani, line 152: nit: maybe we could use the f-strings here as well ?

cbalioglu: This is actually a deliberate choice (also common in logging APIs of other language stdlibs). The reason is that you do not know the effective log level while calling the log functions. If you use f-string, you will be always formatting your messages even if they are never logged. This is a minor, but common optimization (which can actually have an impact for hot code paths)

shagunsodhani: I did not know about this! Thank you!

cbalioglu: np!

PR #273, opened 2024-01-22
Add skip_long_examples to bucket_by_length

cbalioglu:
This PR adds a new `skip_long_examples` to `bucket_by_length` data pipeline op. As its name suggest, it skips examples longer than the specified value without raising an error. Logging is marked as TODO and will be handled in a future PR.


cbalioglu:
The CI failure is unrelated to this PR. Apparently ParquetDataset has a BC breaking change where `validate_schema` parameter is removed from its constructor. Gonna "fix" in a follow-up PR.

PR #281, opened 2024-01-26
Introduce MetricRecorder

cbalioglu:
This PR:
- introduces a new `MetricRecorder` interface and `LogMetricRecorder` and `TensorBoardRecorder` classes.
- revises how we format metric values and extracts the relevant functions from `MetricBag`.
- performs a couple nit readability updates in other parts of the code.

### DIFF
--- a/src/fairseq2/metrics.py
+++ b/src/fairseq2/metrics.py
@@ -4,12 +4,20 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, Dict, Mapping, Optional
+from __future__ import annotations
+
+import logging
+from abc import ABC, abstractmethod
+from functools import partial
+from logging import Logger
+from pathlib import Path
+from typing import Any, Callable, Dict, Mapping, Optional, Tuple, final
 
 from torcheval.metrics import Metric
 from torcheval.metrics.toolkit import sync_and_compute_collection
 
 from fairseq2.gang import Gang
+from fairseq2.typing import finaloverride
 
 
 class MetricBag:
@@ -95,14 +103,6 @@ class MetricBag:
         """Process metric ``values``."""
         pass
 
-    def format_metric_values(self, values: Dict[str, Any]) -> str:
-        """Format metric ``values`` to print to stdout or stderr."""
-        return format_metric_values(values, self)
-
-    def format_metric_value(self, name: str, value: Any) -> Optional[str]:
-        """Format the value of metric ``name`` to print to stdout or stderr."""
-        return None
-
     def state_dict(self) -> Dict[str, Any]:
         state_dict = {}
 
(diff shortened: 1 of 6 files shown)


src/fairseq2/metrics.py
```diff
@@ -158,24 +158,218 @@ def sync_and_compute_metrics(*bags: MetricBag) -> Optional[Dict[str, Any]]:
...
+format_as_loss = partial(format_as_float, decimal=3)
+"""Format metric ``value`` as training loss."""
+
+
+_metric_formatters: Dict[str, Tuple[str, Callable[[Any], str]]] = {
+    # fmt: off
+    "batch_size":          ("Batch Size",                format_as_int),
+    "elapsed_time":        ("Elapsed Time",              format_as_seconds),
+    "elements_per_batch":  ("Elements per Batch",        format_as_int),
+    "elements_per_second": ("Elements per Second",       format_as_int),
```
shagunsodhani, line 201: this can be a float right ? e.g. if we process 1 element per 400 ms ?

src/fairseq2/metrics.py
```diff
@@ -158,24 +158,218 @@ def sync_and_compute_metrics(*bags: MetricBag) -> Optional[Dict[str, Any]]:
...
+        :param values:
+            The metric values.
+        :param step_nr:
+            The number of the run step.
+        :param flush:
+            If ``True``, flushes any intermediate buffer after recording.
+        """
+
+    def close(self) -> None:
+        """Close the recorder."""
```
shagunsodhani, line 258: nit: shouldnt this have a `pass` statement or something like that

src/fairseq2/metrics.py
cbalioglu: These formatters are meant to be used to report progress via stdout/stderr. For values less that 1.0, we just show `<1`. For instance, see the elapsed time in this log output:

```
Valid:eng-yor Metrics (step 200) - Loss: 10.645 | Entropy Loss: 15.358 | Batch Size: 248 | Elements per Batch: 11,304 | Elements per Second: 46,781 | Number of Source Elements: 33,447 | Number of Target Elements: 45,217 | Elapsed Time: <1s
```

src/fairseq2/metrics.py
cbalioglu: It does not necessarily have to have, syntactically this is correct, but to increase the readability I added a "redundant" `pass` in #287.

src/fairseq2/metrics.py
shagunsodhani: Thank you for the clarification!

src/fairseq2/metrics.py
shagunsodhani: Thank you for the clarification!

src/fairseq2/metrics.py
cbalioglu: No problem!

PR #284, opened 2024-01-26
Support returning (final-layer) features

### LINKED ISSUE #283: Support feature extraction in w2v2
**Is your feature request related to a problem? Please describe:**
After pretraining Wav2Vec2, users should be able to fine-tune by using it as a feature extractor. Right now there's no method for doing that (I think?). 

**Describe the solution you would like:**
1. I think there should be a new method `extract_features` for the model that calls `self.run_frontend` and then `self.encoder`.
2. Then `forward` will call this new method and have an additional new argument `return_features` defaulting to `False`. If this argument is `True`, `forward` will return the output of the transformer layers (and optionally convolutional feature extractor) in addition to the `Wav2Vec2` output object that is the output of `self.quantize_and_contrast`. 

**Describe the alternatives you have considered:**
Don't modify `forward`. Disadvantage: harder to do feature extraction and fine-tuning in a DDP context since DDP wraps `model.forward` but not other methods (I think?). 

**Additional Context:**
Add any other context about the feature request here.

mshvartsman:

Fixes #283. Probably could return something more strongly typed, optionally all the layers, etc, but this is a start. 

None, default is same return as previously. 

**Check list:**


src/fairseq2/models/wav2vec2/model.py
```diff
@@ -113,7 +113,9 @@ def __init__(
         self.logit_temp = logit_temp
         self.diversity_loss_weight = diversity_loss_weight
 
-    def forward(self, batch: SequenceBatch) -> Wav2Vec2Output:
+    def forward(
+        self, batch: SequenceBatch, return_features: bool = False
+    ) -> Wav2Vec2Output | Tuple[Wav2Vec2Output, torch.Tensor]:
```
shagunsodhani, line 118: This makes sense as a standalone feature (i.e. we should be able to return intermediate features as part of the forward call). But for @mshvartsman 's usecase (of finetuning a pretrained wav2vec2 model), this approach, while works, is inefficient since we are done extra computations. I would suggest the follwing:

1. We add a `extract_feature` method to extract just the features (which solves @mshvartsman 's usecase efficiently) and
2. We add the support for returning the intermediate features to `Wav2Vec2Output` i.e. `return_features` flag is propogated to (and handled in) `self.quantize_and_contrast`

mshvartsman: The extra compute is because we `quantize_and_contrast` even if we want features only? Equivalent solution would be to change `return_features` flag to `features_only` flag and then return `Wav2Vec2Output | torch.Tensor`, thoughts? Agree that `extract_features` should be its own method though.

mshvartsman: You mean extra compute since we `quantize_and_contrast` even if we only want the features? Fair point, but I think we have a choice here: 
1. If I want logits or both features and logits to be the cheap option, current behavior is fine. Features-only (e.g. fine-tune workflow) requires extra compute. 
2. If we want logits or features but not both to be cheap, we can have `features_only` argument and return `Wav2Vec2Output | torch.Tensor`, but then having both (e.g. to record both val loss and representation metrics during SSL, which is also a valid use case) requires running the encoder twice. 
3. If we want all options to be cheap, we need to return not just features but `padding_mask, targets, temporal_mask` from `extract_features` so we can do whatever we want with the features and then pass everything down to `quantize_and_contrast`. But then I think downstream code will often need to call `features, _, _, _ = extract_features(...)` which is an awkward call unless we want it to also have a `features_only` argument. 

Thoughts? My lean is [3] but all of them feel a bit awkward.

shagunsodhani: The output of the `extract_features` function can be a structured object, just like the output of the `forward` call](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/models/wav2vec2/model.py#L116). So the `forward` call can also use the `extract_features` function. Regarding `extract_features` having a `features_only` argument, I think it is reasonable.

mshvartsman: I think with a structured output of `extract_features` we don't need to worry about `features_only` since the compute cost doesn't change and it's easy to ignore the extra stuff.

shagunsodhani (approved):
looks good to me. What do you think @cbalioglu

cbalioglu:
Hey folks,

Sorry just catching up. Do you mind showing an example use case where you would need a call to `extract_features()` or the new `encoder_output` attribute in `Wav2Vec2Output`? Do you need `targets` or `temporal_mask` in your finetuning job? In more conventional wav2vec finetuning jobs, the flow looks more like this (semi-pseudo):

```python
w2v_model = load_w2v2_model("my_pretrained_w2v2")

# A vanilla transformer model (e.g. an encoder-decoder or an encoder with an output projection layer).
# e.g. https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/models/transformer/model.py#L24
model_to_finetune = create_foo_model(...)

# Assign w2v encoder and frontend to the vanilla transformer.
model_to_finetune.encoder_frontend = w2v_model.encoder_frontend
model_to_finetune.encoder = w2v_model.encoder

finetune(model_to_finetune)
```

If this is for some reason not sufficient for your use case, it would be great if you could share some more insight.

mshvartsman:
Sorry for missing this @cbalioglu! In the flow you describe, I have to know something about how the internals of the fs2 model work, and make a new model that expects to have an `encoder` attribute which we then assign to. But that's not always the case. For example: 
- If I want to compute intermediate metrics on my representations (e.g. RankMe) during validation, I need to call `model(X)` once to get validation loss, and then separately manually call `model.encoder(...)` and `model.frontend(...)` to get the features to compute metrics on (i.e. I pay for two forward passes). 
- If I want to pass my model onto collaborators who want to use it as a feature extractor in some downstream thing then I basically need to make a `Wav2Vec2FeatureExtractor` object that I can send to them which obeys the encoder/frontend semantics, or have them implement this. 

(@shagunsodhani might have more context)

cbalioglu:
@mshvartsman Thanks for the explanation! I will be in the LA office tomorrow and will have more time to check out this PR again.

shagunsodhani:
Another usecase will be multi-task learning where one of the losses is the SSL loss and the other loss requires access to the output of the encoder. Similar to Mike's first example. we will need to do two forward passes in the current setup.

cbalioglu:
@mshvartsman 

> If I want to compute intermediate metrics on my representations (e.g. RankMe) during validation, I need to call model(X) once to get validation loss, and then separately manually call model.encoder(...) and model.frontend(...) to get the features to compute metrics on (i.e. I pay for two forward passes).

What concrete specific information is needed for your intermediate metric computation (e.g. RankMe)? From the PR, I see that `encoder_output` and (non-quantized) `targets` are the two missing pieces right now that we can certainly add to `Wav2Vec2Output` (you already have `encoder_output` in this PR). Would that be enough to avoid two forward calls in your use case?

> If I want to pass my model onto collaborators who want to use it as a feature extractor in some downstream thing then I basically need to make a Wav2Vec2FeatureExtractor object that I can send to them which obeys the encoder/frontend semantics, or have them implement this.

I think the "right" approach to this would be for them to "plug" wav2vec2 frontend and encoder to their model architecture (your latter option). This is essentially what we did in Seamless models which follow the fairseq2 model structure where we have frontends and encoders/decoders in their architecture. However if you want your pretrained wav2vec2 interoperate with models implemented using a different approach (i.e. not following the frontend/encoder split), we can certainly have a convenience `Wav2Vec2FeatureExtractor` module used like:

```python
w2v2_model = load_w2v2_model(...)

w2v2_feat_extractor = Wav2Vec2FeatureExtractor(w2v2_model)
```

Pretty much this is your first suggestion incorporated as a convenience type in fairseq2.

@shagunsodhani 

> Another usecase will be multi-task learning where one of the losses is the SSL loss and the other loss requires access to the output of the encoder. Similar to Mike's first example. we will need to do two forward passes in the current setup.

Can't we resolve this by always returning `encoder_output` as part of `Wav2Vec2Output`?

shagunsodhani:
> @shagunsodhani
> 
> > Another usecase will be multi-task learning where one of the losses is the SSL loss and the other loss requires access to the output of the encoder. Similar to Mike's first example. we will need to do two forward passes in the current setup.
> 
> Can't we resolve this by always returning `encoder_output` as part of `Wav2Vec2Output`?

Depending on how we do multitask learning, this may be in-efficient. e.g. one variant (of multitask) learning is where we compute multiple losses on the same batch of data. In that case, returning `encoder_output` is sufficient. A different variant is where we alternate between the two tasks (1 batch for first task, next batch for second task, which only needs encoder output, etc). In that case, we will be doing extra computation for the second task. 

Another scenario is where we finetune a small mlp model, on top of the representations from the SSL model, during validation. This is similar to Mike's example of RankMe metric, with the difference that we do not care about the SSL loss on the validation dataset and so doing the full forward is un-necessary. The broader motivation for doing this light-weight supervised training is to evaluate the model's performance on a proxy task that is "similar" to the actual downstream task. In our experience, relying on just the SSL loss during validation is not sufficient.

cbalioglu (commented):
Mostly nit comments and one feedback about `encoder_output`.

cbalioglu:
@mshvartsman @shagunsodhani thanks a lot for the context! Based on your use case examples, it makes sense to have this PR. I just reviewed the code. I have one suggestion about `encoder_output` (see the review comments) and a few nit comments. I think once we address them, it will be in a good shape to merge.

src/fairseq2/models/wav2vec2/model.py
```diff
@@ -125,7 +126,23 @@ def forward(self, batch: SequenceBatch) -> Wav2Vec2Output:
...
+        )
+
+    def forward(self, batch: SequenceBatch) -> Wav2Vec2Output:
+        """
+        :param batch:
+            The batch of sequences to process.
+        """
+        feats = self.extract_features(batch)
+
+        return self.quantize_and_contrast(
```
mshvartsman, line 140: @cbalioglu with the removal of `return_features`, this can just take the `Wav2Vec2Features` object directly, or keep the individual args as it was previously. Do you have a preference on this front?

mshvartsman:
> What concrete specific information is needed for your intermediate metric computation (e.g. RankMe)?

In the shorter term, the encoder output is sufficient (and you're right that if that's all we need, we're already computing and storing it and should always return it). Eventually we may want non-final-layer features (we've found metrics on an average of transformer layers to be a better predictor than the final layer only, and at least talking to Wei-Ning it sounds like they've found a weighted sum of transformer layers to sometimes do better than just the final layer in downstream tasks). At that point we might want to revisit the `return_features` arg since intermediate layer outputs aren't returned or held in memory by default (IIRC this is what fs1 did -- return all layer results based on a flag). 

> However if you want your pretrained wav2vec2 interoperate with models implemented using a different approach (i.e. not following the frontend/encoder split), we can certainly have a convenience Wav2Vec2FeatureExtractor module 

This seems sensible, and is something that we can either have in our codebase or eventually upstream to fs2. This PR enables that I believe.

cbalioglu:
@mshvartsman I am traveling right now, but in case you haven’t seen it. The way you can access intermediate layer outputs is via “hooks” in fairseq2. This avoids returning them (i.e. keeping them in memory) unless necessary. You can check out the API here: https://github.com/facebookresearch/fairseq2/blob/779c6cc12d0da51fbb3abff5bba244a35b37f6ce/src/fairseq2/nn/transformer/encoder.py#L68

we can certainly improve it though in case it misses some of your future requirements.

src/fairseq2/models/wav2vec2/model.py
```diff
@@ -113,10 +113,15 @@ def __init__(
         self.logit_temp = logit_temp
         self.diversity_loss_weight = diversity_loss_weight
 
-    def forward(self, batch: SequenceBatch) -> Wav2Vec2Output:
-        """
+    def extract_features(self, batch: SequenceBatch) -> Wav2Vec2Features:
+        """Extract features from the input sequences.
```
cbalioglu, line 117: very nit: Please have a blank line before `:param:` for sphinx to parse it correctly.

src/fairseq2/models/wav2vec2/model.py
```diff
@@ -113,10 +113,15 @@ def __init__(
...
         :param batch:
             The batch of sequences to process.
+
+        :returns:
+            - A `Wav2Vec2Features` object, consisting of the encoder output, targets
```
cbalioglu, line 122: Again, very nit. `- ` causes a bullet point in sphinx. You can start the sentence with "A `Wav2Vec2Features` object, .."

src/fairseq2/models/wav2vec2/model.py
cbalioglu: I think it makes sense to pass the `feats` object directly to `quantize_and_contrast`. It will make the code less noisy.

cbalioglu (commented):
Two very nit comments. I also agree with your suggestion to pass `feats` instead of individual attributes to `quantize_and_constrast()`. Looks like there is also a minor linting issue. Let me know once you address them, and I can merge it right away.

src/fairseq2/models/wav2vec2/model.py
mshvartsman: Ah, was matching how the `run_frontend` formatting was, sorry.

src/fairseq2/models/w2vbert/model.py
```diff
@@ -113,9 +118,10 @@ def hook(
 
         assert w2v2_layer_output is not None
 
-        w2v2_output = self.w2v2_model.quantize_and_contrast(
-            w2v2_layer_output, targets, temporal_mask
+        w2v2_features = Wav2Vec2Features(
+            encoder_output=encoder_output, targets=targets, temporal_mask=temporal_mask
```
cbalioglu, line 122: A subtle bug; we should pass `encoder_output=w2v2_layer_output` here.

cbalioglu (commented):
Ok, one last subtle (but important) bug. Once fixed, good to go.

cbalioglu (approved):
LGTM!

PR #330, opened 2024-02-14
Introduce load_state_dict

cbalioglu:
This PR fixes a subtle, but important bug in PyTorch's `load_state_dict` implementation where keys in `state_dict` that correspond to descendant modules that are set to `None` in `module` are simply ignored instead of being marked as "unexpected". This bug bit us several times last year while working on porting fairseq models to fairseq2. Here we introduce a `load_state_dict()` version that correctly checks *all* keys in `state_dict`.


shagunsodhani (commented):
Very educating!

PR #332, opened 2024-02-14
Introduce NllbDataset

cbalioglu:
This PR introduces the NLLB dataset implementation. The public asset card will follow up later before v0.3 release.

### DIFF
--- a/src/fairseq2/datasets/nllb.py
+++ b/src/fairseq2/datasets/nllb.py
@@ -1,0 +1,409 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from __future__ import annotations
+
+import logging
+from functools import partial
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Sequence, Tuple, TypedDict, cast, final
+
+from torch import Tensor
+
+from fairseq2.assets import AssetCard, AssetCardError
+from fairseq2.assets import asset_store as default_asset_store
+from fairseq2.assets import download_manager as default_download_manager
+from fairseq2.data import (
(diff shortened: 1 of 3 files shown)


src/fairseq2/datasets/nllb.py
```diff
@@ -0,0 +1,409 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from __future__ import annotations
+
+import logging
+from functools import partial
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Sequence, Tuple, TypedDict, cast, final
```
shagunsodhani, line 12: nit but we can avoid importing `Dict` etc (since you are importing `annotations`) and use `dict` instead e.g.

src/fairseq2/datasets/nllb.py
```diff
@@ -0,0 +1,409 @@
...
+    """Represents a line read from a language pair file."""
+
+    split: CString
+    lang_pair: CString
+    line_number: int
+    source_tokens: Tensor
+    target_tokens: Tensor
+
+
+class _NllbDataPipelineFactory:
```
shagunsodhani, line 161: Does the `_` in the name added to suggest these are internal classes ?

cbalioglu: Yes, any module-scope type, function, or variable that is prefixed with an underscore indicates "internal". For class method and variables depending on the use case it might mean either "private" or "protected". If it is "protected", I typically leave a comment like `# protected`.

src/fairseq2/datasets/nllb.py
cbalioglu: I just want to be consistent throughout the code base. Since we still support Python 3.8 where some of the newer type annotation syntax is no supported, I prefer to keep everything "compatible" with Python 3.8.

PR #334, opened 2024-02-14
Support heuristic-based scale window computation

cbalioglu:
This PR introduces the same heuristic method as in fairseq to determine the optimum scale window in `DynamicLossScaler`.

### DIFF
--- a/src/fairseq2/optim/dynamic_loss_scaler.py
+++ b/src/fairseq2/optim/dynamic_loss_scaler.py
@@ -42,8 +42,9 @@ class DynamicLossScaler:
         *,
         init_scale: float = 2.0**15,
         scale_factor: float = 2.0,
-        scale_window: int = 2000,
+        scale_window: Optional[int] = None,
         min_scale: float = 0.0,
+        gradient_accumulation: int = 1,
         enabled: bool = True,
     ) -> None:
         """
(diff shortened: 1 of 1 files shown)


src/fairseq2/optim/dynamic_loss_scaler.py
```diff
@@ -77,6 +83,12 @@ def __init__(
                             f"The parameters held by `optimizer` must be on a 'cuda' device, but at least one parameter is on a '{param.device.type}' device."
                         )
 
+        if scale_window is None:
+            # This is the formula that we use in fairseq.
+            scale_window = max(int(2**14 / gang.size / gradient_accumulation), 1)
```
shagunsodhani, line 88: interesting! Is there any pointers to why this is good - out of curiosity :)

cbalioglu: Good point :) I was also curious about that and traced it down to a fairinternal PR back from 2018 authored by Sergey. There is no mention why specifically they (Sergey and Myle) picked 2**14 and not some other value. I assume they experimented with different values and decided that this was the optimal one. To be consistent with fairseq, I just copied it over to fairseq2. I can ask Sergey next time I see him :)

shagunsodhani: Thank you for the context!

cbalioglu: np!

PR #340, opened 2024-02-16
Enable different encoder and decoder sizes in cross-attention

avidale:

This PR allows creating an attention layer with different dimensions of the inputs and outputs. 
This is useful for encoder-decoder architectures with different dimensionalities (including low-dimensional SONAR encoders and decoders, as an example). 

No breaking changes. 

**Check list:**


src/fairseq2/nn/transformer/multihead_attention.py
```diff
@@ -187,6 +187,7 @@ def __init__(
         model_dim: int,
         num_heads: int,
         *,
+        input_dim: Optional[int] = None,
```
cbalioglu, line 190: I don't think `input_dim` is the right name for this parameter. Pedantically "input" to MHA is the query. K/V values are more like the "state dictionary" to attend against. How about naming this `kv_model_dim` or `kv_dim`?

cbalioglu (commented):
Thanks for the PR and including CHANGELOG and tests as well. Just left a rather nit comment about the naming of this parameter. Once we align on that, I think it will be good to go.

src/fairseq2/nn/transformer/multihead_attention.py
avidale: `kv_dim` sounds just right!

cbalioglu (approved):
Nice! Thanks again for your contribution! Looking forward to future ones! :)

PR #348, opened 2024-02-23
Add force flag to logging setup

jc-audet:
**What does this PR do?**
Logging config is not applied if the config has already been defined in a previous import. This makes `setup_logging` useless. The `force` flag can be added to the basicConfig to overwrite what has been defined before.

I don't think it will break anything. Works locally. Default value set to previous behavior (False).

**Check list:**

### DIFF
--- a/src/fairseq2/utils/logging.py
+++ b/src/fairseq2/utils/logging.py
@@ -18,6 +18,7 @@ def setup_logging(
     log_file: Optional[Path] = None,
     debug: bool = False,
     utc_time: bool = False,
+    force: bool = False,
 ) -> None:
     """Set up logging for a training or eval job.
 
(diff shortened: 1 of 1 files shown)


shagunsodhani (commented):
curious if we can handle it by just calling this function sooner or later ?

jc-audet:
> curious if we can handle it by just calling this function sooner or later ?

I'm interpreting this as you would want to run this in the import section of the code. In that case I think not always, if you want to define the the log file name programmatically to something non-trivial, then you would need to wait to within some main function before setting up the logger.

I think just the fact that the flag exists in the logging package proof enough for that though.

cbalioglu (commented):
Thanks! I was thinking about this a while ago as well. Please make sure that you run `black .` to fix the lint issue. Once the CI is green, will merge it.

cbalioglu (approved):

PR #374, opened 2024-03-07
Make checkpoint manager return the model path

avidale:
After training a model, I would like to automatically create a model card for it, which includes the path to the model checkpoint.
Currently, to determine this path, I use `checkpoint_manager._checkpoint_dir / f"step_{n_steps}" / "model.pt"`, which is not great for several reasons:
- It accesses a private attribute
- It makes an assumption that all the training steps have completed (so it wouldn't work e.g. in case of early stopping).
- It uses the hardcoded `"model.pt"` filename; if ever CheckpointManager changes it, my code will break.

To avoid all these inconveniences, I make `CheckpointManager` explicitly provide the model path for its last checkpoint.

**Check list:**

Regarding the tests: currently, they don't seem to be any for the `CheckpointManager`. Do I need to create one?

### DIFF
--- a/src/fairseq2/checkpoint.py
+++ b/src/fairseq2/checkpoint.py
@@ -554,7 +554,7 @@ class FileCheckpointManager(CheckpointManager):
         if step_nr is None:
             step_nr = self.get_last_step_number()
         if step_nr is None:
-            return
+            return None
         return self._checkpoint_dir.joinpath(f"step_{step_nr}/model.pt")
 
     def _iter_step_numbers(self, with_model: bool) -> Iterator[int]:
--- a/src/fairseq2/checkpoint.py
+++ b/src/fairseq2/checkpoint.py
@@ -154,16 +154,6 @@ class CheckpointManager(ABC):
             If ``True``, only considers checkpoints with a model.
         """
 
-    @abstractmethod
-    def get_model_checkpoint_path(
-        self, step_nr: Optional[int] = None
-    ) -> Optional[Path]:
-        """
-        Return the path for the model checkpoint (by default, the last one).
-        :param step_nr: 
-            The step for which to load the model. If ``None``, then the last checkpoint is loaded.
-        """
-
 
 @final
 class FileCheckpointManager(CheckpointManager):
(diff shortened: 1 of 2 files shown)


src/fairseq2/checkpoint.py
```diff
@@ -154,6 +154,16 @@ def get_last_step_number(self, *, with_model: bool = False) -> Optional[int]:
             If ``True``, only considers checkpoints with a model.
         """
 
+    @abstractmethod
+    def get_model_checkpoint_path(
```
cbalioglu, line 158: It would be nice if we could keep the base `CheckpointManager` "file-system agnostic" since in the future we might have other checkpoint manager types that do not use a file system. Would exposing this function only in `FileCheckpointManager` work for your use case?

src/fairseq2/checkpoint.py
```diff
@@ -534,6 +547,16 @@ def get_last_step_number(self, *, with_model: bool = False) -> Optional[int]:
 
         return None
 
+    @override
+    def get_model_checkpoint_path(
+        self, step_nr: Optional[int] = None
+    ) -> Optional[Path]:
+        if step_nr is None:
+            step_nr = self.get_last_step_number()
```
cbalioglu, line 555: A subtle bug, this should ideally be `self.get_last_step_number(with_model=True)`.

src/fairseq2/checkpoint.py
```diff
@@ -456,7 +466,10 @@ def save_consolidated_model(self, step_nr: int, model: Module) -> None:
     def load_model(
         self, step_nr: int, out: Module, *, device: Optional[Device] = None
     ) -> None:
-        model_file = self._checkpoint_dir.joinpath(f"step_{step_nr}/model.pt")
+        model_file = self.get_model_checkpoint_path(step_nr=step_nr)
+        if model_file is None:
+            # This should never happen, because if the step_nr is provided, model_file is always non-empty.
```
cbalioglu, line 471: This is not entirely correct. For FSDP checkpoints for instance, a checkpoint might have no consolidated model. So I would omit this comment and change the exception message to "Training step {step_number} has no saved model." similar to line 478.

cbalioglu (commented):
Thanks for the PR. Left three nit comments.

src/fairseq2/checkpoint.py
avidale: Yes, it would!
I'll remove it from the base class.

cbalioglu (approved):
LGTM!

PR #388, opened 2024-03-18
More temporary compat APIs

cbalioglu:
Yet more compat APIs that we will hopefully remove after moving internal teams to v0.3


antoine-tran:
thanks

antoine-tran (approved):

PR #415, opened 2024-03-29
Implement property total_batch_size of all ranks in DataPipelineReader.

kauterry:

Fixes #{issue number}

**Check list:**

### DIFF
--- a/src/fairseq2/datasets/data_reader.py
+++ b/src/fairseq2/datasets/data_reader.py
@@ -11,7 +11,7 @@ from typing import Any, Dict, Iterator, List, Mapping, TypeVar, final
 from typing_extensions import Self
 
 from fairseq2.data import DataPipeline
-from fairseq2.datasets.utils import _all_eod
+from fairseq2.datasets.utils import _all_eod, _total_batch_size
 from fairseq2.gang import Gang
 from fairseq2.models import Batch
 from fairseq2.typing import override
@@ -87,6 +87,7 @@ class DataPipelineReader(DataReader[BatchT]):
         self._num_accumulate = num_accumulate
         self._sync_batches = sync_batches
         self._eod = False
+        self._total_batch_size = 0
 
     @override
     def __iter__(self) -> Self:
@@ -106,6 +107,8 @@ class DataPipelineReader(DataReader[BatchT]):
                 break
 
             batches.append(batch)
+
+        self._total_batch_size = _total_batch_size(batches, self._gang)
 
         self._eod = len(batches) != self._num_accumulate
 
@@ -133,3 +136,7 @@ class DataPipelineReader(DataReader[BatchT]):
         self._eod = False
 
         self._pipeline.load_state_dict(state_dict)
+
+    @property
+    def total_batch_size(self):
+        return self._total_batch_size
--- a/src/fairseq2/datasets/data_reader.py
+++ b/src/fairseq2/datasets/data_reader.py
@@ -11,7 +11,7 @@ from typing import Any, Dict, Iterator, List, Mapping, Tuple, TypeVar, final
 from typing_extensions import Self
 
 from fairseq2.data import DataPipeline
-from fairseq2.datasets.utils import _all_batch_sizes
+from fairseq2.datasets.utils import _total_batch_size_and_eod
 from fairseq2.gang import Gang
 from fairseq2.models import Batch
 from fairseq2.typing import override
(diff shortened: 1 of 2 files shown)


src/fairseq2/datasets/data_reader.py
```diff
@@ -104,6 +105,8 @@ def __next__(self) -> List[BatchT]:
 
             batches.append(batch)
 
+        self._total_batch_size = _total_batch_size(batches, self._gang)
```
kauterry, line 108: @cbalioglu how do you recommend I deal with this mypy issue?

src/fairseq2/datasets/data_reader.py
```diff
@@ -130,3 +133,7 @@ def load_state_dict(self, state_dict: Mapping[str, Any]) -> None:
         self._eod = False
 
         self._pipeline.load_state_dict(state_dict)
+
+    @property
+    def total_batch_size(self):
```
kauterry, line 138: will fix this in my revision.

src/fairseq2/datasets/utils.py
```diff
@@ -30,3 +34,22 @@ def _all_eod(eod: bool, gang: Gang, logger: Logger) -> bool:
...
+
+    total_bsz = 0
+    if gang.size == 1:
+        for b in batches:
+            total_bsz += b.source_seqs.size(0)
+        return total_bsz
+
+    for b in batches:
+        batch_sizes = torch.empty((gang.size,), device=gang.device, dtype=torch.int64)
+        gang.all_gather(
```
shagunsodhani, line 50: Wouldnt an all_reduce be more efficient than an all_gather here ?

src/fairseq2/datasets/data_reader.py
cbalioglu: Please rebase this PR on top of latest main. I just merged #416 that introduces a `Batch` interface with a `batch_size` property.

cbalioglu (commented):
Two points:

(1) First we have to think about the API. I don't think that exposing the batch size information as a property is the right way as it indicates that it is the "state" of the object, in this case the `DataReader` instance, which is not true. Ideally we should return it as part of the `__next__()` call, i.e. `Tuple[int, List[BatchT]]` (meaning a tuple where the first element is the total size of *all* batches from *all* ranks and the second element is the list of batches for this current rank).

(2) With this implementation, we have to perform two collective calls per iteration. Ideally we should handle everything in a single all gather or all reduce call. If you see #416, I actually changed the name `sync_eod` to `sync_batches` to give you a bit clue. Instead of collecting a boolean vector from all processes, we can collect the size of their batches, which will help to determine both the total batch size as well as eod (if bsz==0 for a rank, means it reached eod).

kauterry:
> Two points:
> 
> (1) First we have to think about the API. I don't think that exposing the batch size information as a property is the right way as it indicates that it is the "state" of the object, in this case the `DataReader` instance, which is not true. Ideally we should return it as part of the `__next__()` call, i.e. `Tuple[int, List[BatchT]]` (meaning a tuple where the first element is the total size of _all_ batches from _all_ ranks and the second element is the list of batches for this current rank).
> 
> (2) With this implementation, we have to perform two collective calls per iteration. Ideally we should handle everything in a single all gather or all reduce call. If you see #416, I actually changed the name `sync_eod` to `sync_batches` to give you a bit clue. Instead of collecting a boolean vector from all processes, we can collect the size of their batches, which will help to determine both the total batch size as well as eod (if bsz==0 for a rank, means it reached eod).

Totally agree with (1). (2) makes sense too but I have a question. What about the case when we don't need to sync_batches? Do we just compute total batch size as num_accumulate * gang.size * batches[0].batch_size?

src/fairseq2/datasets/utils.py
```diff
@@ -6,27 +6,40 @@
...
+    total_bsz = 0
+    if gang.size == 1:
+        for batch in batches:
+            total_bsz += batch.batch_size
+        return total_bsz, total_bsz == 0
+
+    eod = False
+    for batch in batches:
+        batch_sizes = torch.empty((gang.size,), device=gang.device, dtype=torch.int64)
+        gang.all_gather(batch_sizes, torch.tensor(batch.batch_size, device=gang.device))
```
cbalioglu, line 33: This causes to call `all_gather()` `num_accumulate` times which is not efficient. Ideally you should call it just one with `total_bsz` of each rank.

src/fairseq2/datasets/data_reader.py
```diff
@@ -107,12 +107,17 @@ def __next__(self) -> Tuple[int, List[BatchT]]:
...
-        if self._sync_batches:
-            total_batch_size, self._eod = _all_batch_sizes(batches, self._gang, logger)
+        if self._sync_batches and self._gang.size > 1:
+            total_batch_size, self._eod = _total_batch_size_and_eod(
+                batches, self._eod, self._gang, logger
+            )
         else:
-            total_batch_size = self._gang.size * self._num_accumulate * batch.batch_size
-            self._eod = len(batches) != self._num_accumulate
+            total_batch_size = self._gang.size * sum(
```
shagunsodhani, line 118: This statement, while correct, is a little confusing but it implies we might be double counting the batch size due to use of  both`self._gang_size`  and `sum(...batches)`. This is correct since `self._gang_size` is 1 for this codepath (I assume 0 is not a valid value). If that is the case, we could consider simplifying this statement by dropping `self._gang.size`

kauterry: `self._gang_size` is not forced to be 1 for this codepath, it can be > 1 and `self._sync_batches` can be False. In that can all ranks have the same number of batches which is `sum(batch.batch_size for batch in batches)`, so we do have to multiply by `self._gang_size`. To clarify, when we iterate over `batches` we iterate over `range(self._num_accumulate)` (due to gradient accumulation), so this for loop doesn't take into account the different ranks.

shagunsodhani: Thank you for the context!

cbalioglu (approved):

PR #419, opened 2024-03-31
Further nit improvements to AdamW and DynamicLossScaler

cbalioglu:
This PR performs some nit improvements to `AdamW` and `DynamicLossScaler`.

### DIFF
--- a/src/fairseq2/optim/dynamic_loss_scaler.py
+++ b/src/fairseq2/optim/dynamic_loss_scaler.py
@@ -72,27 +72,32 @@ class DynamicLossScaler:
         if enabled:
             for group in optimizer.param_groups:
                 for param in group["params"]:
-                    if param.dtype != torch.float16:
+                    if param.dtype != torch.float32 and param.dtype != torch.float16:
                         raise ValueError(
-                            f"The parameters held by `optimizer` must be of type `torch.float16`, but at least one parameter is of type `{param.dtype}`."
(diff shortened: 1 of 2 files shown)


src/fairseq2/optim/dynamic_loss_scaler.py
```diff
@@ -72,27 +72,32 @@ def __init__(
         if enabled:
             for group in optimizer.param_groups:
                 for param in group["params"]:
-                    if param.dtype != torch.float16:
+                    if param.dtype != torch.float32 and param.dtype != torch.float16:
                         raise ValueError(
-                            f"The parameters held by `optimizer` must be of type `torch.float16`, but at least one parameter is of type `{param.dtype}`."
+                            f"The parameters held by `optimizer` must be of type `torch.float32` or `torch.float16`, but at least one parameter is of type `{param.dtype}` instead."
```
shagunsodhani, line 77: Curious if there are any plans to add support for bf16 as well.

cbalioglu: fp16 has only 5 bits for its exponent (i.e. 10 bits for its mantissa), while bf16 has 8 bits. This means it has less precision than fp16, but can represent a wider range of numbers (which is typically more important for DNNs). For that reason, loss scaling is not really necessary for bf16.

shagunsodhani: Agreed. But since the PR adds support for the dtypes to be `fp32` along with `fp16`, I assumed some internal codepath goes through dynamic loss scalar for fp32 (hence asked for support for `bf16`).

PR #430, opened 2024-04-02
Introduce normalize_gradients to DDP

cbalioglu:
This PR introduces a new `normalize_gradients` parameter to `to_ddp()` and also makes a nit improvement to bucketing logic.


kauterry:
Can you elaborate a little bit what `normalize_gradients` is doing? Effectively why do we have to multiply the gradients by `self._gang.size / total_batch_size` rather than `1.0 / total_batch_size`?

cbalioglu:
I think we resolved this question offline. The internal recipe was just not updated with the change in this PR. We divide the gradients by `1.0 / total_batch_size` before the optimizer step.

PR #432, opened 2024-04-03
Introduce LogWriter

cbalioglu:
This PR introduces the helper `LogWriter` class that uses modern `format()` strings instead of old format string style of `logging.Logger`.


kauterry:
Why do you prefer this over f-strings?

PR #438, opened 2024-04-03
Compile scaled_dot_product_attention

cbalioglu:
With this PR, we now by default `torch.compile()` `torch.nn.functional.scaled_dot_product_attention` in `TorchSDPA`. This brings significant speed up for batches with padded inputs, which is typically the case for seq2seq models. I will run further benchmarks for non-padded inputs and see how well flash and memory_efficient attentions perform under `torch.compile()`.

### DIFF
--- a/src/fairseq2/nn/transformer/attention.py
+++ b/src/fairseq2/nn/transformer/attention.py
@@ -9,15 +9,15 @@ from contextlib import contextmanager
 from typing import Generator, Optional, Protocol, Tuple, final
 
 import torch
-import torch.nn.functional as F
 from torch import Tensor
 from torch.nn import Module
-from torch.nn.functional import dropout, softmax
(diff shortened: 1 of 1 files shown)


shagunsodhani:
Would love to know how these numbers look!

cbalioglu:
Hey @shagunsodhani, I felt uneasy with this PR and why `torch.compile()` was able to speed up SDPA so much even though it was a highly optimized/fused implementation. Spent quite a bit of time last night and it turned out to be that providing a boolean mask instead of a float mask to `scaled_dot_product_attention` was significantly degrading its runtime performance. I have the "real fix" in #439 and removed this `torch.compile` statement in that PR.

shagunsodhani:
Awesome! thank you @cbalioglu !

cbalioglu:
You are welcome!

PR #446, opened 2024-04-09
Remove WerMetric from __init__

cbalioglu:
This PR removes `WerMetric` from `__init__.py` of `fairseq2.metrics` since it transitively causes a dependency on the editdistance package. It also moves wav2vec 2.0 ASR metric bags to their own module for similar reason.


elbayadm (approved):
Awesome, thanks!

cbalioglu:
You are welcome!

PR #448, opened 2024-04-11
Introduce gradient clip APIs

cbalioglu:
This PR introduces gradient clipping helper APIs and updates some of the docstrs.


src/fairseq2/nn/utils/gradient.py
```diff
@@ -0,0 +1,124 @@
...
+from fairseq2.utils.logging import get_log_writer
+
+log = get_log_writer(__name__)
+
+
+def scale_gradients(module: Module, value: float) -> None:
+    """Scale gradients of ``module`` by ``value``.
+
+    :param module:
+        The module whose gradients to clip.
```
kauterry, line 27: clip?

cbalioglu: good catch! fixed, will merge.

PR #465, opened 2024-04-22
Add ffn_inner_dim_scale to LLama's factory

elbayadm:
Adding `ffn_inner_dim_scale` to LLama's config to allow for creating/loading LLama models with
 `ffn_inner_dim_scale ! = 2/3`

### DIFF
--- a/src/fairseq2/models/llama/factory.py
+++ b/src/fairseq2/models/llama/factory.py
@@ -72,6 +72,9 @@ class LLaMAConfig:
     dropout_p: float = 0.1
     """The dropout probability on outputs of Transformer layers."""
 
+    ffn_inner_dim_scale: float = 2 / 3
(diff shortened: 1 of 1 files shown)


src/fairseq2/models/llama/factory.py
```diff
@@ -72,6 +72,9 @@ class LLaMAConfig:
     dropout_p: float = 0.1
     """The dropout probability on outputs of Transformer layers."""
 
+    ffn_inner_dim_scale: float = 2 / 3
```
cbalioglu, line 75: Very nit: Can you please move this on top of `ffn_inner_dim_to_multiple` to group FFN related fields together?

cbalioglu (approved):

src/fairseq2/models/llama/factory.py
elbayadm: done

PR #470, opened 2024-04-23
Introduce layerwise checkpointing API

cbalioglu:
This PR introduces a convenience function for activation checkpointing that also exposes some additional parameters that are not straightforward to set with the official PyTorch API.

### DIFF
--- a/src/fairseq2/nn/checkpointing.py
+++ b/src/fairseq2/nn/checkpointing.py
@@ -1,0 +1,47 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from functools import partial
+
+from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
+    apply_activation_checkpointing,
+    checkpoint_wrapper,
+)
+from torch.nn import Module
+from torch.utils.checkpoint import checkpoint
+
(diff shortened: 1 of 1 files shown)


src/fairseq2/nn/checkpointing.py
```diff
@@ -0,0 +1,47 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from functools import partial
+
+from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
```
shagunsodhani, line 9: This is depending on an internal API. Do you know if PT plans to make the API public at some point ?

cbalioglu: PyTorch has some confusing API conventions. In this case the underscore prefix indicates that the API is more or less finalized, but the implementation is not stable yet. People are free to use such APIs on their own risk. In fact, just yesterday I realized that `torch.utils.checkpoint.checkpoint` heavily leaks memory when its `debug` parameter is set to `True`.

This particular `torch.distributed.algorithms._checkpoint` API has been out for several years now and used by pretty much all well-known training frameworks. I don't know why it is not marked as stable yet, most likely no one owns it at the moment and people already know and use it as it is.

cbalioglu: You are welcome!

PR #472, opened 2024-04-24
add truncation in text translator

avidale:
This PR allows bypassing the problem of `TextTranslator` failure when the number of source tokens exceeds the maximal length supported by the translation model. As tokenization happens within the `TextTranslator`, truncation, if implemented, should also happen there. The PR therefore adds a `max_src_len` argument to it, and enables truncation if this argument is not empty. 

None

**Check list:**

### DIFF
--- a/src/fairseq2/generation/text.py
+++ b/src/fairseq2/generation/text.py
@@ -153,6 +153,7 @@ class TextTranslator:
     _converter: SequenceToTextConverter
     _pad_idx: int
     _source_text_encoder: TextTokenEncoder
+    max_src_len: Optional[int]
 
     def __init__(
         self,
@@ -160,6 +161,7 @@ class TextTranslator:
         tokenizer: TextTokenizer,
         source_lang: Optional[str] = None,
         target_lang: Optional[str] = None,
+        max_src_len: Optional[int] = None,
     ) -> None:
         """
         :param generator:
@@ -170,6 +172,8 @@ class TextTranslator:
             The source language.
         :param target_lang:
             The target language.
+        :param max_src_len:
+            The number of tokens above which the source sequence gets truncated (None for no truncation)
         """
         self._converter = SequenceToTextConverter(
             generator, tokenizer, "translation", target_lang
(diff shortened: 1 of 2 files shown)


src/fairseq2/generation/text.py
```diff
@@ -153,13 +153,15 @@ class TextTranslator:
     _converter: SequenceToTextConverter
     _pad_idx: int
     _source_text_encoder: TextTokenEncoder
+    max_src_len: Optional[int]
```
cbalioglu, line 156: Nit: Could you please name this `_max_source_len` to be consistent with the rest of the naming convention?

src/fairseq2/generation/text.py
```diff
@@ -153,13 +153,15 @@ class TextTranslator:
...
 
     def __init__(
         self,
         generator: Seq2SeqGenerator,
         tokenizer: TextTokenizer,
         source_lang: Optional[str] = None,
         target_lang: Optional[str] = None,
+        max_src_len: Optional[int] = None,
```
cbalioglu, line 164: Same as about, ideally `max_source_len`.

src/fairseq2/generation/text.py
```diff
@@ -200,6 +205,9 @@ def __call__(self, source_text: str) -> Tuple[str, Seq2SeqGeneratorOutput]:
         """
         source_seq = self._source_text_encoder(source_text)
 
+        if self.max_src_len and source_seq.shape[0] > self.max_src_len:
```
cbalioglu, line 208: Although it would be a very edge case; if `max_src_len` is 0, then the first condition `if self.max_src_len` will be false, but maybe the user deliberately wants to trim to zero length (although does not make much sense, still expected since our parameter contract says that this parameter is ignored only if it is `None`). The "right" check would be `if self.max_src_len is not None and ...`

cbalioglu (commented):
Just left three nit comments.

src/fairseq2/generation/text.py
avidale: If the source is trimmed to 0 elements, the translation is going to fail, because the encoder would consume an empty tensor. Therefore, I don't want to apply truncation to zero, and instead will change the parameter contract.

cbalioglu (approved):

PR #477, opened 2024-04-25
Support APEX in RMSNorm

cbalioglu:
This PR adds support for APEX in `RMSNorm`.


src/fairseq2/nn/normalization.py
```diff
@@ -97,7 +107,7 @@ def forward(self, x: Tensor) -> Tensor:
     def extra_repr(self) -> str:
         return (
             f"normalized_shape={self.normalized_shape}, "
-            f"eps={self.eps}, "
+            f"eps={self.eps:G}, "
```
shagunsodhani, line 110: nice! I didnt know about this argument!

PR #483, opened 2024-04-26
Add support for tensor parallelism to CheckpointManager

cbalioglu:
This PR adds support for tensor parallelism to `FileCheckpointManager` to save/load sharded model checkpoints.

### DIFF
--- a/src/fairseq2/checkpoint.py
+++ b/src/fairseq2/checkpoint.py
@@ -125,7 +125,8 @@ class CheckpointManager(ABC):
         :param out:
             The model to load.
         :param device:
-            The device on which to load ``out`` if it is on the meta device.
+            The device on which to load ``out`` if it is on the meta device;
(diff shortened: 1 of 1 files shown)


src/fairseq2/checkpoint.py
```diff
@@ -204,13 +215,28 @@ def __init__(
...
         """
-        self._gang = gang
+        self._root_gang = gang
+
+        self._dp_gang = gang
+
+        self._shard_suffix = ""
+
+        if dp_gang is not None and tp_gang is not None:
+            self._dp_rank = dp_gang
```
shagunsodhani, line 225: maybe this is a typo ? We are assigning gang to rank

cbalioglu: Thanks a lot for catching it! You are right, a stupid bug from my end. Fixed it in #493.

PR #484, opened 2024-04-27
Revise LayerDrop implementation

cbalioglu:
This PR revises the LayerDrop implementation and moves its from `ModuleList` to `StandardTransformerEncoder` and `StandardTransformerDecoder`. Although the original implementation was ideal, both DDP and FSDP had trouble correctly handling forward/backward passes and either silently failed to sync gradients (DDP), or failed with some cryptic error. This implementation causes redundant computation of dropped layers, but since the autograd graph stays constant, both DDP and FSDP can handle it correctly.


src/fairseq2/nn/transformer/encoder.py
```diff
@@ -215,4 +245,24 @@ def extra_repr(self) -> str:
...
+
+
+class _RecordDropForBackwardFunction(Function):
+    @staticmethod
+    def forward(ctx: Any, x: Tensor, dropped_output: Tensor) -> Tensor:
+        return x
+
+    @staticmethod
+    def backward(ctx: Any, grad_output: Tensor) -> Tuple[Tensor, Tensor]:
+        return grad_output, torch.zeros_like(grad_output)
```
kauterry, line 268: Smartly done! The gradient with respect to x is going to be just the same as the gradient of the output, since this is just the identity function when you drop layers.

src/fairseq2/nn/module_list.py
```diff
@@ -4,85 +4,24 @@
...
 from torch import Generator
 from torch.nn import Module
 from torch.nn import ModuleList as TorchModuleList
 
-from fairseq2.typing import CPU
-
 
+# compat
 @final
 class ModuleList(TorchModuleList):
```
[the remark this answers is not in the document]
cbalioglu: We have several teams using this module for now. Everything tagged with “# compat” will eventually get removed once we migrate those uses (before v0.3 release)

src/fairseq2/nn/module_list.py
```diff
@@ -4,85 +4,24 @@
...
-
-    drop_p: float
-    generator: Optional[Generator]
-
     def __init__(
         self,
         modules: Optional[Iterable[Module]] = None,
         *,
         drop_p: float = 0.0,
         generator: Optional[Generator] = None,
```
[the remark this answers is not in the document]
cbalioglu: You mean the generator? It should be still around in case someone wants to provide a different RNG for layerdrop

src/fairseq2/nn/transformer/encoder.py
cbalioglu: Thanks!

kauterry: Also, what is the advantage of using this method over PyTorch hooks?

src/fairseq2/nn/transformer/encoder.py
```diff
@@ -204,6 +221,19 @@ def forward(
 
         return seqs, padding_mask
 
+    def _drop_iter(self) -> Iterator[Tuple[Module, bool]]:
```
shagunsodhani, line 224: I think this logic (and some other pieces in the forward pass) are used in decoder as well. Maybe we could consider adding a base component in the future (that both encoder and decoder would inherit from)

shagunsodhani:
> both DDP and FSDP had trouble correctly handling forward/backward passes and either silently failed to sync gradients (DDP), or failed with some cryptic error.

> both DDP and FSDP had trouble correctly handling forward/backward passes and either silently failed to sync gradients (DDP), or failed with some cryptic error.

Did this happen because rng across the different gpus was different - causing different layers to be dropped across gpus ?

PR #497, opened 2024-05-03
Workaround PyTorch 2.3 BC breaking changes

cbalioglu:
This PR works around the backwards-incompatible changes introduced in PyTorch 2.3


src/fairseq2/nn/utils/gradient.py
```diff
@@ -103,7 +103,7 @@ def clip_gradient_norm(
 
         return module.clip_grad_norm_(max_norm, norm_type)
 
-    return clip_grad_norm_(
+    return clip_grad_norm_(  # type: ignore[no-any-return]
```
shagunsodhani, line 106: I didnt follow what is the change here ?

PR #498, opened 2024-05-04
Adding support for wav2vec2 large models.

kauterry:
Adds support for wav2vec2 large models trained on Librispeech 960h (ls960) and LibriVox 60k (lv60k).

Tested with this simple code snippet:
lv60k
```
from fairseq2.models.wav2vec2.setup import load_wav2vec2_model
import torch

model = load_wav2vec2_model("wav2vec2_large_lv60k", device=torch.device("cuda:0"), dtype=torch.float32)
print(model)
```
Output: P1230849615

ls960:
```
from fairseq2.models.wav2vec2.setup import load_wav2vec2_model
import torch

model = load_wav2vec2_model("wav2vec2_large_ls960", device=torch.device("cuda:0"), dtype=torch.float32)
print(model)
```
Output: P1230849916

Fixes #{issue number}

**Check list:**


src/fairseq2/models/wav2vec2/factory.py
```diff
@@ -222,6 +222,32 @@ def _base() -> Wav2Vec2Config:
     return Wav2Vec2Config()
 
 
+@wav2vec2_arch("large_ls960")
+def _large_ls960() -> Wav2Vec2Config:
```
cbalioglu, line 226: Nit: Since the meaning of "ls" and "lv" prefixes are not immediately clear, could you please add a short doctstr to these two config functions, referring to which setting they refer in the original paper?

src/fairseq2/models/wav2vec2/factory.py
```diff
@@ -222,6 +222,32 @@ def _base() -> Wav2Vec2Config:
...
+            layer_drop_p=0.2,
+        ),
+        quantized_dim=768,
+        final_dim=768,
+    )
+
+
+@wav2vec2_arch("large_lv60k")
+def _large_lv60k() -> Wav2Vec2Config:
+    large_config = _large_ls960()
```
cbalioglu, line 243: Very nity: you can call this variable just `config` for "less" noise.

cbalioglu (commented):
Thanks a lot for the new configs! Just left two nit comments.

src/fairseq2/assets/cards/models/wav2vec2.yaml
```diff
@@ -20,6 +20,22 @@ checkpoint: "https://ai.meta.com/pseudo_dinosr_base/;gated=true"
 
 ---
 
+name: wav2vec2_large_ls960
```
cbalioglu, line 23: I don't think "ls960" suffix/prefix is overall necessary. I consider it the default one (since the base is also in fact a pretraining of ls960). So this can be just "wav2vec2_large".

Nit: please move it on top of "dinosr_base" which is a temporary asset that we will eventually remove.

src/fairseq2/models/wav2vec2/asr/factory.py
```diff
@@ -39,6 +39,30 @@ def _base_100h_encoder() -> Wav2Vec2EncoderConfig:
     return config
 
 
+def _ls960_encoder() -> Wav2Vec2EncoderConfig:
```
cbalioglu, line 42: I would call this `_large_encoder` to be consistent with `base`.

src/fairseq2/models/wav2vec2/asr/factory.py
```diff
@@ -39,6 +39,30 @@ def _base_100h_encoder() -> Wav2Vec2EncoderConfig:
...
+    config.feature_gradient_scale = 1.0
+    config.dropout_p = 0.0
+    config.attn_dropout_p = 0.0
+    config.ffn_inner_dropout_p = 0.1
+    config.layer_drop_p = 0.1
+
+    return config
+
+
+def _lv60k_encoder() -> Wav2Vec2EncoderConfig:
```
cbalioglu, line 54: And this `_large_lv60k_encoder`, again to be consistent with the pretrained `large_lv60k` name.

src/fairseq2/models/wav2vec2/asr/factory.py
```diff
@@ -96,10 +120,49 @@ def _base_100h() -> Wav2Vec2AsrConfig:
     config = _base_10h()
 
     config.encoder_config = _base_100h_encoder()
-
     return config
 
 
+@wav2vec2_asr_arch("large_ls960_10h")
+def _large_ls960_10h() -> Wav2Vec2AsrConfig:
```
cbalioglu, line 127: As I mentioned, "ls960" is not really needed for this and the next 100h.

src/fairseq2/models/wav2vec2/asr/factory.py
```diff
@@ -96,10 +120,49 @@ def _base_100h() -> Wav2Vec2AsrConfig:
...
+    """wav2vec2 large arch trained on the Librispeech 960h dataset."""
+    return Wav2Vec2AsrConfig(
```
cbalioglu, line 129: For better readability, I prefer the convention:

```
config = _base_10h()

config.encoder_config = ...

return config
```

instead of `config = Wav2Vec2AsrConfig(...)`.

cbalioglu (commented):
Just left a few nit comments about naming. Otherwise, looks good. Once you address them, we can merge it.

src/fairseq2/assets/cards/models/wav2vec2.yaml
kauterry: I will move it on top of dinosr_base, but I don't think this should be the default one. Simply bcoz the other model is a much better models, and we should have the default model to be the better model. Also, you can see the large model configs to be named as "vox_*.yaml" here, referring to the model trained on LibriVox 60k. I'm happy to call the other one "large" and this one "large_ls960".

cbalioglu: I don't think the default has to be necessarily to "best" model. It should be just consistent with the literature/paper. According to the last paragraph of section 4.2 (pretraining) in wav2vec2 paper, the base model was solely trained with LS-960. So if we call it just "base", then "large" should imply that it is the same training recipe with just a larger architecture. So we have three variants: base with ls960h, large with ls960h, and large with lv60k. In my opinion "base", "large", and "large_lv60k" sound more consistent.

cbalioglu (approved):
LGTM! Thanks for the contribution! :)

PR #501, opened 2024-05-08
Allow overwriting when recording formatters

elbayadm:
Allow for overwriting metric formatters when calling `register_metric_formatter`

### DIFF
--- a/src/fairseq2/metrics/recorder.py
+++ b/src/fairseq2/metrics/recorder.py
@@ -72,7 +72,7 @@ def register_metric_formatter(
     display_name: str,
     priority: int,
     format_fn: Callable[[Any], str],
-    overwrite: Optional[bool] = False,
+    overwrite: bool = False,
 ) -> None:
     """Register a string formatter for the specified metric.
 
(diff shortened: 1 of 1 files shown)


src/fairseq2/metrics/recorder.py
```diff
@@ -68,7 +68,11 @@ def format_as_float(value: Any, *, postfix: Optional[str] = None) -> str:
 
 
 def register_metric_formatter(
-    name: str, display_name: str, priority: int, format_fn: Callable[[Any], str]
+    name: str,
+    display_name: str,
+    priority: int,
+    format_fn: Callable[[Any], str],
+    overwrite: Optional[bool] = False,
```
cbalioglu, line 75: Nit: No need for `Optional`. This can be just `overwrite: bool = False`.

cbalioglu: Can you also write a short doctstr for `overwrite`? Something like `If ``True``, overwrites any existing metric with the same name.`.

cbalioglu (commented):
Thanks for the PR! Just left two nit comments. Once you address them, it is good to go.

cbalioglu (approved):
LGTM!

PR #502, opened 2024-05-08
Support dinosr inference (pseudo w2v2 approach)

iamyuanchung:

The PR provides an adhoc way to support inference of the DinoSR tokenizer pre-trained in fairseq.  The idea is to reuse the existing wav2vec2 architecture in fairseq2 but load it with dinosr weights.  This approach is feasible because besides the pretraining-specific modules in wav2vec2 (inclduing mask_emb, quantizer, final_proj, and project_q) and dinosr (mask_emb, codebook_* and ema), which will not be used in fine-tuning or inference, both models are essentially identical as a stack of Transformer layers.

### DIFF
--- a/src/fairseq2/assets/cards/wav2vec2.yaml
+++ b/src/fairseq2/assets/cards/wav2vec2.yaml
@@ -16,7 +16,7 @@ name: dinosr_base
 model_type: wav2vec2  # compat
 model_family: wav2vec2
 model_arch: pseudo_dinosr_base
-checkpoint: "https://dl.fbaipublicfiles.com/fairseq/wav2vec/wav2vec2_fs2_small.pt"
+checkpoint: "/fsx-ust/andyyuan/pt_models/xlsr_ckpts/pseudo_dinosr_spin_ft_use7_frz5_cl4096.pt"
 
 ---
 
--- a/src/fairseq2/assets/cards/wav2vec2.yaml
+++ b/src/fairseq2/assets/cards/wav2vec2.yaml
@@ -16,7 +16,7 @@ name: dinosr_base
 model_type: wav2vec2  # compat
 model_family: wav2vec2
 model_arch: pseudo_dinosr_base
-checkpoint: "/fsx-ust/andyyuan/pt_models/xlsr_ckpts/pseudo_dinosr_spin_ft_use7_frz5_cl4096.pt"
+checkpoint: "https://ai.meta.com/pseudo_dinosr_base/;gated=true"
 
 ---
 
--- a/src/fairseq2/assets/cards/wav2vec2.yaml
+++ b/src/fairseq2/assets/cards/wav2vec2.yaml
@@ -12,6 +12,14 @@ checkpoint: "https://dl.fbaipublicfiles.com/fairseq/wav2vec/wav2vec2_fs2_small.p
 
 ---
 
+name: dinosr_base
+model_type: wav2vec2  # compat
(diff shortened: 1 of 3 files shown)


src/fairseq2/assets/cards/wav2vec2.yaml
```diff
@@ -12,6 +12,14 @@ checkpoint: "https://dl.fbaipublicfiles.com/fairseq/wav2vec/wav2vec2_fs2_small.p
 
 ---
 
+name: dinosr_base
+model_type: wav2vec2  # compat
+model_family: wav2vec2
+model_arch: pseudo_dinosr_base
+checkpoint: "/fsx-ust/andyyuan/pt_models/xlsr_ckpts/pseudo_dinosr_spin_ft_use7_frz5_cl4096.pt"
```
iamyuanchung, line 19: I wonder if it's actually better to have this config, which is likely to only be tentative, only exists in seamless_next rather than the public fairseq2 repo.

src/fairseq2/models/wav2vec2/setup.py
```diff
@@ -57,7 +57,7 @@ def convert_wav2vec2_checkpoint(
         r"^encoder\.layers\.([0-9]+)\.fc2\.":                 r"encoder.layers.\1.ffn.output_proj.",
         r"^encoder\.layers\.([0-9]+)\.final_layer_norm\.":    r"encoder.layers.\1.ffn_layer_norm.",
         r"^encoder\.embed_tokens\.":                          r"encoder_frontend.embed.",
-        r"^encoder\.pos_conv\.0\.":                           r"encoder_frontend.pos_encoder.conv.",
+        r"^encoder\.pos_conv\.([0-4])\.":                     r"encoder_frontend.pos_encoder.conv.",
```
mavlyutovr, line 60: is it possible that there will be multiple pos_conv layers in the same checkpoint ? (pos_conv.0, .. pos_conv.3)

cbalioglu (commented):
I think seamless_next would be better to host this model declaration. I have been very busy with the launcher work (finally finished). Let me revise the AWS installer script tomorrow and merge the initial seamless_next PR. As a quick follow-up we can move this to seamless_next then.

src/fairseq2/models/wav2vec2/setup.py
iamyuanchung: Yes, the number of pos_conv layers is determined by the passed in `pos_conv_depth` value, which is 1 by default: https://github.com/fairinternal/fairseq-py/blob/seamless_main/fairseq/models/wav2vec/wav2vec2.py#L258

In dinosr we use 5.

iamyuanchung:
> I think seamless_next would be better to host this model declaration. I have been very busy with the launcher work (finally finished). Let me revise the AWS installer script tomorrow and merge the initial seamless_next PR. As a quick follow-up we can move this to seamless_next then.

Sounds good, I kind of agree seamless_next would be a better place to host this model declaration.  Let me know when the initial seamless_next PR is merged.  Thanks!

mavlyutovr (approved):

src/fairseq2/models/wav2vec2/setup.py
cndn: I updated this line to 
> r"^encoder\.pos_conv\.([0-4])\.0.":                    r"encoder_frontend.pos_encoder.layers.\1.conv.",

to make ckpt loaded for me.

PR #503, opened 2024-05-09
Fix nn/transformer/encoder.py

lifeiteng:
Fix error
```
... site-packages/fairseq2/nn/transformer/encoder.py", line 191, in forward
    if self._layer_output_hooks and self.layers.drop_p > 0.0 and self.training:

... site-packages/torch/nn/modules/module.py", line 1695, in __getattr__
    raise AttributeError(f"'{type(self).__name__}' object has no attribute '{name}'")
AttributeError: 'ModuleList' object has no attribute 'drop_p'
```


cbalioglu (approved):
Thanks for the fix!

PR #510, opened 2024-05-15
Fix fake gang bug in CheckpointManager

cbalioglu:
This PR fixes the `FakeGang` handling bug in `CheckpointManager.load_checkpoint()`.


elbayadm (approved):
LGTM, thanks for the fix!

cbalioglu:
Thanks for reporting it!

PR #516, opened 2024-05-16
Reorganize wav2vec2 model configs

cbalioglu:
This PR reorganizes the model configuration setup of wav2vec2 model variants.


src/fairseq2/models/wav2vec2/asr/factory.py
```diff
@@ -12,29 +12,42 @@
...
     Wav2Vec2EncoderConfig,
+    wav2vec2_encoder_archs,
 )
 from fairseq2.models.wav2vec2.masker import Wav2Vec2Masker
 from fairseq2.typing import DataType, Device
 
 WAV2VEC2_ASR_FAMILY: Final = "wav2vec2_asr"
 
 
+def _base_10h_encoder() -> Wav2Vec2EncoderConfig:
```
kauterry, line 23: It's misleading to just call this as just "encoder", could we call them "encoder_for_asr"?

PR #519, opened 2024-05-17
Remove faircluster references

cbalioglu:
This PR removes references to faircluster in OSS Code.


avidale:
So what is the recommended way to work with the `faircluster` env resolver now?
We have plenty of `@faircluster` model cards in SONAR and MERES.

PR #542, opened 2024-05-24
Introduce more logging

cbalioglu:
This PR introduces more logging to LM finetuning and wav2vec2 recipes.

### DIFF
--- a/src/fairseq2/checkpoint.py
+++ b/src/fairseq2/checkpoint.py
@@ -32,13 +32,34 @@ from torch.distributed.fsdp.api import FullStateDictConfig, StateDictType
 from torch.nn import Module
 
 from fairseq2.gang import Gang
+from fairseq2.logging import get_log_writer
 from fairseq2.models.utils.checkpoint import load_checkpoint
 from fairseq2.typing import CPU, DataClass, override
 from fairseq2.utils.dataclass import to_safe_dict
 
+log = get_log_writer(__name__)
+
 
 class CheckpointManager(ABC):
     """Saves and loads training checkpoints."""
+
+    @abstractmethod
+    def set_model_metadata(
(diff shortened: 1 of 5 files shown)


src/fairseq2/checkpoint.py
```diff
@@ -32,14 +32,35 @@
 from torch.nn import Module
 
 from fairseq2.gang import Gang
+from fairseq2.logging import get_log_writer
 from fairseq2.models.utils.checkpoint import load_checkpoint
 from fairseq2.typing import CPU, DataClass, override
 from fairseq2.utils.dataclass import to_safe_dict
 
+log = get_log_writer(__name__)
```
shagunsodhani, line 40: nit: maybe this should be called `logger` or `log_writer`. `log` is a confusing name for the writer

src/fairseq2/checkpoint.py
```diff
@@ -448,6 +461,8 @@ def raise_error(cause: Exception) -> NoReturn:
 
     @override
     def save_consolidated_fsdp_model(self, step_nr: int, model: Module) -> None:
+        log.info("Extracting consolidated FSDP state dictionary of the model.")
```
shagunsodhani, line 464: nit: should this be a debug log instead of info log ?

cbalioglu: Hey @shagunsodhani, sorry just catching up with a pile of GitHub PRs and issues. This could be debug as well. The reason I kept it as info is because for large models (e.g. LLaMA 70B) this extraction step can take a while and it is helpful to know at what stage the checkpointing is while doing a regular training.

shagunsodhani: Makes sense. Thank you for the clarification.

PR #581, opened 2024-06-07
Introduce broadcast_module

cbalioglu:
This PR introduces the `broadcast_module()` helper function and updates the wav2vec2 ASR evaluation recipe to use it. This significantly reduces the pressure on disk I/O for large models and instead broadcasts the module state over the network fabric. DDP and FSDP already offer a similar feature, but this is standalone and can be used with evaluation and inference jobs as well. 

As of today, we rely on the private `torch.distributed._broadcast_coalesced` function although its c10d counterpart is a public API. Once the P0 items are delivered today, I will expose `c10d::broadcast_coelesced` in fairseq2n and remove the private API use.

### DIFF
--- a/src/fairseq2/nn/utils/module.py
+++ b/src/fairseq2/nn/utils/module.py
@@ -30,6 +30,7 @@ from torch import Tensor
 from torch.nn import Module, Parameter
 from torch.nn.utils import remove_weight_norm  # type: ignore[attr-defined]
 
+from fairseq2.gang import Gang
 from fairseq2.logging import LogWriter
 from fairseq2.typing import CPU, META, Device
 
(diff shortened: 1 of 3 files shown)


src/fairseq2/nn/utils/module.py
```diff
@@ -389,6 +390,59 @@ def infer_device(
...
+
+    tensors = []
+
+    for param in module.parameters():
+        if param in memo:
+            continue
+
+        memo.add(param)
+
+        tensors.append(param.detach())
```
shagunsodhani, line 422: should this also support broadcasting the gradients for the tensors ?

cbalioglu: Broadcasting grads is trickier since if the `module` on the receiving side does not have grads (which would be the common case), than you also have to find a way to copy over the autograd graph and create grad tensors on the fly instead of doing just an all-reduce call. For now, we can log a warning if we see that the source module has grads, telling to the user that they won't be synced. If we really find a legitimate use case for syncing grads in the future, we can reconsider extending this function.

cbalioglu: #594 594

shagunsodhani: Sounds good :)

PR #591, opened 2024-06-13
Introduce temporary_manual_seed()

cbalioglu:
This PR introduces a new `temporary_manual_seed()` function which is a "cleaner" way to `torch.random.fork_rng() + torch.manual_seed()` as it internally uses `RngBag` to correctly set the seed only for the passed devices. It also extends `reset_parameters()` and `to_device()` utility functions with an extra convenience `seed` parameter.

### DIFF
--- a/src/fairseq2/assets/store.py
+++ b/src/fairseq2/assets/store.py
@@ -173,7 +173,7 @@ class StandardAssetStore(AssetStore):
             name = name[:-1]
 
         raise AssetNotFoundError(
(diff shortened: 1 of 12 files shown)


shagunsodhani:
This seems cool! Could you share an example of where this is commonly used ?

cbalioglu:
We use it in `reset_parameters()` and `to_device()` functions in `fairseq2.nn.utils.module`.

- https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/nn/utils/module.py#L83
- https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/nn/utils/module.py#L171

We also use it in unit/integration tests to avoid changing the global seed.

PR #598, opened 2024-06-17
allow loading experimental models

antoine-tran:

We want to use fairseq2 to develop models. Before declaring the final model with model card in certain asset store, we want to be able to load the experimental models, e.g. for evaluation in downstream tasks.

This PR assumes the model card is in the form of a .YAML file (such files can be generated automatically , for instance as part of the training loop). It will try to load the .YAML file if the given model card name is not found in its officially registered stores 

### DIFF
--- a/src/fairseq2/models/loader.py
+++ b/src/fairseq2/models/loader.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+from pathlib import Path
 from pickle import PickleError
 from typing import Any, Dict, Generic, Optional, Protocol, TypeVar, Union, final
 
@@ -15,6 +16,7 @@ from fairseq2.assets import (
     AssetCardError,
     AssetDownloadManager,
     AssetError,
+    AssetNotFoundError,
     AssetStore,
     default_asset_store,
     default_download_manager,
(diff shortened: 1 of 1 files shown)


src/fairseq2/models/loader.py
```diff
@@ -356,7 +358,18 @@ def __call__(
...
-            card = self._asset_store.retrieve_card(model_name_or_card)
+            try:
+                card = self._asset_store.retrieve_card(model_name_or_card)
+            except AssetNotFoundError as err:
+                model_path = Path(model_name_or_card)
+                # If the card is not found, try looking it up by interpreting model_name as a path to the yaml card.
+                if model_path.exists() and model_path.suffix == ".yaml":
+                    import yaml
+                    with open(model_path, "r", encoding="utf-8") as f:
+                        card_data = yaml.full_load(f)
```
cbalioglu, line 369: I would prefer this to be `yaml.safe_load()` instead since we take an arbitrary path. Is there a specific requirement for `full_load()`?

cbalioglu (commented):
Thanks for the PR! I just have a small feedback. I also see that black is complaining about formatting. Having said that once we merge this, I might try to refactor this logic to a common place and use it for datasets and tokenizer as well. Let's first unblock your use case though.

cbalioglu (approved):
LGTM!

PR #599, opened 2024-06-18
Add hf metric wrapper

antoine-tran:

This PR adds a wrapper to HuggingFace's `evaluate.Metric` to make it compatible to fairseq2.metrics APIs. This enables evaluating fairseq2 model on many downstream tasks available in HuggingFace, using standard evaluation metrics.

Fixes #{issue number}

**Check list:**

### DIFF
--- a/src/fairseq2/metrics/bag.py
+++ b/src/fairseq2/metrics/bag.py
@@ -34,6 +34,7 @@ class MetricBag:
         super().__setattr__("_original_metrics", None)
 
         self._gang = gang
+        self.auto_sync = False
 
     def __getattr__(self, name: str) -> Any:
         if "_metrics" in self.__dict__ and name in self._metrics:
@@ -62,6 +63,13 @@ class MetricBag:
             if name in self._persistent_metrics:
                 del self._persistent_metrics[name]
         else:
+            assert name not in [
+                "auto_sync",
+                "_gang",
+                "_metrics",
+                "_persistent_metrics",
+                "_original_metrics",
+            ], f"Cannot delete protected attribute: {name}"
             super().__delattr__(name)
 
     @final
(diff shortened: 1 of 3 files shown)


cbalioglu:
I think we might need to reconsider the generic metrics API in fairseq2 before merging this PR. Right now it pretty much uses torcheval which apparently uses a different approach to syncing metrics. I have a fundamental question though. In your use case, how do you combine fairseq2 metrics and HG evaluate? Looks like HG has some similar APIs like `combine` that acts like `MetricBag`. Which approach does make more sense in your opinion (1) using the two APIs in parallel in your code or (2) abstracting away HG under fairseq2.metrics (i.e. this PR)? If the latter, we might consider defining a `Metric` interface in fairseq2 (instead of using torcheval'a `Metrics` interface) and have derived interfaces for torcheval and HG that fit better to the corresponding libraries.

antoine-tran:
> I think we might need to reconsider the generic metrics API in fairseq2 before merging this PR. Right now it pretty much uses torcheval which apparently uses a different approach to syncing metrics.

I think the decision of binding fairseq2 to torcheval is a good one. It is a thin, low-level library to help developing further libraries on top. HG on the other hand targets ready-to-use use cases and it makes sense to hide the synchronization logic inside.

>  I have a fundamental question though. In your use case, how do you combine fairseq2 metrics and HG evaluate? Looks like HG has some similar APIs like `combine` that acts like `MetricBag`. Which approach does make more sense in your opinion (1) using the two APIs in parallel in your code or (2) abstracting away HG under fairseq2.metrics (i.e. this PR)? If the latter, we might consider defining a `Metric` interface in fairseq2 (instead of using torcheval'a `Metrics` interface) and have derived interfaces for torcheval and HG that fit better to the corresponding libraries.

Good questions. Until now I always followed (1) to have a separate metrics API in evaluation tasks. It served well my use cases (pure evaluation), and now I am thinking of using the evaluation within the training loop. Having the same MetricBag API helps me "register" the states and continue later.

antoine-tran:
@cbalioglu Maybe we can indeed wait a bit for the use case to be clearer and gather more requirements before merging this PR.

PR #601, opened 2024-06-20
Add wrapper for HuggingFace Dataset

antoine-tran:
This PR adds a light-weight wrapper to HuggingFace datasets to make it compatible to fairseq2.data APIs. We can now construct a HuggingFace `datasets.Dataset` however we want:

```python
from datasets import load_dataset, Audio
from transformers import AutoModelForAudioClassification, AutoFeatureExtractor

dataset = load_dataset("PolyAI/minds14", "en-US", split="train")
dataset = dataset.cast_column("audio", Audio(sampling_rate=16000))
dataset = dataset.map(writer_batch_size=10)
... # Other transformation
```

and then later on create a fairseq2 DataPipeline out of it, with a custom converter function to convert the HuggingFace data items (in dictionary format) to fairseq2 batching format

```python

def example_to_batch(example: Example) -> Seq2SeqBatch:
    return Seq2SeqBatch(.....)

from fairseq2.datasets.huggingface create_hf_reader
from fairseq2.gang import FakeGang

pipeline_reader = create_hf_reader(dataset, converter=example_to_batch, gang=FakeGang(), prefetch=3,....)

for data in pipeline_reader:
    # data is a Python dictionary
    # handle logic
```

** Batching:

We can use the datasets' batching API, or the fairseq2 batching API in `fairseq2.datasets.batching` . In the later case, simply define the `batching` parameters: 

```python

from fairseq2.datasets.batching import StaticBatching

pipeline_reader = create_hf_reader(dataset, converter=example_to_batch, batching=StaticBatching(batch_size=10),....)
```


src/fairseq2/recipes/hf/datasets.py
```diff
@@ -0,0 +1,188 @@
...
+        # Resolve the batching strategies
+        def resolve_batching_strategy() -> int:
+            if bucket_size is None and max_num_elements is None:
+                return 0  # NO_BATCH
+            elif bucket_size is not None and max_num_elements is not None:
+                raise ValueError("Can only batch by size or by length, not both")
+            elif bucket_size is not None:
+                return 1  # BATCH_BY_SIZE
+            else:
+                assert (
```
cbalioglu, line 98: Please use a `ValueError` here similar to line 94. We want to avoid "bad research library habits". `assert` statements should only be used to check some implementation invariants. That means; a failed assert should mean a bug in the implementation. Any error in user-specific inputs should be handled via exceptions.

src/fairseq2/recipes/hf/datasets.py
```diff
@@ -0,0 +1,188 @@
...
+                return 2  # BATCH_BY_LEN
+
+        batch_mode = resolve_batching_strategy()
+        if batch_mode == 0:
+            data = dataset.data.to_batches()  # list of RecordBatch
+        else:
+            data = dataset.to_list()
+        builder = read_sequence(data)
+
+        seed += 1
```
cbalioglu, line 110: We don't really have any RNG operation up to this point. This line is not necessary.

src/fairseq2/recipes/hf/datasets.py
```diff
@@ -0,0 +1,188 @@
...
+                skip_below_min_examples = min_seq_len is not None
+                min_seq_len = min_seq_len or 1
+
+                # Bucket by audio length.
+                assert max_num_elements and max_seq_len
+                bucket_sizes = create_bucket_sizes(
+                    max_num_elements=max_num_elements,
+                    max_seq_len=max_seq_len,
+                    min_seq_len=min_seq_len,
+                    num_seqs_multiple_of=8,
```
cbalioglu, line 137: Is this "multiple-of" really needed for this dataset?

src/fairseq2/recipes/hf/datasets.py
```diff
@@ -0,0 +1,188 @@
...
+@final
+class HFDatasetLoader(DatasetLoader[HFDataset]):
+    def __call__(
+        self,
+        dataset_name_or_card: Union[str, AssetCard],
+        *,
+        force: bool = False,
+        progress: bool = True,
+    ) -> HFDataset:
+        assert (
```
cbalioglu, line 171: Please use a `TypeError` instead of `assert`.

src/fairseq2/recipes/hf/datasets.py
```diff
@@ -0,0 +1,188 @@
...
+        Convert HF Dataset into a fairseq2 PipelineReader.
+        The HF-specific processing logic is assumed to be
+        done before the conversion.
+
+        **Batching**: The batching can be done externally or using
+        bucketing strategies in fairseq2. If both ``bucket_size``
+        and ``max_num_elements`` are None, then we assume the
+        ``dataset`` is already batched and do nothing. Otherwise:
+
+        - If ``bucket_size`` is not None, generate batches of fixed size
```
cbalioglu, line 51: Wondering whether it would be cleaner if we just had two separate functions. One using `bucket_by_length`, the other `bucket`. `bucket_size` and `max_num_elements` are mutually exclusive, so they feel a bit out of place when taken as parameters in a single function.

src/fairseq2/recipes/hf/datasets.py
```diff
@@ -0,0 +1,188 @@
...
+        self.name = name
+    
+    def __repr__(self) -> str:
+        return self.name
+
+    def create_reader(
+        self,
+        dataset: Dataset,
+        gang: Gang,
+        bucket_size: Optional[int] = None,
```
cbalioglu, line 31: Please make parameters below `gang`, keyword-only by using the `<args>, *, <kwargs>` syntax of Python.

src/fairseq2/recipes/hf/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
cbalioglu, line 1: No need for copyright in empty files.

cbalioglu (commented):
Ok, left a few comments. Also, I think it would make more sense to have this under `fairseq2.datasets` instead of `fairseq2.recipes.hf`.

antoine-tran:
> Ok, left a few comments. Also, I think it would make more sense to have this under `fairseq2.datasets` instead of `fairseq2.recipes.hf`.

I didn't want to claim one dependency of fairseq2 on transformers just because of one types of pipelines, that's why

src/fairseq2/recipes/hf/datasets.py
antoine-tran: ok, 8 is not a good default value.

I would still keep num_seqs_multiple_of, but now as an extra option. The point is we can also use this HuggingFace dataset wrapper to load an ASR dataset, so we should keep the same options used in the ASR use case.

src/fairseq2/recipes/hf/datasets.py
antoine-tran: ok, I create a `Batcher` interface that act as a "visitor" to be applied on the data pipeline. Hope this modularization is enough.

src/fairseq2/datasets/huggingface.py
```diff
@@ -0,0 +1,192 @@
...
+except ImportError:
+    has_datasets = False
+else:
+    has_datasets = True
+
+
+Example = Dict[str, Any]
+
+
+class Batcher(abc.ABC):
```
cbalioglu, line 34: I suggest using the batching types defined in [here](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/datasets/batching.py) along with how they are used in other datasets (e.g. see [here](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/datasets/instruction.py#L332-L352)).

src/fairseq2/datasets/huggingface.py
```diff
@@ -0,0 +1,192 @@
...
+            min_data_len=self.min_seq_len,
+            skip_below_min_examples=self.skip_below_min_examples,
+            skip_above_max_examples=True,
+        )
+
+
+def create_hf_reader(
+    dataset: Dataset,
+    gang: Gang,
+    converter: Callable[[Example], Seq2SeqBatch],
```
cbalioglu, line 104: I think instead of hard-coding to `Seq2SeqBatch`, making the batch type a generic type would be a lot more flexible (for an example see [here](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/datasets/data_reader.py#L23).

src/fairseq2/datasets/huggingface.py
```diff
@@ -0,0 +1,192 @@
...
+    *,
+    batcher: Optional[Batcher] = None,
```
cbalioglu, line 106: This can be `Optional[Union[StaticBatching, LengthBatching]] = None`. See other dataset implementations.

src/fairseq2/datasets/huggingface.py
```diff
@@ -0,0 +1,192 @@
...
+    num_accumulate: int = 1,
+    num_prefetch: int = 1,
+    **extra: Any,
+) -> DataPipelineReader[Seq2SeqBatch]:
+    """
+    Convert HF Dataset into a fairseq2 PipelineReader.
+    The HF-specific processing logic is assumed to be
+    done before the conversion.
+
+    **Batching**: The batching can be done externally or using
```
cbalioglu, line 116: I assume we can delete these two paragraphs now.

cbalioglu (commented):
I left two suggestions. Once is about using a generic batch type instead of hardcoding the return type to `Seq2SeqBatch` (which should be very trivial to change), the other is using the batching types we already have in fairseq2 instead of `Batching`. Otherwise the rest looks good to me. Please let me know if you have any questions.

antoine-tran:
@cbalioglu Thanks for the comments - they have been taken now

src/fairseq2/datasets/huggingface.py
```diff
@@ -0,0 +1,153 @@
...
+                "get datasets.DatasetDict. Make sure you specify `split` "
+                "when loading the dataset"
+            )
+        raise TypeError(f"Expect datasets.Dataset type, get {type(dataset)}")
+
+    if batching is None:
+        data = dataset.data.to_batches()  # list of RecordBatch
+    else:
+        data = dataset.to_list()
+    builder = read_sequence(data)
```
cbalioglu, line 99: Nit: we now have a new `read_iterator` function that might be more efficient than `read_sequence`.

cbalioglu (approved):
Thanks for addressing the feedback! I just left a nit comment. It is not a blocker, so I am stamping.

PR #602, opened 2024-06-20
Add `target_twoway` mode in the NLLB tokenizer

avidale:

This PR introduces a `target_twoway` mode in the NLLB tokenizer. It represents the target text as `</s>__lang_code__ The text.</s>`. 

This mode is handy for training translation models (like NLLB) or text decoding models (like SONAR decoders). With this mode, all-but-the last tokens (e.g. `</s>__lang_code__ The text.`) can be used as teacher-forced inputs to the decoder using its training, and all-but-the-first tokens (e.g. `__lang_code__ The text.</s>`) can be used as the next-token targets for the same decoder. Which is how NLLB (and also SONAR) models are supposed to be trained. 

Without this change, the similar result could be achieved in less straightforward ways:
- either by tokenizing the decoder inputs in the `target` mode and the decoder outputs in the `source` mode, and passing both the sequences to the training step, and truncating/collating/padding both of them identically - which seems slightly brittle;
- or by tokenizing the decoder inputs in the `target` mode and then turning them into decoder targets by removing the first `</s>` token and adding the one after the text - which is not an easy operation on a padded batch of different-length sequences, and may add an `</s>` token incorrectly to a sequence that has been truncated;
- or by tokenizing the decoder targets in the `source` mode and then turning them into decoder inputs by removing the last token (either `</s>` or pad token) and adding an `</s>` token in the beginning - which is actually fine (e.g. the `transformers` implementation of NLLB does it with their [shift_tokens_right](https://github.com/huggingface/transformers/blob/main/src/transformers/models/m2m_100/modeling_m2m_100.py#L60) function), but it requires either concatenating tensors or copying and filling them in place, which is slightly less efficient then just selecting all-but-the-first or all-but-the-last tokens.

I believe that by having a single token sequence and selecting the right tokens from it to be decoder inputs or decoder targets is the simplest strategy, so this new tokenization mode is introduced to enable it.

None

**Check list:**

### DIFF
--- a/src/fairseq2/models/nllb/tokenizer.py
+++ b/src/fairseq2/models/nllb/tokenizer.py
@@ -100,6 +100,13 @@ class NllbTokenizer(SentencePieceTokenizer):
             # the language token.
             prefix_tokens = ["</s>", f"__{lang}__"]
             suffix_tokens = []
+        elif mode == "target_twoway":
+            # Encode a sentence with the EOS both in the beginning and in the end.
+            # Then its prefix without the last token (e.g. "</s> __eng_Latn__ Hello world !") could serve as the decoder input,
+            # and its suffix without the first token (e.g. "__eng_Latn__ Hello world ! </s>") could serve as the decoder target.
+            # This is an elegant way to make sure that decoder inputs and targets are always consistent.
+            prefix_tokens = ["</s>", f"__{lang}__"]
+            suffix_tokens = ["</s>"]
         else:
             raise ValueError(
                 f"`mode` must be 'source' or 'target', but is '{mode}' instead."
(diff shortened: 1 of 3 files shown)


src/fairseq2/models/nllb/tokenizer.py
```diff
@@ -100,6 +100,13 @@ def create_encoder(
             # the language token.
             prefix_tokens = ["</s>", f"__{lang}__"]
             suffix_tokens = []
+        elif mode == "target_twoway":
```
cbalioglu, line 103: Do we need the "target" `mode` at all after this PR? My understanding is that the "target" mode as it is (without an appended <eos>) is pretty much a bug and should have been what "target_twoway" is from the beginning.

avidale: Well, there are at least two reasons why one might want to keep the `target` mode:
1. Just for backward compatibility. If anyone already uses the `target` mode, let them keep using it!
2. I could imagine a case where one needs only the inputs to the decoder, but not its targets; for example, to sample from the teacher-forced next token distribution. In this case, the terminal `</s>` is not needed indeed, so the `target` tokenization mode is fine.

cbalioglu (approved):

PR #613, opened 2024-06-27
Revise the implementation of Trainer

cbalioglu:
This is a relatively big PR that revises the implementation of `Trainer`. I believe it is almost in its final state at this point with only a couple nit features remaining. Along with `Trainer`, this PR also updates the instruction finetuning and wav2vec2 ASR finetuning recipes.


cbalioglu:
Hey @orena1, yes, actively working on it! Most of the work is done, will start parity checking with fairseq and original paper very soon. Hope the recipe will land in the next 2-3 week at latest.

orena1:
Thanks!!
Just wondering any plans for data2vec2 ? I've found it to be faster than wav2vec with similar performance.

cbalioglu:
It is on the roadmap. Except EMA modules it is pretty similar to wav2vec2, so I expect it to be less work once we finalize the wav2vec2 recipe.

PR #616, opened 2024-06-29
Add dynamic_bucket op

### LINKED ISSUE #453: Dynamic bucketing
**Is your feature request related to a problem? Please describe:**
Samples participating in data pipeline can possess different characteristics (like length, number of tokens, ...).
Currently we can only bucket a fixed number of elements with `.bucket(...)` which may not be well adapted to more complex situations. Thus, the idea is to offer a new method (to challenge) like this :
```python
.dynamic_bucket(threshold: float =1000, 
                cost_fn=lambda sample: len(sample["tokens"]),
                nb_min:Optional[int] = 2,
                nb_max:Optional[int] = 20)
```
which would have the following behavior:
* `cost_fn` is callable that for each element of pipeline returns a positive float/int
* we bucket several consecutive elements until their total cost becomes bigger than `threshold`
* we also optionally control the min/max number of element to put into a bucket.

**Describe the solution you would like:**
There should be native implementation in cpp of this dynamic buckets creation.

**Describe the alternatives you have considered:**
I don't know any simple workarounds (except maybe creating very large buckets first and next doing smaller dynamic buckets with `yield_from`).

**Additional Context:**
I would like to keep `cost_fn` function open so it could fit various situations.

syleshfb:
Adds a dynamic bucket operator according to the spec in #453 

Fixes #453 

No

**Check list:**


tests/unit/data/data_pipeline/test_dynamic_bucket.py
```diff
@@ -0,0 +1,230 @@
...
+        with pytest.raises(
+            ValueError, match=r"^`threshold` must be greater than zero\.$"
+        ):
+            read_sequence(list(range(100))).dynamic_bucket(-1, lambda x: 1).and_return()
+
+    def test_op_saves_and_restores_its_state(self) -> None:
+        seq = list(range(1, 7))
+
+        threshold = 2
+        cost_fn = lambda x: x
```
artemru, line 191: let's take a least one example with different cost function (like x**2 for instance)

native/src/fairseq2n/data/data_pipeline.cc
```diff
@@ -391,6 +392,27 @@ data_pipeline_builder::bucket_by_length(
...
+    cost_fn fn,
+    std::optional<std::size_t> maybe_nb_min,
+    std::optional<std::size_t> maybe_nb_max,
+    bool drop_remainder) &&
+{
+    if (threshold <= 0)
+        throw_<std::invalid_argument>("`threshold` must be greater than zero.");
+    if (maybe_nb_max && *maybe_nb_max == 0)
+        throw_<std::invalid_argument>("`nb_max` must be greater than zero.");
+
```
artemru, line 407: do you check somewhere that `nb_max>=nb_min` ?

src/fairseq2/data/data_pipeline.py
```diff
@@ -231,6 +231,35 @@ def collate(
             See :py:class:`fairseq2.data.Collater` for details.
             """
 
+        def dynamic_bucket(
+            self,
+            threshold: float,
+            cost_fn: Callable[[Any], float],
```
artemru, line 237: i think we may add an extra check that makes sure that `cost_fn` function returns only non-negative scores.

native/src/fairseq2n/data/data_pipeline.cc
syleshfb: i'll add this check

native/python/src/fairseq2n/bindings/data/data_pipeline.cc
```diff
@@ -472,6 +472,25 @@ def_data_pipeline(py::module_ &data_module)
...
             py::arg("overrides") = std::nullopt,
             py::arg("num_parallel_calls") = 1)
+        .def(
+            "dynamic_bucket",
+            [](
+                data_pipeline_builder &self,
+                float64 threshold,
+                cost_fn fn,
+                std::optional<std::size_t> maybe_nb_min,
+                std::optional<std::size_t> maybe_nb_max,
```
cbalioglu, line 482: Same as above; `maybe_max_num_examples`.

native/python/src/fairseq2n/bindings/data/data_pipeline.cc
```diff
@@ -472,6 +472,25 @@ def_data_pipeline(py::module_ &data_module)
...
+                bool drop_remainder) -> data_pipeline_builder &
+            {
+                self = std::move(self).dynamic_bucket(threshold, std::move(fn), maybe_nb_min, maybe_nb_max, drop_remainder);
+
+                return self;
+            },
+            py::arg("threshold"),
+            py::arg("fn"),
+            py::arg("nb_min") = std::nullopt,
```
cbalioglu, line 491: Again, `min/max_num_examples`.

native/src/fairseq2n/data/dynamic_bucket_data_source.cc
```diff
@@ -0,0 +1,82 @@
...
+// LICENSE file in the root directory of this source tree.
+
+#include "fairseq2n/data/dynamic_bucket_data_source.h"
+
+#include <vector>
+
+namespace fairseq2n::detail {
+
+dynamic_bucket_data_source::dynamic_bucket_data_source(
+    std::unique_ptr<data_source> &&inner, float64 threshold, cost_fn &&fn, std::optional<std::size_t> maybe_nb_min, std::optional<std::size_t> maybe_nb_max, bool drop_remainder) noexcept
```
cbalioglu, line 14: Having checked the char length, but if this line (or any other line in your PR) is longer than 100 characters, please do a line break. You can check out other files on how we do it.

native/src/fairseq2n/data/dynamic_bucket_data_source.cc
```diff
@@ -0,0 +1,82 @@
...
+{
+    data_list output{};
+
+    if (maybe_nb_min_)
+        output.reserve(*maybe_nb_min_);
+
+    float64 cost = 0;
+
+    auto bucket_ready = [&]() {
+        bool cost_threshold_met = (cost >= threshold_);
```
cbalioglu, line 29: Nit: In this line and the following two lines, you don't really need parenthesis in the expressions. Just:

```
bool cost_threshold_met = cost >= threshold_;
``

is enough.

File: native/src/fairseq2n/data/dynamic_bucket_data_source.cc [hunk 8]
```diff
@@ -0,0 +1,82 @@
...
+        bool minimum_size_met = (maybe_nb_min_ ? (output.size() >= *maybe_nb_min_) : true);
+        bool maximum_size_met = (maybe_nb_max_ ? (output.size() >= *maybe_nb_max_) : false);
+        return maximum_size_met || (cost_threshold_met && minimum_size_met);
```
cbalioglu, line 32: Nit: You can potentially return earlier with `if (cost_threshold_met && minimum_size_met) return;` before running the statement on line 31.

native/src/fairseq2n/data/dynamic_bucket_data_source.cc
```diff
@@ -0,0 +1,82 @@
...
+    };
+
+    while (!bucket_ready()) {
+        std::optional<data> maybe_example = inner_->next();
+        if (!maybe_example)
+            break;
+        cost += invoke_function(*maybe_example);
```
cbalioglu, line 39: Nit: you can directly call `cost_fn_` here since `invoke_function` does not really have any additional logic.

cbalioglu (commented):
Neat! Really solid work. This is one of the P0 items for Artyom's team. So if you can address the (fairly minor) feedback and have an update, I can merge it right away. This is a relatively small, but quite high impact work.

cbalioglu (approved):
LGTM!

PR #619, opened 2024-07-02
Introduce resume_checkpoint_dir to wav2vec2 ASR finetuning

cbalioglu:
This is a follow-up PR to #617 to introduce `resume_checkpoint_dir` to wav2vec2 ASR finetuning recipe.

### DIFF
--- a/src/fairseq2/recipes/wav2vec2/asr/train.py
+++ b/src/fairseq2/recipes/wav2vec2/asr/train.py
@@ -14,8 +14,9 @@ import torch
 from torch import Tensor
 from torch.nn import Module
 
+from fairseq2.assets import default_asset_store
 from fairseq2.assets.utils import retrieve_asset_card
-from fairseq2.checkpoint import FileCheckpointManager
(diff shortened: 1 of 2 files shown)


kauterry:
What's the use-case for this in your opinion? When will we resume ASR fine-tuning with a setting?

src/fairseq2/recipes/wav2vec2/asr/train.py
```diff
@@ -53,21 +54,21 @@
...
     """
 
     # Data
     dataset: Union[str, Path] = "librilight_asr_10h"
-    """The name or path to the asset card of the dataset to train with."""
+    """The name or path to the asset card of the ASR dataset."""
 
     train_split: str = "train"
-    """The name of the dataset split to train with."""
+    """The name of the dataset split."""
```
shagunsodhani, line 68: `The name of the dataset train split`

cbalioglu:
> What's the use-case for this in your opinion? When will we resume ASR fine-tuning with a setting? Wondering if this is useful for pre-training as well.

This setting is useful if you want to continue the training with a different set of hyperparameters (ablation, curriculum learning, so on).

PR #621, opened 2024-07-02
Introduce generation metric bags

cbalioglu:
This PR introduces two new metric bags for sequence generation to be used in first-party recipes.

### DIFF
--- a/src/fairseq2/recipes/common_metrics.py
+++ b/src/fairseq2/recipes/common_metrics.py
@@ -10,16 +10,17 @@ from typing import Any, Dict, Optional
 
 import torch
 from torch import Tensor
-from torcheval.metrics import Mean, Sum
+from torcheval.metrics import Mean, Sum, Throughput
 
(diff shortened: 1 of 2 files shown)


src/fairseq2/recipes/common_metrics.py
```diff
@@ -167,21 +175,168 @@ def update_batch_metrics(self, batch: Seq2SeqBatch) -> None:
...
+
+        self._batch_size = Mean(device=d)
+
+        self._elements_per_batch = Mean(device=d)
+
+        self._num_examples = Sum(device=d)
+
+        self._num_elements = Sum(device=d)
+
+        self._prefill_size = Sum(device=d)
```
shagunsodhani, line 227: Shouldnt this also be a `Mean` ?

cbalioglu: num_elements and prefill_size report the total number of elements (i.e. tokens) processed by the task (e.g. evaluation, inference)

PR #622, opened 2024-07-03
Add speech dataset for wav2vec2 pretraining.

kauterry:
Run this code snippet for testing
```
import torch

from fairseq2.datasets.speech import load_speech_dataset
from fairseq2.gang import setup_default_gang

dataset = load_speech_dataset("librispeech_960h")

gang = setup_default_gang()

data_reader = dataset.create_reader(
    "train",
    gang,
    dtype=torch.float16,
    min_audio_len=32000,
    max_audio_len=250000,
    max_num_elements=1400000,
    normalize_audio=False,
    example_shuffle_window=1000,
    batch_shuffle_window=1000,
    num_accumulate=1,
    num_prefetch=4,
    seed=2,
)

while True:
    try:
        batch = next(data_reader)
        print(batch[0].shape)
    except StopIteration:
        print("End of data reached.")
        break
```
Output:
torch.Size([5, 240320])
torch.Size([5, 242000])
torch.Size([5, 245840])
torch.Size([5, 235040])
torch.Size([5, 237360])
torch.Size([6, 200720])
torch.Size([8, 140800])
torch.Size([5, 234640])
torch.Size([6, 218720])
torch.Size([5, 239840])
torch.Size([8, 128240])
torch.Size([24, 56320])
torch.Size([7, 176561])
torch.Size([5, 233680])
torch.Size([8, 128240])
torch.Size([8, 155920])
torch.Size([5, 233360])

**Check list:**

### DIFF
--- a/src/fairseq2/datasets/speech.py
+++ b/src/fairseq2/datasets/speech.py
@@ -8,6 +8,7 @@ from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Any, Dict, List, Set, cast, final
 
+import numpy as np
 import torch
 from torch import Tensor
 from torch.nn.functional import layer_norm
@@ -17,6 +18,7 @@ from fairseq2.data import (
     CollateOptionsOverride,
     Collater,
     DataPipeline,
+    DataPipelineBuilder,
     FileMapper,
     SequenceData,
     create_bucket_sizes,
@@ -25,28 +27,23 @@ from fairseq2.data import (
 from fairseq2.data.audio import AudioDecoder
 from fairseq2.data.text import (
     StrSplitter,
-    TextTokenizer,
-    default_raw_sentencepiece_tokenizer_loader,
-    load_text_tokenizer,
     read_text,
 )
 from fairseq2.datasets.data_reader import DataPipelineReader, DataReader
 from fairseq2.datasets.error import DatasetError
 from fairseq2.datasets.loader import AbstractDatasetLoader, DelegatingDatasetLoader
 from fairseq2.gang import Gang
-from fairseq2.models.seq2seq import Seq2SeqBatch
 from fairseq2.nn.padding import get_seqs_and_padding_mask
 from fairseq2.typing import DataType, override
 
 
-class AsrDataset(ABC):
-    """Represents an automatic speech recognition dataset."""
+class SpeechDataset(ABC):
+    """Represents an speech dataset."""
 
     @abstractmethod
     def create_reader(
         self,
         split: str,
-        tokenizer: TextTokenizer,
         gang: Gang,
         max_audio_len: int,
         max_num_elements: int,
@@ -60,18 +57,16 @@ class AsrDataset(ABC):
         num_prefetch: int = 1,
         seed: int = 2,
         **extras: Any,
-    ) -> DataReader[Seq2SeqBatch]:
+    ) -> DataReader[Tensor]:
         """Create a dataset reader.
 
         :param split:
             The split to read.
-        :param tokenizer:
-            The tokenizer to encode target text.
         :param gang:
             The gang over which to shard the dataset.
         :param max_audio_len:
             The maximum audio length of each example. Examples longer than
-            this value will be dropped.
+            this value will be cropped.
         :param max_num_elements:
             The maximum number of elements in each batch.
         :param dtype:
@@ -105,7 +100,7 @@ class AsrDataset(ABC):
         """Return the set of splits."""
 
 
-load_asr_dataset = DelegatingDatasetLoader[AsrDataset]()
+load_speech_dataset = DelegatingDatasetLoader[SpeechDataset]()
 
 
 # TODO: FIX, INFER
@@ -114,8 +109,8 @@ npc = 10
 
 # TODO: Work in progress!
 @final
-class GenericAsrDataset(AsrDataset):
-    """Represents a generic manifest-based ASR dataset."""
+class GenericSpeechDataset(SpeechDataset):
+    """Represents a generic manifest-based Speech dataset."""
 
     _dataset_name: str
     _manifest_dir: Path
@@ -140,7 +135,6 @@ class GenericAsrDataset(AsrDataset):
     def create_reader(
         self,
         split: str,
-        tokenizer: TextTokenizer,
         gang: Gang,
         max_audio_len: int,
         max_num_elements: int,
@@ -155,7 +149,7 @@ class GenericAsrDataset(AsrDataset):
         seed: int = 2,
         cached_fd_count: int = 1000,
         **extras: Any,
-    ) -> DataPipelineReader[Seq2SeqBatch]:
+    ) -> DataPipelineReader[Tensor]:
         """
         :param cached_fd_count:
             The maximum number of file descriptors to keep open while reading
@@ -168,9 +162,7 @@ class GenericAsrDataset(AsrDataset):
 
         root_data_dir = self._retrieve_data_directory(split)
 
-        manifest = self._load_manifest(split)
-
-        builder = read_sequence(manifest)
+        builder = self._builder_from_manifest(split)
 
         # Shuffle examples. Must be consistent across all processes.
         if example_shuffle_window != 1:
@@ -184,9 +176,10 @@ class GenericAsrDataset(AsrDataset):
         seed += gang.rank
 
         # Bucket by audio length.
+        # We allow audios greater than max_audio_len since we crop them later manually.
         bucket_sizes = create_bucket_sizes(
             max_num_elements=max_num_elements,
-            max_seq_len=max_audio_len,
+            max_seq_len=max_num_elements,
             min_seq_len=min_audio_len,
             num_seqs_multiple_of=8,
         )
@@ -196,7 +189,7 @@ class GenericAsrDataset(AsrDataset):
             selector="audio_size",
             min_data_len=min_audio_len,
             skip_below_min_examples=True,
-            skip_above_max_examples=True,
+            skip_above_max_examples=False,
         )
 
         # Shuffle buckets.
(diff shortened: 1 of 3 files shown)


src/fairseq2/datasets/speech.py
```diff
@@ -0,0 +1,302 @@
...
+
+
+load_speech_dataset = DelegatingDatasetLoader[SpeechDataset]()
+
+
+# TODO: FIX, INFER
+npc = 10
+
+
+# TODO: Work in progress!
```
kauterry, line 97: Not anymore for this right?

src/fairseq2/datasets/speech.py
```diff
@@ -0,0 +1,302 @@
...
+        file_mapper = FileMapper(root_data_dir, cached_fd_count=cached_fd_count)
+
+        builder.map(file_mapper, selector="[*].audio")
+
+        # Decode audio.
+        audio_decoder = AudioDecoder(dtype=torch.float32 if normalize_audio else dtype)
+
+        builder.map(audio_decoder, selector="[*].audio.data")
+
+        # TODO(balioglu): Check/adjust sample size
```
kauterry, line 198: What dyou mean? This doesn't apply for this dataset right?

src/fairseq2/datasets/speech.py
```diff
@@ -0,0 +1,302 @@
...
+        builder.skip(1)
+
+        field_splitter = StrSplitter(names=["audio", "audio_size"])
+
+        builder.map(field_splitter, num_parallel_calls=npc)
+
+        # Cast audio size to integer.
+        builder.map(int, selector="audio_size")
+
+        # TODO: Use `cache()` op.
```
kauterry, line 285: Not anymore since we don't load the entire manifest in memory right?

src/fairseq2/datasets/speech.py
```diff
@@ -0,0 +1,332 @@
...
+            The name of the dataset.
+        :param manifest_dir:
+            The directory under which the manifest files resides.
+        """
+        self._dataset_name = dataset_name
+        self._manifest_dir = manifest_dir
+
+        self._splits = set()
+
+        for tsv_file in manifest_dir.glob("*.tsv"):
```
orena1, line 124: This thing force to use all tsv in the folder, it there an option to load only few of them, either for debugging or for specific runs

src/fairseq2/recipes/wav2vec2/train.py
```diff
@@ -0,0 +1,360 @@
...
+    """Load a :class:`Trainer` for wav2vec 2.0 model training."""
+    wall_watch = Stopwatch(start=True)
+
+    gang = setup_root_gang(log, monitored=config.monitored_gang)
+
+    checkpoint_manager = FileCheckpointManager(output_dir.joinpath("checkpoints"), gang)
+
+    # Load the dataset.
+    dataset_card = retrieve_asset_card(config.dataset)
+
```
orena1, line 188: Do you need the extra empty line here?

src/fairseq2/recipes/wav2vec2/train.py
```diff
@@ -0,0 +1,360 @@
...
+    log.info("Loading {} speech dataset.", dataset_card.name)
+
+    dataset = load_speech_dataset(config.dataset)
+
```
orena1, line 192: and also here? I think the logging can be line-attached to the function that it suprvise

src/fairseq2/recipes/wav2vec2/train.py
```diff
@@ -0,0 +1,360 @@
...
+        ddp_find_unused_parameters=False,
+        fsdp_skip_init=True,
+        fsdp_broadcast_state=not has_checkpoint,
+        fsdp_mixed_precision_dtype=config.dtype,
+        fsdp_fp32_reduce=True,
+        fsdp_wrap_granularity=config.fsdp_wrap_granularity,
+    )
+
+    if config.torch_compile:
+        model.encoder = compile_model(model.encoder, log)  # type: ignore[assignment]
```
orena1, line 235: Why compile only the encoder?

cbalioglu:
> Hi, thanks! General question about the dataloader: This assumes that all files are saved as individual audio files in the file system. Is this approach scalable for future use, especially as pre-training data requirements grow? Are there any plans to support other storage solutions, such as tar or parquet?

Hi @orena1, you are totally right. This PR is mainly intended for the parity work with the original fairseq implementation. For better scalability, we internally use two different techniques which we will eventually upstream to this recipe as well. The first technique uses non-compressed zip files to bundle multiple audio files together. We already have support for reading such files using `FileMapper` (see [here](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/data/data_pipeline.py#L471)). The other (more recent) technique is to leverage Parquet files instead of plain TSV files. We also have preliminary support for reading Parquet files in fairseq2 which we plan to extend in the near future. Unfortunately our documentation right now is almost non-existent, but starting next week we plan to invest a lot more time writing proper documentation.

src/fairseq2/datasets/speech.py
```diff
@@ -0,0 +1,332 @@
...
+    @abstractmethod
+    def splits(self) -> Set[str]:
+        """Return the set of splits."""
+
+
+load_speech_dataset = DelegatingDatasetLoader[SpeechDataset]()
+
+
+# TODO: FIX, INFER
+npc = 10
```
shagunsodhani, line 100: what is this ?

src/fairseq2/recipes/wav2vec2/train.py
shagunsodhani, line 235: should this be `ddp_model.encoder` ?

src/fairseq2/datasets/speech.py
kauterry: Number of parallel calls for the collater.

src/fairseq2/recipes/wav2vec2/train.py
kauterry: @cbalioglu we should fix this in the ASR pipeline too right?

orena1:
> > Hi, thanks! General question about the dataloader: This assumes that all files are saved as individual audio files in the file system. Is this approach scalable for future use, especially as pre-training data requirements grow? Are there any plans to support other storage solutions, such as tar or parquet?
> 
> Hi @orena1, you are totally right. This PR is mainly intended for the parity work with the original fairseq implementation. For better scalability, we internally use two different techniques which we will eventually upstream to this recipe as well. The first technique uses non-compressed zip files to bundle multiple audio files together. We already have support for reading such files using `FileMapper` (see [here](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/data/data_pipeline.py#L471)). The other (more recent) technique is to leverage Parquet files instead of plain TSV files. We also have preliminary support for reading Parquet files in fairseq2 which we plan to extend in the near future. Unfortunately our documentation right now is almost non-existent, but starting next week we plan to invest a lot more time writing proper documentation.

Thanks a lot! happy to see how it progress, and how exactly you handle random access with zip, thanks [🙏](https://emojipedia.org/folded-hands)

src/fairseq2/models/wav2vec2/frontend.py
```diff
@@ -161,9 +161,11 @@ def extract_features(
         if self.feature_extractor is not None:
             seqs, padding_mask = self.feature_extractor(seqs, padding_mask)
 
+        feature_penalty = seqs.float().pow(2).mean()
```
cbalioglu, line 164: I suggest:

```python
raw_features = seqs

seqs = self.post_extract_layer_norm(seqs)

return seqs, padding_mask, raw_features
```

and computing the `float().pow(2).mean()` (or better yet use `torch.linalg.vector_norm()`) only during loss calculation if the penalty weight is greater than zero.

src/fairseq2/models/wav2vec2/frontend.py
```diff
@@ -129,15 +129,15 @@ def forward(
...
+        seqs, padding_mask, _ = self.extract_features(seqs, padding_mask)
 
         seqs, padding_mask, _ = self.process_features(seqs, padding_mask)
 
         return seqs, padding_mask
 
     def extract_features(
         self, seqs: Tensor, padding_mask: Optional[PaddingMask]
-    ) -> Tuple[Tensor, Optional[PaddingMask]]:
+    ) -> Tuple[Tensor, Optional[PaddingMask], Tensor]:
```
cbalioglu, line 140: We should coordinate with Arrival since this would be a breaking change. @shagunsodhani

src/fairseq2/models/wav2vec2/model.py
```diff
@@ -42,6 +42,8 @@ class Wav2Vec2Model(Model):
     final_target_proj: Linear
     num_distractors: int
     logit_temp: float
+    diversity_loss_weight: float
```
cbalioglu, line 45: See my comment below. I think loss weights should only be specified as parameters of `compute_loss()`.

src/fairseq2/models/wav2vec2/model.py
```diff
@@ -124,19 +130,23 @@ def extract_features(self, batch: SequenceBatch) -> Wav2Vec2Features:
         :param batch:
             The batch of sequences to process.
         """
-        seqs, padding_mask, targets, temporal_mask = self.run_frontend(
+        seqs, padding_mask, targets, temporal_mask, feature_penalty = self.run_frontend(
```
cbalioglu, line 133: Again, this has to be coordinated with Arrival. @shagunsodhani

src/fairseq2/metrics/accuracy.py
```diff
@@ -0,0 +1,67 @@
...
+import torch
+from torch import Tensor
+from torcheval.metrics import Metric
+from typing_extensions import Self
+
+from fairseq2.typing import Device, override
+
+
+@final
+class AccuracyMetric(Metric[Tensor]):
```
cbalioglu, line 20: Any chance of using one of the implementations in https://github.com/pytorch/torcheval/blob/main/torcheval/metrics/classification/accuracy.py instead of this implementation?

src/fairseq2/models/wav2vec2/frontend.py
shagunsodhani: +1 to returning `raw_features` (and not `feature_penalty`). @cbalioglu do you want to consider returning an object here (instead of a tuple of tensors) ?

src/fairseq2/models/wav2vec2/model.py
```diff
@@ -359,7 +376,13 @@ class Wav2Vec2Output:
...
     sequence length."""
 
-    def compute_loss(self, *, diversity_loss_weight: float = 0.1) -> Wav2Vec2Loss:
+    feature_penalty: Tensor
+
+    diversity_loss_weight: float = 0.1
+
+    feature_penalty_weight: float = 10.0
+
+    def compute_loss(self) -> Wav2Vec2Loss:
```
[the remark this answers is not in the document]
kauterry: These parameters change with different configurations of wav2vec2, for instance the large_lv60k wav2vec2 pretraining config sets the feature_penalty_weight to 0.0. How do we achieve this?

src/fairseq2/models/wav2vec2/frontend.py
cbalioglu: Yes, I think it makes sense to convert this to a proper dataclass instead of a tuple at this point. Once we land these changes, let's make sure that we reflect them in your repo. We are pretty close to v0.3 release, it would be good timing.

src/fairseq2/models/wav2vec2/model.py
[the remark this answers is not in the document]
cbalioglu: If we include them in recipe configuration, we can have different values for different presets.

src/fairseq2/nn/utils/mask.py
```diff
@@ -171,3 +171,20 @@ def _generate_mask(indices: Tensor, max_row_len: int) -> Tensor:
     bool_mask = torch.full_like(float_mask, False, dtype=torch.bool)
 
     return bool_mask.scatter_(1, indices, True)
+
+
+def get_num_masks(mask: Tensor) -> Tensor:
```
kauterry, line 176: This is a convenience function which returns the number of masks present in a mask tensor of shape (bsz, seq_len). This is implemented in an efficient way, I believe this might be useful for some users.

src/fairseq2/metrics/accuracy.py
kauterry: Done! Thank you for the pointer, removed this unnecessary Metric.

cbalioglu:
Let's keep the branch around for further data loading experiments, but closing this PR. Please continue doing further development on team repo.

PR #625, opened 2024-07-03
Introduce KV cache metrics

cbalioglu:
This PR introduces recipe metrics for tracking KV cache (i.e. incremental state) size and capacity.

### DIFF
--- a/src/fairseq2/generation/beam_search.py
+++ b/src/fairseq2/generation/beam_search.py
@@ -18,7 +18,7 @@ from fairseq2.data import VocabularyInfo
 from fairseq2.generation.generator import (
     AbstractSeq2SeqGenerator,
     AbstractSequenceGenerator,
-    GenerationStats,
+    GenerationCounters,
     Hypothesis,
     Seq2SeqGeneratorOutput,
     SequenceGeneratorOutput,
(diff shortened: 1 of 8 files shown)


cbalioglu:
Hey @antoine-tran, first of all sorry for the breaking change. I wasn't aware that you had a custom generator (I grepped for `SequenceGenerator` in MERES code base instead of `SequenceGeneratorOutput`)

Do you mind opening up a little what you mean by "a good counter to add if the profiling is not set"?

antoine-tran:
no worries, we will add KV cache soon to MERES anyway. My question is only to understand the design choice here: It seems the `counters` is required every time we want to output the sequence. So I was wondering if the profiling is assumed to always be the case in the client code.

In my use case, I just cared about evaluating the models' accuracy (no performance profiling), so I created a dummy Counter that is useless:

https://github.com/fairinternal/MERES_internal/blob/eval/exp_booksum/meres/inference/diffusion_lcm/generator.py#L197-L199

but I was wondering if this is a right way, i.e. if there is a default `Counter` singleton that is good to use then.

cbalioglu:
I see. Yeah, that is totally fine. If you look at the source code of `GenerationCounters`, all fields default to zero. You can just pass a default constructed `GenerationCounters` to your output object and things should just work fine.

PR #626, opened 2024-07-03
Add aggregation metrics

cbalioglu:
This PR subclasses the torcheval aggregation metrics and relaxes the type requirement for their `input` parameter from being Tensor-only to Tensor, int, or float.

### DIFF
--- a/src/fairseq2/recipes/wav2vec2/asr/common.py
+++ b/src/fairseq2/recipes/wav2vec2/asr/common.py
@@ -10,15 +10,15 @@ import math
 
 import torch
 from torch import Tensor
-from torcheval.metrics import Mean, Sum
 
 from fairseq2.gang import Gang
 from fairseq2.metrics import MetricBag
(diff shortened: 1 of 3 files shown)


src/fairseq2/recipes/wav2vec2/asr/common.py
```diff
@@ -64,11 +64,9 @@ def update_ctc_loss(self, batch: Seq2SeqBatch, loss: Tensor) -> None:
         :param ctc_loss:
             The loss of ``batch``.
         """
-        batch_size = torch.tensor(batch.batch_size)
+        normalized_loss = loss / batch.batch_size / math.log(2)
 
-        normalized_loss = loss.cpu() / batch_size / math.log(2)
```
kauterry, line 69: Why was it `loss.cpu()` before?

cbalioglu: Just for a bit of consistency. `batch_size` was a CPU tensor, so wanted them to be on the same device.

PR #646, opened 2024-07-08
saving structured rank based jsonl in generate recipe

uralik:

This PR adds generation output saving in structured jsonl format for easier post-processing and usage. We still want to keep more human readable outputs, so now both streams (txt and jsonl) are used in the recipe.

### DIFF
--- a/src/fairseq2/recipes/lm/text_generate.py
+++ b/src/fairseq2/recipes/lm/text_generate.py
@@ -396,7 +396,6 @@ class TextGenerateUnit(AbstractGeneratorUnit[SequenceBatch]):
         if (
             txt_stream is not None and jsonl_stream is not None
         ):  # Means in the first tensor parallel group.
-
             for prompt, hypotheses in zip(prompts, output.hypotheses):
                 if len(hypotheses) == 0:
                     raise RuntimeError(
--- a/src/fairseq2/recipes/lm/text_generate.py
+++ b/src/fairseq2/recipes/lm/text_generate.py
@@ -6,6 +6,7 @@
 
 from __future__ import annotations
 
+import json
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Literal, Optional, TextIO, Union, final
@@ -106,7 +107,7 @@ class SamplingConfig:
     min_gen_len: int = 1
     """The minimum generation length."""
 
-    max_gen_len: int = 512
+    max_gen_len: int = 2048
     """The maximum generation length."""
 
     max_seq_len: Optional[int] = None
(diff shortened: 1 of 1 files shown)


src/fairseq2/recipes/lm/text_generate.py
```diff
@@ -106,7 +107,7 @@ class SamplingConfig:
     min_gen_len: int = 1
     """The minimum generation length."""
 
-    max_gen_len: int = 512
+    max_gen_len: int = 2048
```
uralik, line 110: changing this to a more reasonable length by default since 512 is too small, and sampler handles EOS nicely, so overgenerations won't be an issue

src/fairseq2/recipes/lm/text_generate.py
```diff
@@ -332,26 +349,32 @@ class TextGenerateUnit(AbstractGeneratorUnit[SequenceBatch]):
...
         self._text_decoder = tokenizer.create_decoder()
 
-        self._output_stream = output_stream
+        self._output_txt_txt_stream = output_txt_txt_stream
+
+        self._output_jsonl_txt_stream = output_jsonl_txt_stream
 
         self._metric_bag = SequenceGenerationMetricBag(gang)
 
+        self._gang = gang
```
uralik, line 376: this made debugging much easier as I can access gang from the `__call__`.

src/fairseq2/recipes/lm/text_generate.py
```diff
@@ -332,26 +349,32 @@ class TextGenerateUnit(AbstractGeneratorUnit[SequenceBatch]):
...
+    _output_jsonl_txt_stream: Optional[TextIO]
     _metric_bag: SequenceGenerationMetricBag
 
     def __init__(
         self,
         generator: SequenceGenerator,
         tokenizer: TextTokenizer,
         gang: Gang,
-        output_stream: Optional[TextIO],
+        output_txt_txt_stream: Optional[TextIO],
```
cbalioglu, line 361: I suggest naming these: `output_text_stream` and `output_json_stream`. In particular, `output_txt_txt_stream` sounds a little off to me.

src/fairseq2/recipes/lm/text_generate.py
cbalioglu: Just curious, how were you using `gang` within `__call__` during debugging? Wondering whether we can streamline it.

cbalioglu (commented):
Left only a few nit comments. I see some linter complaints though. Once you address them, it is good to go.

src/fairseq2/recipes/lm/text_generate.py
uralik: maybe there is a better way, what I did during the run

```
if self._gang.rank == 0: import pdb; pdb.set_trace()

self._gang.barrier()
```

is there a better way to do this without having the gang object in the scope?

src/fairseq2/recipes/lm/text_generate.py
uralik: oh that sounds like a bad ctrf-f -> replace :)

src/fairseq2/recipes/lm/text_generate.py
cbalioglu: You might consider using the global `get_rank()` function defined [here](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/gang.py#L572). It is not dependent on `Gang` and can be used pretty much anywhere (e.g. deep inside a call stack)

cbalioglu (approved):
LGTM!

PR #654, opened 2024-07-10
[lm generate] save id if provided in the jsonl file

uralik:

Introduce "id" in the example in both src+tgt and prompt readers of the generic instruction dataset.
Then text generate recipe saves the id field into the jsonl file.

Tested using both dataset with IDs and without, in the latter case id field will have null (None) value.


cbalioglu (approved):
LGTM

PR #656, opened 2024-07-10
Nit prep fixes for MT recipe

cbalioglu:
Two nit fixes before landing the machine translation recipe.

### DIFF
--- a/src/fairseq2/datasets/parallel_text.py
+++ b/src/fairseq2/datasets/parallel_text.py
@@ -33,7 +33,7 @@ from fairseq2.nn.padding import get_seqs_and_padding_mask
 from fairseq2.typing import Device, override
 
 
-@dataclass(frozen=True)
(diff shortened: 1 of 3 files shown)


src/fairseq2/datasets/parallel_text.py
```diff
@@ -33,7 +33,7 @@
 from fairseq2.typing import Device, override
 
 
-@dataclass(frozen=True)
+@dataclass(unsafe_hash=True)  # Due to FSDP, we cannot freeze.
```
shagunsodhani, line 36: This is pretty interesting. Do you have a reference for this ?

cbalioglu: I don't think there is a written reference. In fact, this is mostly an unintentional side-effect of FSDP implementation, since it blindly tries to set attributes on model input/outputs.

shagunsodhani: Thank you for the context!

PR #658, opened 2024-07-10
Fix DinoSR checkpoint loading

syleshfb:
Fixes DinoSR tokenizer checkpoint loading with fairseq2.models.wav2vec2.load_wav2vec2_model


mavlyutovr (approved):

cbalioglu (approved):
Thanks for the PR! More an organizational question. Since we now host Dinosr implementation in our internal repository, wondering whether it would make sense to move this architecture definition to there.

Otherwise, the PR itself looks good to me.

PR #659, opened 2024-07-11
[lm generate ]setting args to be more default

uralik:

* changed topp topk temperature to a more default values


cbalioglu (approved):
Thanks!

PR #667, opened 2024-07-13
Introduce support for early stopping in Trainer

cbalioglu:
This PR introduces support for early-stopping in `Trainer`. A follow-up PR will introduce support for `keep_best_n_checkpoints` in `CheckpointManager`.

### DIFF
--- a/src/fairseq2/recipes/lm/instruction_finetune.py
+++ b/src/fairseq2/recipes/lm/instruction_finetune.py
@@ -138,7 +138,7 @@ class InstructionFinetuneConfig:
     """The step interval at which to checkpoint."""
 
     keep_last_n_checkpoints: Optional[int] = 1
-    """The number of checkpoints to keep. If ``None``, none will be deleted."""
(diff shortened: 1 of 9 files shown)


src/fairseq2/recipes/lm/instruction_finetune.py
```diff
@@ -138,7 +138,7 @@ class InstructionFinetuneConfig:
     """The step interval at which to checkpoint."""
 
     keep_last_n_checkpoints: Optional[int] = 1
-    """The number of checkpoints to keep. If ``None``, none will be deleted."""
```
shagunsodhani, line 141: If the `None` behaviour is no longer supported, maybe we can restrict the type to be just int ?

cbalioglu: Oh, let me fix the docstr. The behavior has not changed. I was updating the docstrs for the upcoming training recipe and apparently I accidentally removed the second sentence.

Issue #353, opened 2024-02-28
`from_generator` function for BuilderDataPipeline


artemru:
**Is your feature request related to a problem? Please describe:**
Currently there's only `read_sequence` method that requires to know all list of sample in advance,
`from_generator` would offer more flexible interface useful in various contexts (such as async processing).

artemru:
done with [#696](https://github.com/facebookresearch/fairseq2/pull/696)

PR #696, opened 2024-07-21
Iterator data source 

syleshfb:
Introduces a read_iterator data pipeline factory that uses an Iterator as a data source.

**Check list:**


native/python/src/fairseq2n/bindings/data/iterator_data_source.cc
```diff
@@ -0,0 +1,78 @@
...
+    return (*iterator_)++->cast<py_object>();
+}
+
+void
+iterator_data_source::reset(bool)
+{
+    py::gil_scoped_acquire acquire;
+
+    reloaded_ = false;
+    reset_fn_(*iterator_);
```
cbalioglu, line 35: I think the whole `reset` approach to the Python iterator needs a bit rethinking. There are many cases where the user has no way to reset a Python iterator. Think for instance that they pass a `range(1234)` or `enumerate(other_iter)` to `read_iterator()`, there is no formal API in Python to "reset" such iterators. I think the safer approach would be for `reset_fn` (or whatever we call it) to return a new Python iterator. It will give the user the flexibility to return a new iterator (like in `range`, `enumerate`, or `itertools` iterators), or if they have a custom reset API, just call it, and return the same instance.

cbalioglu (commented):
Thanks again for this PR! Overall looks nice. I think the way how we handle resetting a Python iterator needs a different approach though. I left my feedback in `iterator_data_source::reset`. Let me know what you think.

cbalioglu (approved):
Really solid, thanks a lot again! The reset functionality looks much nicer now. LGTM!

artemru:
I'll need to read it in more details but I think this PR closes https://github.com/facebookresearch/fairseq2/issues/353 !

native/python/src/fairseq2n/bindings/data/iterator_data_source.cc
```diff
@@ -0,0 +1,92 @@
...
+    py::function pickle_dump_fn = py::module::import("pickle").attr("dumps");
+    t.record(pickle_dump_fn(*iterator_).cast<py_object>());
+
+    t.record(to_return);
+
+    t.record(reset_);
+}
+
+void
+iterator_data_source::reload_position(tape &t, bool)
```
artemru, line 68: I think we need to add into the doc string explicitly that iterator should be picklable (and correctly) since the pipeline state will be serialized using pickle (which can make weird thing sometimes). 
I wonder if we do want to add some custom function for state_load/state_save along with `reset_fn`.
One different approach that could make sense in some case (where the elements production is fast but serialization can be complicated) to restore the state is to do `reset` + `skip` the number of element that were already produced.

syleshfb: >I think we need to add into the doc string explicitly that iterator should be picklable (and correctly) since the pipeline state will be serialized using pickle (which can make weird thing sometimes).

PR forthcoming for this---the plan is to add a allow_unpickleable flag that is False by default (in which case it will raise an error on `__init__`), if that sounds good.

>I wonder if we do want to add some custom function for state_load/state_save along with reset_fn.
One different approach that could make sense in some case (where the elements production is fast but serialization can be complicated) to restore the state is to do reset + skip the number of element that were already produced.

I believe the user can already do this through `__getstate__` and `__setstate__` methods in the iterator class. For example:

```python
from fairseq2.data import read_iterator
from typing import Iterator

class A(Iterator):
    def __init__(self):
        self.counter = 0
        self.generator = iter(range(100))
    def __iter__(self):
        return self
    def __next__(self):
        self.counter += 1
        return next(self.generator)
    def __getstate__(self):
        return self.counter
    def __setstate__(self, counter):
        self.generator = iter(range(100))
        for i in range(counter):
            next(self.generator)
        self.counter = counter

pipeline = read_iterator(A(), reset_fn = lambda x : A(), infinite=False).and_return()
it = iter(pipeline)
print(next(it), next(it), next(it))
d = pipeline.state_dict()
print(next(it), next(it), next(it))
pipeline.load_state_dict(d)
print(next(it), next(it), next(it))
``` 
unless you'd prefer this built-in.

PR #708, opened 2024-07-26
feat: Introduce CLI Interface `eval` for Model Evaluation

Ahmedsaed:
This PR introduces a new CLI interface named `eval` to the Fairseq2 framework. The eval CLI aims to streamline the process of evaluating machine learning models, providing users with an efficient and user-friendly method to assess model performance.

Changes:
- Add a CLI group `eval`
- Add `wav2vec2-asr` command to evaluating wav2vec2 models
- Implement `load_wav2vec2_asr_evaluator` along with other helper functions to 
     - load and preprocess dataset
     - load model and tokenizer
     - construct and initialize HFEvaluator
- Implement HFEvaluator
    - Load and combine huggingface metrics to effectively evaluate models

Here's an example of how to use it:
```bash
fairseq2 eval wav2vec2-asr /tmp/fairseq2 --config max_samples=2 # for debugging
```

Sample output:
```
...
[07/26/24 15:32:11] INFO     fairseq2.recipes.evaluator - Running evaluation on 1 device(s).                                                                                               
[07/26/24 15:32:38] INFO     fairseq2.recipes.evaluator - Eval Metrics - BLEU: 0.586519 | Elapsed Time: 26s | Wall Time: 27s | brevity_penalty: 0.9428731438548749 | length_ratio:         
                             0.9444444444444444 | precisions: [0.7647058823529411, 0.6666666666666666, 0.5384615384615384, 0.5454545454545454] | reference_length: 18 | translation_length:
                             17                                                                                                                                                            
                    INFO     fairseq2.recipes.evaluator - Evaluation complete in 27 seconds                                                                                                
```

I have created this [notebook](https://colab.research.google.com/drive/1vxynmAHknfrw1P_Ka-b9dm28Hf-uva9q?usp=sharing) to demonstrate the use of the new eval command

**Check list:**
- [X] Was the content of this PR **discussed and approved** via a GitHub issue? (no need for typos or documentation improvements)
- [X] Did you read the [contributor guideline](https://github.com/facebookresearch/fairseq2/blob/main/CONTRIBUTING.md)?
- [X] Did you verify new and **existing tests pass** locally with your changes?


Ahmedsaed:
> Just a fundamental question before jumping into a technical review. How is this evaluation recipe different than the existing wav2vec2_asr eval recipe we already have in fairseq2?

The new evaluation recipe differs from the existing wav2vec2_asr eval recipe in Fairseq2 in that it is designed to leverage Hugging Face datasets and evaluation libraries. This new interface aims to facilitate evaluating models across various tasks and metrics. Additionally, Future pull requests (PRs) will expand support to include more models and metrics.

antoine-tran:
> Just a fundamental question before jumping into a technical review. How is this evaluation recipe different than the existing wav2vec2_asr eval recipe we already have in fairseq2?

To add up to Ahmed's response: If I'm not mistaken, the existing wav2vec2_asr eval recipe is for evaluating the (ongoing) model with respect to some loss / criterion closely related to the training loop. On the other hand, what we develop here is the downstream evaluation of the final model using standard metrics such as BLEU, RougeL, etc.

src/fairseq2/datasets/huggingface.py
```diff
@@ -18,6 +18,7 @@
     )
 except ImportError:
     has_datasets = False
+    Dataset = Any
```
antoine-tran, line 21: A better option would be to make the module fail fast with the explicit message for the ModuleNotFoundError (lines [81-84](https://github.com/facebookresearch/fairseq2/pull/708/files#diff-97566db29a807a8ee911a6217dfed4f0b5cc60003492927a515b874262024fb0R81-R84))

Ahmedsaed: I believed the purpose of the try-except block was to enable the CLI to perform tasks that don’t rely on the datasets library without encountering import errors. If I place the ModuleNotFoundError at the top level, it will be raised even when simply printing the help message with `fairseq2 --help`.

antoine-tran: We can isolate the list of tasks in fairseq2.recipes.eval.__init__._setup_eval_cli() that uses HuggingFace imports, and copy the explicitt message for the ModuleNotFoundError there.

src/fairseq2/recipes/evaluator.py
```diff
@@ -274,3 +278,178 @@ def _publish_metrics(self, unit: EvalUnit[BatchT], elapsed_time: float) -> None:
...
+                    break
+
+                progress.update(eval_task, refresh=True, advance=1)
+
+                log.debug("Running step {}.", step_nr)
+
+                for batch in batches:
+                    inputs, targets = self._preprocessor(batch)
+                    outputs = self._model(inputs)
+                    hypotheses, _ = outputs.generate_hypotheses(pad_idx=pad_idx)
```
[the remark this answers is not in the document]
Ahmedsaed: I have pushed a commit with the requested changes https://github.com/facebookresearch/fairseq2/pull/708/commits/2bd9d5b439b4679cf2b0ebd30cbc3d0612e8bf95

src/fairseq2/datasets/huggingface.py
antoine-tran: Something like this:

src/fairseq2/recipes/eval/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+        "wav2vec2-asr",
+        handler,
+        help="evaluate a wav2vec 2.0 ASR model in downstream benchmark",
+    )
+
+
+def _setup_eval_cli(cli: Cli) -> None:
+    group = cli.add_group("eval", help="Evaluate fairseq2 models in downstream tasks")
+
+    _add_wav2vev2_asr_eval_cli(group)
```
antoine-tran, line 28: See my other comments. It might be better that we add sub-command conditionally on the availability of huggingface's datasets or evaluate libs

antoine-tran:
This PR looks good to me, once we resolve the (minor) issue with the dynamic loading of hf libs such as `datasets`.

Another thing is we maintain 2 `Evaluator` and `HFEvaluator` in tandem, I understand that this is because @cbalioglu decided to have Evaluator as a final class.

@cbalioglu : Do you think we should maintain 2 variants of evaluators here, or are you fine removing the "final" tag and consolidate them further ?

src/fairseq2/datasets/huggingface.py
Ahmedsaed: Thank you for the guidance! I’ve implemented the suggested changes in this commit: https://github.com/facebookresearch/fairseq2/pull/708/commits/bc89baa9785f349da3bebd88b111499a910ed396

src/fairseq2/recipes/eval/__init__.py
Ahmedsaed: I’m still getting familiar with handling import exceptions in this way, so I appreciate your patience.

Do you expect something like this?

```python
try:
    _add_wav2vev2_asr_eval_cli(group)
except ImportError as e:
    # Log the reason for skipping this command
```

However, wouldn't this approach raise an exception for all sub-commands since the main idea behind this eval recipe is to use huggingface evaluate

antoine-tran: > In other words, no sub-commands will be available if hf libs are not installed unless this recipe is going to be extended with commands that doesn't require hf libs

No, fairseq2.recipes.eval is an umbrella to test fairseq2 models , with or without using huggingface datasets. The public version here relies on public benchmarks that are available via huggingface, but we can extend to other benchmarks / metrics too.

And only the hf-based recipes are isolated within the try-except block

Ahmedsaed:
1- I noticed this warning while running wav2vec2-asr

```
                    WARNING  datasets.fingerprint - Parameter 'function'=<function load_wav2vec2_asr_evaluator.<locals>.<lambda> at 0x740de80ed1b0> of the transform                  
                             datasets.arrow_dataset.Dataset._map_single couldn't be hashed properly, a random hash was used instead. Make sure your transforms and parameters are     
                             serializable with pickle or dill for the dataset fingerprinting and caching to work. If you reuse this transform, the caching mechanism will consider it 
                             to be different from the previous calls and recompute everything. This warning is only showed once. Subsequent hashing failures won't be showed.         ```

File: src/fairseq2/recipes/eval/asr.py [hunk 4]
```diff
@@ -0,0 +1,221 @@
...
+    if not isinstance(config, AsrEvalConfig):
+        raise ValueError(f"Expect AsrEvalConfig, get {type(config)}")
+
+    iterable_ds = load_dataset(config.dataset_name, split=config.split, streaming=True)
+    max_samples = config.max_samples if config.max_samples is not None else math.inf
+    # Load a subset of the dataset if max_samples is set
+    ds = Dataset.from_generator(
+        lambda: (
+            yield from (
+                item for idx, item in enumerate(iterable_ds) if idx < max_samples
```
antoine-tran, line 174: `itertools.islice(iterable_ds, 0, config.max_samples)`

antoine-tran (commented):
I left some comments. Can you also resolve the conflicts with the main branch ? Thanks

src/fairseq2/recipes/eval/asr.py
```diff
@@ -0,0 +1,221 @@
...
+    return Seq2SeqBatch(
+        source_seqs,
+        source_padding_mask,
+        target_seqs,
+        target_padding_mask,
+        examples,
+    )
+
+
+def _preprocess_example(
```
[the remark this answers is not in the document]
Ahmedsaed: This function is called for each example, resulting in the tokenizer being loaded repeatedly, which can impact the performance and clutter the logs with messages like `fairseq2.assets.download_manager - Using the cached tokenizer of librispeech_asr.` for every single example. However, this approach does address the caching issue, so preprocessing won't be repeated in subsequent runs.

Is this acceptable?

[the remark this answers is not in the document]
Ahmedsaed: I have addressed all requested in https://github.com/facebookresearch/fairseq2/pull/708/commits/abc7f5c18ec8ced1fedd056ef62366bc9333c054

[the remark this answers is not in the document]
antoine-tran: I see. We can make a separate function decorated with lru-cache to generate the decoder from its tokenizer name.

src/fairseq2/datasets/huggingface.py
```diff
@@ -4,29 +4,25 @@
...
+    Dataset,
+    DatasetDict,
+)
 
 from fairseq2.data.data_pipeline import Collater, create_bucket_sizes, read_sequence
 from fairseq2.datasets.batching import LengthBatching, StaticBatching
 from fairseq2.datasets.data_reader import BatchT, DataPipelineReader
 from fairseq2.gang import Gang
 
-try:
```
antoine-tran, line 14: hmm, why do we want to remove this safeguard here ?

Although we have implemented the similar safeguard in fairseq2 CLI, this package in `fairseq2.datasets` can be consumed by other non-CLI packages / client too.

src/fairseq2/recipes/eval/__init__.py
```diff
@@ -0,0 +1,59 @@
...
+        return True
+    except ImportError:
+        log.warning(
+            "`datasets` is required but not found. Please install it with `pip install datasets`. "
+            "Some functions will be disabled"
+        )
+        return False
+
+
+def has_evaluate() -> bool:
```
antoine-tran, line 42: Good, thanks !

src/fairseq2/recipes/eval/asr.py
```diff
@@ -0,0 +1,223 @@
...
+        output_dir (Path): The output directory to store the evaluation results.
+
+    Returns:
+        HFEvaluator: Evaluation process results.
+    """
+    if not isinstance(config, AsrEvalConfig):
+        raise ValueError(f"Expect AsrEvalConfig, get {type(config)}")
+
+    iterable_ds = load_dataset(config.dataset_name, split=config.split, streaming=True)
+    max_samples = cast(
```
antoine-tran, line 174: We don't need this. If `max_samples` is None, `islice` will iterate until the end

src/fairseq2/recipes/eval/asr.py
```diff
@@ -0,0 +1,223 @@
...
+        init_device = META
+
+    ds = ds.map(lambda x: _preprocess_example(x, config.tokenizer_name, init_device))
+    format = {
+        "type": "torch",
+        "format_kwargs": {"dtype": torch.float16, "device": init_device},
+    }
+    ds.set_format(**format, columns=["audio", "text"])
+
+    tokenizer = load_text_tokenizer(config.tokenizer_name)
```
antoine-tran, line 197: As said, we can make this as part of the lru-cached function too

src/fairseq2/recipes/eval/asr.py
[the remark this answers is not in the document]
Ahmedsaed: Thanks for the suggestion. This worked as expected. Now, caching works and the tokenizer is loaded once.

src/fairseq2/datasets/huggingface.py
Ahmedsaed: Alright!. I have reverted this back https://github.com/facebookresearch/fairseq2/pull/708/commits/6bb31febe0b67f061bc8dca8179d0580751abc6f

src/fairseq2/recipes/eval/asr.py
Ahmedsaed: Done! https://github.com/facebookresearch/fairseq2/pull/708/commits/6bb31febe0b67f061bc8dca8179d0580751abc6f

PR #709, opened 2024-07-30
Simple wrapper for read_iterator pickleability

syleshfb:
Adds a utility function, read_and_pickle_wrap_iterator, to which any iterator/generator can be passed. 
This will wrap the iterator/generator in a class whose pickle behavior is to track an internal counter for the
number of times `__next__()` is called, and to replay this during reloading.
If the given iterator is already pickleable, the wrapping is elided.

N/A

**Check list:**


src/fairseq2/utils/read_and_pickle_wrap_iterator.py
```diff
@@ -0,0 +1,48 @@
...
+
+from typing_extensions import Self
+
+from fairseq2.data import DataPipelineBuilder, read_iterator
+
+T = TypeVar("T")
+
+
+class IteratorPickleWrapper(Iterator[T]):
+    def __init__(self, iterator_factory: Callable[[], Iterator[T]]) -> None:
```
cbalioglu, line 11: You might use a type alias for `Callable[[], Iterator[T]]` to make the code more readable:

```python
from typing_extensions import TypeAlias

IteratorFactory: TypeAlias = Callable[[], Iterator[T]]

class IteratorPickleWrapper(Iterator[T]):
  def __init__(self, factory: IteratorFactory[T]) -> None:
    ...
```

src/fairseq2/utils/read_and_pickle_wrap_iterator.py
```diff
@@ -0,0 +1,48 @@
...
+        self.iterator_factory = iterator_factory
```
cbalioglu, line 12: I suggest prefixing the member variables with an underscore to indicate that they are private state:

```python
self._iterator_factory = ...
self._iterator = ...
self._counter = ...
```

src/fairseq2/utils/read_and_pickle_wrap_iterator.py
```diff
@@ -0,0 +1,48 @@
+from typing import Callable, Iterator, Tuple, TypeVar
```
cbalioglu, line 1: We recently moved to Python 3.10 and updated the code base to use some of the modern syntax features that came with 3.9 and 3.10.

We now import `Callable` and `Iterator` from `collections.abc`, and instead of `Tuple`, `List`, and `Dict`, we now use `tuple`, `list`, and `dict` in typing annotations (e.g. `foo: tuple[int, str]`). You can check out the code base to see their uses.

It would be nice if you could make the relevant changes in this PR as well. I see a few `Tuple`s at least that can be changed to `tuple`.

src/fairseq2/utils/read_and_pickle_wrap_iterator.py
```diff
@@ -0,0 +1,48 @@
...
+
+def read_and_pickle_wrap_iterator(
+    iterator_factory: Callable[[], Iterator[T]]
+) -> DataPipelineBuilder:
+    iterator = iterator_factory()
+    try:
+        return read_iterator(
+            iterator, reset_fn=lambda x: iterator_factory(), infinite=False
+        )
+    except TypeError:
```
cbalioglu, line 43: Is there a way for us to make this a bit more fail-proof? A `TypeError` might also be raised for other reasons. The most naive way would be to check the error message (e.g. `str(ex).startswith("pickling error message from CPython")` if there is no other way to distinguish.

src/fairseq2/utils/__init__.py
```diff
@@ -0,0 +1,3 @@
+from fairseq2.utils.read_and_pickle_wrap_iterator import (
```
cbalioglu, line 1: I would prefer this helper to reside under `fairseq2.data.utils` instead of `fairseq2.utils`.

src/fairseq2/utils/read_and_pickle_wrap_iterator.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def __setstate__(self, state: Tuple[Callable[[], Iterator[T]], int]) -> None:
+        self.iterator_factory, counter = state
+        self.iterator = self.iterator_factory()
+        for i in range(counter):
+            next(self.iterator)
+        self.counter = counter
+
+
+def read_and_pickle_wrap_iterator(
```
cbalioglu, line 35: How about `read_pickle_wrapped_iterator()`?

cbalioglu (commented):
Overall looks good to me. I left mostly nit comments. It would be also nice if you could include some (brief) docstrings. In particular, we should make sure that the user is aware that state restoration requires linear time unlike most other data pipeline ops.

src/fairseq2/utils/__init__.py
syleshfb: See the latest change - let me know how exactly you want the path to look like.

sysuresh:
Please close this in favor of #754

PR #733, opened 2024-08-06
Move autograd detach to loss metrics

cbalioglu:
A small PR that handles autograd detaching of loss tensors within metric bags to be on the safe side.

### DIFF
--- a/src/fairseq2/nn/utils/gradient.py
+++ b/src/fairseq2/nn/utils/gradient.py
@@ -75,7 +75,7 @@ class _GradientScaleFunction(Function):
 
         ctx.scale = scale
 
-        return x.detach().clone().requires_grad_(True)
+        return x.detach().clone()
 
     @staticmethod
     def backward(ctx: Any, grad_output: Tensor) -> tuple[Tensor, None]:  # type: ignore[override]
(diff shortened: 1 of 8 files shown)


src/fairseq2/nn/utils/gradient.py
```diff
@@ -75,7 +75,7 @@ def forward(ctx: Any, x: Tensor, scale: float) -> Tensor:  # type: ignore[overri
 
         ctx.scale = scale
 
-        return x.detach().clone().requires_grad_(True)
```
kauterry, line 78: Why do we detach and clone here? Why not just return x? This removes x from the autograd graph right, why is it necessary?

src/fairseq2/recipes/wav2vec2/asr/train.py
```diff
@@ -511,7 +511,7 @@ def __call__(self, batch: Seq2SeqBatch) -> tuple[Tensor, int]:
 
         loss = output.compute_loss(batch.target_seqs, batch.target_padding_mask)
 
-        self._metric_bag.update_ctc_loss(batch, loss.detach())
```
kauterry, line 514: Why do we do this? I don't understand.

PR #734, opened 2024-08-06
Support feature penalty in wav2vec2 loss

cbalioglu:
This PR updates the implementation of wav2vec2 to support feature penalty during pretraining. The majority of the work has been performed by @kauterry in #622. This is mostly a cleaned-up and slightly optimized version of his implementation. The bitwise parity has been verified by a local script that compares loss and model gradients.

### DIFF
--- a/src/fairseq2/models/wav2vec2/asr/factory.py
+++ b/src/fairseq2/models/wav2vec2/asr/factory.py
@@ -58,7 +58,7 @@ class Wav2Vec2AsrConfig:
     temporal_mask_span_len: int = 10
     """The length of each temporal mask span that is applied over time steps."""
 
-    max_temporal_mask_prob: float = 0.70
+    max_temporal_mask_prob: float = 0.69
     """The maximum probability of masking a time step. Note that, due to mask
     span overlap, the effective probability will be lower."""
 
--- a/src/fairseq2/models/wav2vec2/factory.py
+++ b/src/fairseq2/models/wav2vec2/factory.py
@@ -76,7 +76,7 @@ class Wav2Vec2Config:
     temporal_mask_span_len: int = 10
     """The length of each temporal mask span that is applied over time steps."""
 
-    max_temporal_mask_prob: float = 0.70
+    max_temporal_mask_prob: float = 0.69
     """The maximum probability of masking a time step. Note that, due to mask
     span overlap, the effective probability will be lower."""
 
--- a/src/fairseq2/models/wav2vec2/factory.py
+++ b/src/fairseq2/models/wav2vec2/factory.py
@@ -76,7 +76,7 @@ class Wav2Vec2Config:
     temporal_mask_span_len: int = 10
     """The length of each temporal mask span that is applied over time steps."""
 
-    max_temporal_mask_prob: float = 0.65
+    max_temporal_mask_prob: float = 0.70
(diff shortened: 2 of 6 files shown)


src/fairseq2/models/wav2vec2/model.py
```diff
@@ -286,7 +271,7 @@ def _compute_logits(
         if distractor_is_target.any():
             logits[:, :, 1:][distractor_is_target] = -torch.inf
 
-        return logits.type_as(seqs)
```
kauterry, line 289: Why remove this? fairseq1 has this.

src/fairseq2/models/wav2vec2/frontend.py
```diff
@@ -153,27 +153,31 @@ def extract_features(
...
             - The padding mask of the extracted features. *Shape:*
               :math:`(N,S_{out})`, where :math:`N` is the batch size and
               :math:`S_{out}` is the output sequence length.
+            - The raw features. *Shape*: Same as the normalized features (i.e.
+              first element of the returned tuple).
         """
         if self.feature_extractor is not None:
             seqs, padding_mask = self.feature_extractor(seqs, padding_mask)
 
+        raw_features = seqs
```
kauterry, line 169: Do we need a .clone() here? Will `raw_features` change when `seqs` changes?

cbalioglu: We do not do any in-place ops, but yeah, let me do a clone to be on the safe side.

src/fairseq2/models/wav2vec2/model.py
cbalioglu: It is redundant since we are casting back to float32 for cross_entropy anyways.

kauterry (approved):
Thanks!

src/fairseq2/models/wav2vec2/asr/factory.py
```diff
@@ -58,7 +58,7 @@ class Wav2Vec2AsrConfig:
     temporal_mask_span_len: int = 10
     """The length of each temporal mask span that is applied over time steps."""
 
-    max_temporal_mask_prob: float = 0.70
+    max_temporal_mask_prob: float = 0.69
```
shagunsodhani, line 61: Wondering if this change if just to use the default values from the paper or if there are other implications here.

src/fairseq2/models/wav2vec2/factory.py
```diff
@@ -76,7 +76,7 @@ class Wav2Vec2Config:
     temporal_mask_span_len: int = 10
     """The length of each temporal mask span that is applied over time steps."""
 
-    max_temporal_mask_prob: float = 0.65
+    max_temporal_mask_prob: float = 0.69
```
shagunsodhani, line 79: Wondering if this change if just to use the default values from the paper or if there are other implications here.

kauterry: I checked the average masking rate with fairseq1, 0.69 gets us the closest to the hyperparam for wav2vec2 in fairseq1. This is just to match the default average masking rate in fairseq1 as specifid in the paper.

shagunsodhani: Makes sense.

PR #738, opened 2024-08-09
Dynamic bucket creation customization options

syleshfb:
Add a new optional argument to dynamic_bucket---bucket_creation_fn---to allow the user to customize how buckets are created. 

bucket_creation_fn allows the user to customize what dynamic_bucket yields as a bucket once the cost threshold is met as well as what remains in the subsequent bucket.

N/A

**Check list:**

### DIFF
--- a/native/python/src/fairseq2n/bindings/data/data_pipeline.cc
+++ b/native/python/src/fairseq2n/bindings/data/data_pipeline.cc
@@ -484,6 +484,8 @@ def_data_pipeline(py::module_ &data_module)
                 data_pipeline_builder &self,
                 float64 threshold,
                 cost_fn fn,
+                std::optional<bucket_creation_fn> maybe_bucket_fn,
+                std::optional<postprocess_remainder_fn> maybe_remainder_fn,
                 std::optional<std::size_t> maybe_min_num_examples,
                 std::optional<std::size_t> maybe_max_num_examples,
                 bool drop_remainder) -> data_pipeline_builder &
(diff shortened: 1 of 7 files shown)


cbalioglu:
Do we have a specific use case where these two new parameters are required? If I recall correctly, we had a discussion about whether to include the example that exceeds the threshold within the returned bucket or to leave it in the buffer for the next bucket. It makes sense to expose it as an option. Just curious what other use cases we have to require this level of flexibility.

syleshfb:
> Do we have a specific use case where these two new parameters are required? If I recall correctly, we had a discussion about whether to include the example that exceeds the threshold within the returned bucket or to leave it in the buffer for the next bucket. It makes sense to expose it as an option. Just curious what other use cases we have to require this level of flexibility.

I took a closer look at the use case in question and it seems like we need more flexibility than just including/excluding the last example in a threshold-exceeding bucket. 

For example, we might need to extract the highest multiple of a desired batch size from a bucket that causes the cost threshold to be exceeded, and leave the rest to be included the subsequent bucket. So, it's not necessarily just excluding the last example, but dynamically deciding how much of the tail should be excluded. Seems like the user needs to have a mechanism to, for example, yield only part of the bucket that exceeds the threshold and to keep the rest for the subsequent bucket. And to deal with remainders specially (in this particular use case, there's a need to selectively keep or discard remainders).

sysuresh:
Please close this in favor of #753

PR #742, opened 2024-08-12
Make quantizer encoder grad configurable

mshvartsman:

The fix in #731 caused a regression in some of our training pipelines (but I also know that it improved other training pipelines). So it seems like maybe whether gradients are propagated from the quantizer to the encoder should be configurable. Not sure if this is a big enough change to have its own issue, happy to add one if so (same with tests). 

No -- defaults should remain the same. 

**Check list:**


cbalioglu (approved):
Thanks for the PR! Looks good to me.

PR #740, opened 2024-08-09
Integrate Whisper model with `hg` evaluation CLI interface

Ahmedsaed:
**What does this PR do? Please describe:**
This PR adds integration for whisper model with the  `hg` evaluation CLI interface.

In the process, I have refactored some functions to be more generic and support huggingface transformers API.

**Demo**:
```bash
$ fairseq2 hg asr /tmp/fairseq2 --config max_num_elements=20 model_name=openai/whisper-tiny.en dtype=torch.float32
```
```
[08/09/24 19:00:49] INFO     fairseq2.recipes.hg.evaluator - Running evaluation on 1 device(s).     
eval: ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   147   
[08/09/24 19:03:37] INFO     fairseq2.recipes.hg.evaluator - Eval Metrics - BLEU: 0.497937 | Elapsed
                             Time: 167s | Wall Time: 169s | brevity_penalty: 1.0 | length_ratio:    
                             1.1557228282673901 | precisions: [0.676166231361788,                   
                             0.5496829119972201, 0.4500613362139993, 0.3675025152851946] |          
                             reference_length: 52343 | translation_length: 60494                    
                    INFO     fairseq2.recipes.hg.evaluator - Evaluation complete in 169 seconds     
```

**Does your PR introduce any breaking changes? If yes, please list them:**
Hopefully none.

**Check list:**
- [X] Was the content of this PR **discussed and approved** via a GitHub issue? (no need for typos or documentation improvements)
- [X] Did you read the [contributor guideline](https://github.com/facebookresearch/fairseq2/blob/main/CONTRIBUTING.md)?
- [X] Did you make sure that your **PR does only one thing** instead of bundling different changes together?
- [X] Did you make sure to **update the documentation** with your changes? (if necessary)
- [X] Did you write any **new necessary tests**?
- [X] Did you verify new and **existing tests pass** locally with your changes?
- [X] Did you **update the [CHANGELOG](https://github.com/facebookresearch/fairseq2/blob/main/CHANGELOG.md)**? (no need for typos, documentation, or minor internal changes)

PR #749, opened 2024-08-15
Add support for other datasets through `hg` recipes

Ahmedsaed:
This PR builds upon #740 by incorporating the refactoring changes here and adding new commits that enable support for defining other datasets via configuration.

Example:
```
fairseq2 hg asr /tmp/fairseq2 --config max_samples=2 dataset_config.dataset_path=google/fleurs dataset_config.
dataset_name=en_us dataset_config.target_column=['transcription'] # override default preset
```

```
[08/15/24 03:55:44] INFO     fairseq2.recipes.hg.evaluator - Running evaluation on 1 device(s).                                                                                  
[08/15/24 03:57:55] INFO     fairseq2.recipes.hg.evaluator - Eval Metrics - BLEU: 0 | Elapsed Time: 130s | Wall Time: 131s | brevity_penalty: 1.0 | length_ratio: 1.0 |          
                             precisions: [0.375, 0.15789473684210525, 0.05555555555555555, 0.0] | reference_length: 40 | translation_length: 40                                  
                    INFO     fairseq2.recipes.hg.evaluator - Evaluation complete in 131 seconds                                                                                  
```

This evaluates wav2vec2 model on google/fleurs dataset by just overriding the relevant configs.

None.

**Check list:**
- [X] Did you make sure to **update the documentation** with your changes? (if necessary)
- [X] Did you verify new and **existing tests pass** locally with your changes?
- [X] Did you **update the [CHANGELOG](https://github.com/facebookresearch/fairseq2/blob/main/CHANGELOG.md)**? (no need for typos, documentation, or minor internal changes)

### DIFF
--- a/src/fairseq2/recipes/hg/asr_eval.py
+++ b/src/fairseq2/recipes/hg/asr_eval.py
@@ -5,10 +5,10 @@
 # LICENSE file in the root directory of this source tree.
 
 import itertools
+from collections.abc import Callable
 from dataclasses import dataclass
-from functools import lru_cache
 from pathlib import Path
-from typing import Any, cast
+from typing import Any, Optional, cast
 
 import torch
 from datasets import (  # type: ignore[attr-defined,import-untyped,import-not-found]
@@ -31,7 +31,7 @@ from fairseq2.recipes.hg.dataset import Example, create_hf_reader
 from fairseq2.recipes.hg.evaluator import HFEvaluator
 from fairseq2.recipes.utils.asset import retrieve_asset_card
 from fairseq2.recipes.utils.setup import setup_root_gang
-from fairseq2.typing import META, DataType
+from fairseq2.typing import META, DataType, Device
 from fairseq2.utils.profiler import Stopwatch
 
 log = get_log_writer(__name__)
@@ -48,9 +48,6 @@ class AsrEvalConfig:
     # Model
     model_name: str
     """The name of the model to evaluate."""
-
-    # converter: Callable[[Example], Seq2SeqBatch]
-    # """The converter function to convert collated data into Seq2SeqBatch"""
 
     tokenizer_name: str = "librispeech_asr"
     """The tokenizer to use."""
@@ -95,10 +92,10 @@ def _default_asr_config() -> AsrEvalConfig:
         dataset_name="librispeech_asr",
         model_name="wav2vec2_asr_base_10h",
         split="test.other",
-        # converter=librispeech_asr_to_batch,
     )
 
-def _librispeech_asr_to_batch(examples: Example) -> Seq2SeqBatch:
+
+def to_batch(examples: Example, model_type: str, device: Device) -> Seq2SeqBatch:
     """
     Converts a collated batch of examples into a Seq2SeqBatch.
 
@@ -111,11 +108,19 @@ def _librispeech_asr_to_batch(examples: Example) -> Seq2SeqBatch:
     source_data = cast(SequenceData, examples["audio"])
     target_data = cast(SequenceData, examples["text"])
 
-    source_seqs, source_padding_mask = get_seqs_and_padding_mask(source_data)
-    target_seqs, target_padding_mask = get_seqs_and_padding_mask(target_data)
+    if model_type == "wav2vec2":
+        source_seqs, source_padding_mask = get_seqs_and_padding_mask(source_data)
+    elif model_type == "whisper":
+        source_seqs = cast(source_data)
+        source_padding_mask = None
+    else:
+        raise ValueError(f"Unknown model type: {model_type}")
+
+    target_seqs = target_data
+    target_padding_mask = None
 
     return Seq2SeqBatch(
-        source_seqs,
+        source_seqs.to(device),
         source_padding_mask,
         target_seqs,
         target_padding_mask,
@@ -123,42 +128,28 @@ def _librispeech_asr_to_batch(examples: Example) -> Seq2SeqBatch:
     )
 
 
-@lru_cache(maxsize=None)
-def get_cached_tokenizer(tokenizer_name: str) -> TextTokenizer:
-    return load_text_tokenizer(tokenizer_name)
-
-
-def _preprocess_example(
-    example: Example, tokenizer_name: str, device: torch.device
-) -> Example:
+def extract_features(example: Example) -> Example:
     """
     Preprocesses an individual example by converting the audio array to a PyTorch tensor
     and encoding the text.
 
     Args:
         example (dict): A dictionary containing "audio" and "text" keys.
-        tokenizer_name (str): The name of the tokenizer to use.
         device (torch.device): The device to store the tensors.
 
     Returns:
         dict: A dictionary with "audio" and "text" as PyTorch tensors.
     """
-    tokenizer = get_cached_tokenizer(tokenizer_name)
-    encoder = tokenizer.create_encoder(device=device)
-    audio_tensor = (
-        torch.from_numpy(example["audio"]["array"]).to(torch.float16).to(device)
-    )
-    text_tensor = encoder(example["text"].lower()).to(device)
-    return {"audio": audio_tensor, "text": text_tensor}
+    return {"audio": example["audio"]["array"], "text": example["text"].lower()}
 
 
-def seq2seq_preprocessor(batch: Seq2SeqBatch) -> tuple[SequenceBatch, SequenceBatch]:
+def evaluator_preprocessor(batch: Seq2SeqBatch) -> tuple[SequenceBatch, SequenceBatch]:
     return SequenceBatch(batch.source_seqs, batch.source_padding_mask), SequenceBatch(
         batch.target_seqs, batch.target_padding_mask
     )
 
 
-def postprocesser(
+def evaluator_postprocesser(
     outputs: Any, targets: SequenceBatch, tokenizer: TextTokenizer
 ) -> tuple[list[str], list[str]]:
     decoder = tokenizer.create_decoder()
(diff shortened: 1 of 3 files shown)


src/fairseq2/recipes/hg/asr_eval.py
```diff
@@ -17,44 +18,86 @@
...
 from fairseq2.utils.profiler import Stopwatch
 
 log = get_log_writer(__name__)
 
 
+@dataclass
+class AsrDatasetConfig:
+    """Configuration for an automatic speech recognition dataset."""
+
+    dataset_path: str
```
cbalioglu, line 41: I found it confusing what exactly `dataset_path` and `dataset_name` mean here. Their docstrings (along with the docstr of `split`) does not make much sense.

src/fairseq2/recipes/hg/asr_eval.py
```diff
@@ -17,44 +18,86 @@
...
+    """The name of the dataset."""
+
+    dataset_name: Optional[str] = None
+    """The name of the dataset split."""
+
+    split: str = "test"
+    """The name of the dataset split to evaluate with."""
+
+    source_column: List[str] = field(default_factory=list)
```
cbalioglu, line 50: If we expect this to accept multiple column names, I would prefer `source_columns` and `target_columns`.

src/fairseq2/recipes/hg/asr_eval.py
```diff
@@ -17,44 +18,86 @@
...
+    """The path of the column containing the source audio."""
+
+    target_column: List[str] = field(default_factory=list)
+    """The path of the column containing the target text."""
+
+    @classmethod
+    def from_dict(cls, config_dict: dict[str, Any]) -> "AsrDatasetConfig":
```
cbalioglu, line 57: Please add a `from __futures__ import annotation` at the top of the file (see other files in the code base), and use the type name as the return type annotation (i.e. `-> AsrDatasetConfig`)

src/fairseq2/recipes/hg/asr_eval.py
```diff
@@ -17,44 +18,86 @@
...
+        results = self._get_data(ds, self.source_column)
+        return results
+
+    def get_target_data(self, ds: Example) -> Any:
+        """Retrieve the target (text) data from the dataset."""
+        results = self._get_data(ds, self.target_column)
+        return results
+
+    @staticmethod
+    def _get_data(ds: Example, path: List[str]) -> Union[Example, List[int], str]:
```
cbalioglu, line 78: Nit: please use Python 3.9/3.10 syntax for type annotations:

```python
def _get_data(ds: Example, path: list[str]) -> Example | list[int] | str:
  ...
```

src/fairseq2/recipes/hg/evaluator.py
```diff
@@ -172,6 +173,11 @@ def _do_run(self) -> None:
                         predictions=predictions, references=references
                     )
 
+                    del inputs
+                    del targets
+                    del outputs
+                    torch.cuda.empty_cache()
```
cbalioglu, line 179: What is the reason for this explicit `empty_cache()` call?

src/fairseq2/recipes/hg/asr_eval.py
```diff
@@ -17,44 +18,86 @@
...
+        """Create an AsrDatasetConfig instance from a configuration dictionary."""
+        return cls(
+            dataset_path=config_dict.get("dataset_path", ""),
+            dataset_name=config_dict.get("dataset_name"),
+            source_column=config_dict.get("source_column", []),
+            target_column=config_dict.get("target_column", []),
+            split=config_dict.get("split", "test"),
+        )
+
+    def get_source_data(self, ds: Example) -> Any:
```
cbalioglu, line 67: I would prefer `get_source_data` and `get_target_data` to be standalone functions (along with `_get_data`), instead of being part of the config type. A configuration dataclass should ideally contain only fields and functions (e.g. `from_dict`) to construct instances.

src/fairseq2/recipes/hg/asr_eval.py
```diff
@@ -82,97 +125,133 @@ class AsrEvalConfig:
     """The data type of the model."""
 
 
-asr_eval_presets = ConfigRegistry[AsrEvalConfig]()
+@dataclass
+class EvalSeqBatch:
```
cbalioglu, line 129: Any reason for not using `Seq2SeqBatch` [here](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/models/seq2seq.py#L53) like the native ASR evaluator [here](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/recipes/wav2vec2/asr/eval.py#L232)

src/fairseq2/recipes/hg/asr_eval.py
```diff
@@ -82,97 +125,133 @@ class AsrEvalConfig:
...
     Returns:
-        Seq2SeqBatch: A batch of audio and text sequences.
+        EvalSeqBatch: A batch of audio and text sequences.
     """
-    source_data = cast(SequenceData, examples["audio"])
-    target_data = cast(SequenceData, examples["text"])
 
-    source_seqs, source_padding_mask = get_seqs_and_padding_mask(source_data)
-    target_seqs, target_padding_mask = get_seqs_and_padding_mask(target_data)
+    if model_type == "wav2vec2":
```
cbalioglu, line 187: I assume any ASR model would expect the same input right? An audio sequence and its corresponding transcription. Any reason why we explicitly check a model type here?

src/fairseq2/recipes/hg/asr_eval.py
Ahmedsaed: This is the docs from the huggingface documentation. 

```
- path (str) — Path or name of the dataset. Depending on path, the dataset builder that is used comes from a generic dataset script (JSON, CSV, Parquet, text etc.) or from the dataset script (a python file) inside the dataset directory.
- name (str, optional) — Defining the name of the dataset configuration.
- split (Split or str) — Which split of the data to load. If None, will return a dict with all splits (typically datasets.Split.TRAIN and datasets.Split.TEST). If given, will return a single Dataset. Splits can be combined and specified like in tensorflow-datasets.
```

I will correct the docstrings to match the huggingface docs

src/fairseq2/recipes/hg/asr_eval.py
Ahmedsaed: Not exactly. The `source_column` and `target_column` parameters define the paths to the source and target data within the dataset structure.

Consider the following dataset structures:

- **LibriSpeech:**
    ```json
    dataset: {
        audio: {
            array: [AUDIO_WAVE_DATA]
        },
        text: [TEXT_DATA]
    }
    ```
    - Source data: `dataset['audio']['array']`
    - Target data: `dataset['text']`

- **Google/Fleurs:**
    ```json
    dataset: {
        audio: {
            array: [AUDIO_WAVE_DATA]
        },
        transcriptions: [TEXT_DATA]
    }
    ```
    - Source data: `dataset['audio']['array']`
    - Target data: `dataset['transcriptions']`

- **Another Dataset:**
    ```json
    dataset: {
        audio: {
            waveform: {
                array: [AUDIO_WAVE_DATA]
            },
            other_format: {
                array: [AUDIO_WAVE_DATA]
            }
        },
        text: {
            raw: {
                data: [TEXT_DATA]
            }
        }
    }
    ```
    - Source data: `dataset['audio']['waveform']['array']`
    - Target data: `dataset['text']['raw']['data']`

To make the code adaptable to different datasets, regardless of their structure, the `AsrDatasetConfig` can be used to specify the paths to the source and target data as lists. This allows for greater flexibility and compatibility across various dataset formats.

For example:

- **Dataset 1:**
    - Source: `['audio', 'array']`
    - Target: `['text']`
  
- **Dataset 2:**
    - Source: `['audio', 'array']`
    - Target: `['transcriptions']`
  
- **Dataset 3:**
    - Source: `['audio', 'waveform', 'array']`
    - Target: `['text', 'raw', 'data']`

Is there a better approach to handle this? Should I rename the fields to `source_path` and `target_path` instead?

src/fairseq2/recipes/hg/asr_eval.py
Ahmedsaed: After reconsidering, I've decided to remove this method entirely. I initially planned to use it, but later realized that `@dataclass` already provides the same functionality.

I have done this here https://github.com/facebookresearch/fairseq2/pull/749/commits/dc684f03a6b91f858d8a5f106e30013cdcee7d79

src/fairseq2/recipes/hg/asr_eval.py
Ahmedsaed: I have addressed this in https://github.com/facebookresearch/fairseq2/pull/749/commits/dc684f03a6b91f858d8a5f106e30013cdcee7d79

src/fairseq2/recipes/hg/asr_eval.py
Ahmedsaed: I have addressed this in https://github.com/facebookresearch/fairseq2/pull/749/commits/dc684f03a6b91f858d8a5f106e30013cdcee7d79

src/fairseq2/recipes/hg/evaluator.py
Ahmedsaed: I noticed that the data from previous batches were not freed which lead to the evaluator crashing OOM after a some batches.

Quoting https://github.com/facebookresearch/fairseq2/pull/740#issuecomment-2278592218
> I noticed that running the evaluation with `max_samples > 20` on a 15GB VRAM setup was causing a CUDA out-of-memory error. To address this, I made improvements to better manage and free up memory. With these updates, you can now run the evaluation on the entire dataset, provided the `batch_size` fits within the available memory. I successfully tested this on the whole test split (approximately 2939 examples) using `whisper-tiny` and `wav2vec2_asr_base_10h` with `max_num_elements=10`.
>...
> 
> Edit: `max_num_elements` can also be increased significantly by defining `max_audio_len` as some audio samples were quite long and the whole batch gets padded to the longest sequence causing some batches to require a lot more memory than expected.

src/fairseq2/recipes/hg/asr_eval.py
Ahmedsaed: The `Seq2SeqBatch` class expects targets to be tensors, which is useful for training. However, during evaluation, encoding the target only to decode it again without utilizing the tensors seems redundant.

Is there a better approach? another fairseq2 data-structure to use?

src/fairseq2/recipes/hg/asr_eval.py
Ahmedsaed: Yes, but the required input formats differ. For wav2vec2 models, the input should be in the `SequenceBatch` format, including `seqs` and `padding_mask`. In contrast, Whisper (huggingface) model expect just an array of input sequences.

This was part of the refactoring that was made to allow for the integration of more models in future PRs like https://github.com/facebookresearch/fairseq2/pull/751

src/fairseq2/recipes/hg/asr_eval.py
Ahmedsaed: The `get_source_data` and `get_target_data` functions are essentially wrappers around `_get_data` to simplify its usage. Removing these functions from the class would eliminate their purpose. Instead, I can refactor the code to move `_get_data` outside the class and use it directly.

Do you want me to do this?

src/fairseq2/recipes/hg/asr_eval.py
Ahmedsaed: In https://github.com/facebookresearch/fairseq2/pull/749/commits/667194fd968c82f46ff43e85aaf57f13127e3be7, I have renamed the fields to `source_column_path` and `target_column_path`

I'll mark the comment as resolved, but please let me know if this change doesn't align with your expectations.

src/fairseq2/recipes/hg/asr_eval.py
Ahmedsaed: In https://github.com/facebookresearch/fairseq2/pull/749/commits/667194fd968c82f46ff43e85aaf57f13127e3be7, I have moved the `_get_data` outside of the dataclass, renamed it to `_get_column_data`, removed both `get_source_data` and `get_target_data` and updated the code to use it instead.

I will mark this as resolved

PR #756, opened 2024-08-19
fix assert order bug

orena1:
can not access layer_descs[-1][0] is len(layer_descs)==0

@cbalioglu


cbalioglu (approved):
Thanks for the fix! LGTM.

PR #776, opened 2024-08-27
Fix wav2vec2 loss metrics

cbalioglu:
This PR fixes the loss metric calculation bug in wav2vec2 pretraining.


kauterry:
What bug is this?

PR #778, opened 2024-08-27
Set family in model builders

cbalioglu:
A small PR that calls `set_family` in model builders' `build_model()`.


kauterry:
This is a refactoring change, and doesn't change functionality right?

cbalioglu:
No

PR #818, opened 2024-09-28
llms recipe config updates 

uralik:

* making sure the default presets will have dropout 0 
* preset names wont include model size unless its 70b preset where model is changed explicitly
* remove unnecessary/outdated preset 
* include default DPO config in the DPO preset


cbalioglu (approved):
LGTM!

PR #836, opened 2024-10-03
llama 3.2 text models added

uralik:

archs and assets added for llama 3.2 text models

verified via training 3.2 1b model with DPO


cbalioglu (approved):
Nice! Thanks!

PR #840, opened 2024-10-19
Add asset integrity check for asset downloads

### LINKED ISSUE #839: Support checksum in AssetDownloadManager
As of today, the [`InProcAssetDownloadManager`](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/assets/download_manager.py#L118) does not have any checks to make sure that downloaded assets (e.g. model checkpoints, datasets, tokenizers) have not been tampered with. Ideally we would like to have an optional "checksum" field in asset cards that the download manager can use to verify the integrity of a downloaded file.

```python
from fairseq2.assets import default_asset_store, InProcDownloadManager

card = default_asset_store.retrieve_card("mistral_7b")

checkpoint_uri = card.field("checkpoint").as_uri()

# We should check if the card has a checksum field; if not, `get_as_` returns None.
checksum = card.field("checksum").get_as_(str)

download_manager = InProcDownloadManager()

download_manager.download_checkpoint(checkpoint_uri, model_name="mistral")

# The API of the download manager should be extended with an additional "checksum" parameter.
# download_manager.download_checkpoint(checkpoint_uri, checksum, model_name="mistral")
```

PeanutButterRat:
This PR adds a file integrity check for tokenizers, models, and datasets to ensure they were downloaded properly with a SHA-1 checksum. If the checksum listed on the asset card doesn't match, the `download_manager` will raise an `AssetDownloadError`. If there is no listed checksum on the asset card, the `download_manager` will proceed like normal but make a call to `log.warning` about the missing checksum field.

Fixes #839

Adds an additional `checksum` parameter to the `download_manager`'s methods for downloading tokenizers, models, and datasets.

**Check list:**

### DIFF
--- a/src/fairseq2/metrics/recorder.py
+++ b/src/fairseq2/metrics/recorder.py
@@ -489,7 +489,7 @@ else:
 class WandbRecorder(MetricRecorder):
     """Records metric values to Weights & Biases."""
 
-    def __init__(self, project: str, output_dir: Path) -> None:
+    def __init__(self, project: str, name: str, output_dir: Path) -> None:
         """
         :param project: The W&B project name.
         :param output_dir: The base directory under which to store the W&B files.
@@ -502,7 +502,9 @@ class WandbRecorder(MetricRecorder):
 
             self._run = None
         else:
-            self._run = wandb.init(project=project, dir=output_dir.parent)
+            self._run = wandb.init(
+                project=project, name=name, dir=output_dir.parent, resume="allow"
+            )
 
     @override
     def record_metrics(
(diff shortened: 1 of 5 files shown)


PeanutButterRat:
I added a check to see if the checksum is `None` before calculating the hash. If there is no checksum on the asset card, it will just skip the integrity check and make a call to `log.warning`.

PeanutButterRat:
@cbalioglu If the checksum fails, should the asset files be cleaned up, or is just raising an exception acceptable?

cbalioglu (approved):
LGTM!

PR #841, opened 2024-10-23
adding run name option for wandb, adding resume allow

uralik:
* wandb was not taking any run name before, that makes it harder to group runs within e.g. the sweep.
* currently wandb_run_name is another config name that is parsed by the legacy config parser.
* wandb init resume allow was added so that interrupted runs can resume logging as long as the project and name are the same.


cbalioglu (approved):
LGTM!

PR #843, opened 2024-10-29
Update PyTorch THPVariable_Wrap to use const ref

huydhn:
This comes from a change from PyTorch https://github.com/pytorch/pytorch/pull/138880 in which the signature of `THPVariable_Wrap` changes to 

```
TORCH_PYTHON_API PyObject * THPVariable_Wrap(const at::TensorBase& var);
```

A build failure happens when importing the change to fbcode D65031075 for one project that is using `fairseq2`

This makes fairseq2 compatible with latest PyTorch, so I don't think there is any breaking change here.

**Check list:**

### DIFF
--- a/native/python/src/fairseq2n/bindings/type_casters/torch.cc
+++ b/native/python/src/fairseq2n/bindings/type_casters/torch.cc
@@ -20,7 +20,7 @@ struct THPVariable {
 
 extern PyObject *THPVariableClass;
 
-PyObject *THPVariable_Wrap(at::TensorBase var);
+PyObject *THPVariable_Wrap(const at::TensorBase& var);
 
 // Taken from <torch/bindings/Device.h>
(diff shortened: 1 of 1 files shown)


huydhn:
I think we have a chick-and-egg issue here.  So here is the current interlocking setup:

* Some Meta-internal code uses pytorch (latest) and fairseq2 (imported from this repo).  They fails because these twos are not compatible anymore w.r.t `THPVariable_Wrap`
* fairseq2 is open source and it tests against stable PyTorch >= 2.2.0 https://github.com/facebookresearch/fairseq2/blob/main/.github/workflows/ci_build_wheels.yaml#L33-L52.  This is not latest PyTorch.

I think I need to try writing an internal patch instead of waiting for this PR.

huydhn:
I will close this for now as the internal patch works (but brittle).  I'll leave it to fairseq2 maintainers to decide when and how they want to make fairseq2 compatible with PT when it comes to this public function

PR #846, opened 2024-10-31
Add src/tgt encode options for inst finetuning

niansong1996:
Add options in instruction finetuning to select the tokenizer encode mode for src and tgt sequences, which allows you to control what special tokens are added.

For certain use cases (e.g., multi-turn conversations), we would like to add our own special tokens, so I also added an option `as_is` to not add any special tokens, and this can be set in the `instruction_finetune` recipe.

Since this is a small change, no tests are written, but I have tried with the following config using the `instruction_finetune` recipe and the training seems to be starting just fine:
```
...
src_encode_mode: as_is
tgt_encode_mode: prompt_response
...
```

### DIFF
--- a/src/fairseq2/datasets/instruction.py
+++ b/src/fairseq2/datasets/instruction.py
@@ -56,6 +56,8 @@ class InstructionDataset(ABC):
         max_num_batches: int | None = None,
         num_accumulate: int = 1,
         num_prefetch: int = 1,
+        src_encode_mode: str = "prompt",
+        tgt_encode_mode: str = "prompt_response",
         seed: int = 2,
         **extras: Any,
     ) -> DataReader[SequenceBatch]:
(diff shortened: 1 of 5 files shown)


uralik:
@cbalioglu  me and @niansong1996  discussed this offline and this will be a very useful functionality to use.

@niansong1996 once you test this out and it is working as expected, could you also add similar option for the preference finetune recipe so that we can train dpo-like setups with configurable modes? thanks!

niansong1996:
Just added the same options for the preference finetuning. 

@uralik @cbalioglu Let me know if the changes look good to you!

niansong1996:
Adding that the new options are (functionally) tested with runs both for dpo and instruction finetuning recipes.

uralik (approved):
thanks for adding this useful feature!

PR #851, opened 2024-11-07
introduce batch_size and static batching to instruction finetune recipe

uralik:

This is specifically useful when users train on very small datasets with variable size sequences. In this case dynamic batching is hard to adjust when e.g. we just need batch size 1.

This mimics the preference finetune recipe logic. Tested on SFT training with gsm8k.

This will be tested from branch by team before merging.


cbalioglu (approved):
LGTM!

PR #854, opened 2024-11-22
more ffn multipliers added

uralik:

Fixing wrong ffn multiplier being written in the params json for various llama models

### DIFF
--- a/src/fairseq2/recipes/llama/convert_checkpoint.py
+++ b/src/fairseq2/recipes/llama/convert_checkpoint.py
@@ -153,8 +153,19 @@ class ConvertCheckpointCommandHandler(CliCommandHandler):
             if model_config.num_attn_heads != model_config.num_key_value_heads:
                 params["model"]["n_kv_heads"] = model_config.num_key_value_heads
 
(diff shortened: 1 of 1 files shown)


src/fairseq2/recipes/llama/convert_checkpoint.py
```diff
@@ -153,8 +153,19 @@ def __call__(self, args: Namespace, resolver: DependencyResolver) -> None:
...
+                "llama2_70b": 1.3,
+                "llama3_8b": 1.3,
+                "llama3_70b": 1.3,
+                "llama3_1_8b": 1.3,
+                "llama3_1_70b": 1.3,
+                "llama3_1_405b": 1.2,
+                "llama3_2_1b": 1.5,
+            }
+
+            if args.arch in ffn_dim_multipliers:
```
cbalioglu, line 167: Not necessarily for this PR, but should we print a warning message if the architecture name is not known? The quickest way would be to do some basic string/regex check. A more sophisticated way would be to check the LLaMA ConfigRegistry.

cbalioglu (approved):
Thanks for the fix!

cbalioglu (approved):
Nice! Thanks for the fix!

PR #871, opened 2024-12-05
Fix string splitter to return a string when a single index is provided

MartinGleize:
String splitter now returns a string when a single index is provided as argument.
This relieves the user from unpacking the previous returned singleton list.

Fixes #{issue number}

No

**Check list:**

### DIFF
--- a/native/src/fairseq2n/data/text/string_splitter.cc
+++ b/native/src/fairseq2n/data/text/string_splitter.cc
@@ -18,13 +18,11 @@ using namespace fairseq2n::detail;
 namespace fairseq2n {
 
 static std::vector<std::size_t>
-wrapIndices(std::variant<std::size_t, std::vector<std::size_t>> indices) {
+wrapIndices(std::variant<std::size_t, std::vector<std::size_t>> &&indices) {
     if (std::holds_alternative<std::size_t>(indices)) {
-        // Wrap the size_t value in a new vector
-        return {std::get<std::size_t>(indices)};
+        return {std::get<std::size_t>(std::move(indices))};
     } else {
-        // Move the existing vector
-        return std::move(std::get<std::vector<std::size_t>>(indices));
+        return std::get<std::vector<std::size_t>>(std::move(indices));
     }
 }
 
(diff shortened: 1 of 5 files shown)


native/src/fairseq2n/data/text/string_splitter.cc
```diff
@@ -17,17 +17,26 @@ using namespace fairseq2n::detail;
...
+}
+
 string_splitter::string_splitter(
     char separator,
     std::vector<std::string> names,
-    std::vector<std::size_t> indices,
+    std::variant<std::size_t, std::vector<std::size_t>> indices,
     bool exclude)
-  : separator_{separator}, names_(std::move(names)), indices_{std::move(indices)}, exclude_{exclude}
+  : separator_{separator}, names_(std::move(names)), indices_{wrap_indices(std::move(indices))}, exclude_{exclude}, single_column_{std::holds_alternative<std::size_t>(indices)}
```
cbalioglu, line 34: There is a sneaky bug here, that is dormant right now, but might bite in the future if the implementation of `std::variant` changes (or with another C++ stdlib). While assigning a boolean to `single_column_`, you check which alternative `indices` holds; however, at this point `indices` has already been moved earlier during the assignment to `indices_`. So you should consider the state of `indices` as an implementation-specified unknown (but "valid") state.

cbalioglu, line 34: Do you mind also limiting your lines to 100 chars as long as it does not reduce readability?

cbalioglu (commented):
Neat! Looks very clean, thanks a lot! There is just one subtle bug that we should fix.

native/src/fairseq2n/data/text/string_splitter.cc
MartinGleize: Thanks @cbalioglu, nice catch. Split the variant constructor into 2.

MartinGleize: Done.

cbalioglu (approved):
LGTM!

PR #884, opened 2024-12-11
Add the hybrid sharding strategy in FSDP (HSDP)

MartinGleize:
Add the hybrid sharding strategy in FSDP (HSDP).
Integration in recipes is only done and tested on instruction finetune, for now.

Fixes #{issue number}

No

**Check list:**

### DIFF
--- a/src/fairseq2/gang.py
+++ b/src/fairseq2/gang.py
@@ -707,7 +707,7 @@ def fake_gangs(device: Device) -> Gangs:
     return Gangs(gang, gang, gang)
 
 
-def setup_2D_mesh_gangs(
+def _setup_2D_mesh_gangs(
     root_gang: Gang,
     *,
     row_length: int = 1,
@@ -749,21 +749,6 @@ def setup_2D_mesh_gangs(
         A ``dict`` of two gangs; 0 maps to the gang of 2D mesh row,
         1 maps to the gang of the 2D mesh column.
     """
-    if row_length <= 0:
-        raise ValueError(
-            f"`row_length` must be greater than 0, but is {row_length} instead."
-        )
-
-    if row_length > root_gang.size:
-        raise ValueError(
-            f"`row_length` must be less than or equal to `root_gang.size` ({root_gang.size}), but is {row_length} instead."
-        )
-
-    if root_gang.size % row_length != 0:
-        raise ValueError(
-            f"`root_gang.size` ({root_gang.size}) must be divisible by `row_length` ({row_length})."
-        )
-
     row_count = root_gang.size // row_length
 
     mesh = torch.arange(root_gang.size).view(row_count, row_length)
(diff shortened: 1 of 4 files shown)


src/fairseq2/gang.py
```diff
@@ -707,6 +707,113 @@ def fake_gangs(device: Device) -> Gangs:
     return Gangs(gang, gang, gang)
 
 
+def setup_2D_mesh_gangs(
```
cbalioglu, line 710: My preference would be to make this an internal function (i.e. `_setup_2d_mesh_gangs()`). We already expose the actual functionality in `setup_parallel_gangs()` and `to_fsdp()` that are higher-level and abstract away the details of the mesh creation.

src/fairseq2/gang.py
```diff
@@ -707,6 +707,113 @@ def fake_gangs(device: Device) -> Gangs:
...
+        If ``True``, create an underlying ``dist.ProcessGroup`` even for single-rank gangs.
+        The gang is faked otherwise.
+    :param dim_descriptions:
+        String descriptions of returned gangs, used in log and error messages.
+
+    :returns:
+        A ``dict`` of two gangs; 0 maps to the gang of 2D mesh row,
+        1 maps to the gang of the 2D mesh column.
+    """
+    if row_length <= 0:
```
cbalioglu, line 752: If we make this function internal (as I mentioned above), we can ideally move argument checks to the public functions (`setup_parallel_gangs` and `to_fsdp` with more descriptive error messages. For instance, if the user calls `setup_parallel_gangs` with a wrong `tp_size`, the error message says "`row_length` must be less than or equal to root_gang.size" which is totally clear now.

src/fairseq2/nn/fsdp.py
```diff
@@ -89,10 +90,19 @@ def to_fsdp(
...
+        if local_world_size < 1:
             raise ValueError(
-                f"`local_world_size` must be greater than 0, but is {local_world_size} instead."
+                f"`local_world_size` must be greater than 1, but is {local_world_size} instead."
+            )
+
+        if local_world_size == 1:
+            raise ValueError(
+                f"`local_world_size` must be greater than 1, but is {local_world_size} instead. "
+                "This hybrid configuration would force FSDP to switch to use `NO_SHARD`, "
```
cbalioglu, line 104: Nit. As a convention, we do not line-break individual sentences of an error and log message within the code. This makes it possible for users to "grep" any part of the sentence (e.g. if they see it in the log file and want to figure out where it originates in the code base). This was mostly inspired by Linux kernel coding conventions.

tests/unit/test_gang.py
```diff
@@ -0,0 +1,131 @@
...
+            # mesh for 2 hosts, 2 GPUs each:
+            # Host 0: g0 | g1
+            # Host 1: g2 | g3
+            (0, 4, 2, ([0, 1], [0, 2])),
+            (1, 4, 2, ([0, 1], [1, 3])),
+            (2, 4, 2, ([2, 3], [0, 2])),
+            (0, 8, 4, ([0, 1, 2, 3], [0, 4])),
+        ],
+    )
+    def test_setup_2D_mesh_works(
```
cbalioglu, line 94: Nice! Thanks for writing tests! :)

cbalioglu (commented):
Looks quite good. I just have one comment regarding the visibility of `setup_2d_mesh_gangs`.

src/fairseq2/gang.py
MartinGleize: I implemented this but added an extra "setup_" function in `gang.py` to keep it coherent with `setup_parallel_gangs()`.

src/fairseq2/gang.py
MartinGleize: Good point, less defensive programming in the internal function now.

src/fairseq2/nn/fsdp.py
MartinGleize: Ah I actually was wondering about that, thank you for the explanation, implemented this.

PR #920, opened 2024-12-16
llama3.3 instruct 70b asset added

uralik:

added the llama3.3 instruct 70b asset. the model uses the same arch as 3.1 following HF: https://huggingface.co/meta-llama/Llama-3.3-70B-Instruct

so we follow that and use 3.1 arch as is


cbalioglu (approved):
LGTM!

PR #922, opened 2024-12-17
introduce final_lr_scale that is used by default

uralik:

We introduce final_lr_scale in cosine lr scheduler config so that we don't need to set final_lr manually (while its still an option). 
Also adding a WARNING log when the final lr is larger than the lr defined by the optimizer. For simplicity I took the first optimizer group lr in this case as the optimizer's lr.


src/fairseq2/optim/lr_scheduler/factory.py
```diff
@@ -85,14 +91,42 @@ def create_cosine_annealing_lr(
...
+            config.final_lr_scale is None
+        ), f"final_lr_scale must be None when final_lr is used, but config.final_lr_scale={config.final_lr_scale}"
+
+    if (config.final_lr is None and config.final_lr_scale is None) or (
+        config.final_lr is not None and config.final_lr_scale is not None
+    ):
+        # both are none, this is erroneous setup
+        raise ValueError(
+            "Either final_lr or final_lr_scale must be specified, but not both at the same time."
+        )
```
zyaoj, line 115: [Optional] Do we want to refactor this chunk a little bit to make it easier to read and remove duplicated logic, by:
- move the validation to the front to check errors earlier
- combine the `final_lr_scale` and `final_lr` to an if else since we have checked the errorneous setups already

The code snippet will look like:

```py
# Validate config and set final_lr
if (config.final_lr is not None) and (config.final_lr_scale is not None):
    raise ValueError(
        f"Invalid configuration: Both `final_lr` ({config.final_lr}) and `final_lr_scale` "
        f"({config.final_lr_scale}) are set. Please specify only one."
    )

if config.final_lr is None and config.final_lr_scale is None:
    raise ValueError(
        "Invalid configuration: Either `final_lr` or `final_lr_scale` must be specified."
    )

# Compute final_lr based on the configuration
if config.final_lr_scale is not None:
    final_lr = optimizer.param_groups[0]["lr"] * config.final_lr_scale
else:
    final_lr = config.final_lr

# Warn if final_lr exceeds the optimizer's current learning rate
if final_lr > optimizer.param_groups[0]["lr"]:
    log.warning(
        f"ATTENTION: Optimizer LR ({optimizer.param_groups[0]["lr"]}) < Final LR scheduler value ({final_lr}). "
        "This means your learning rate will increase over the course of training."
    )
```

zyaoj (approved):
LGTM

src/fairseq2/optim/lr_scheduler/factory.py
```diff
@@ -85,14 +91,42 @@ def create_cosine_annealing_lr(
     else:
         cycle_len = config.cycle_len
 
+    final_lr: float = 0.0  # using dummy value to make mypy happy
+
+    # set final lr based on config
+    if config.final_lr_scale is not None:
+        final_lr = optimizer.param_groups[0]["lr"] * config.final_lr_scale
+        assert (
```
cbalioglu, line 99: Do you mind converting these `assert` statements to `ValueError`s? So instead this, writing:
```
if config.final_lr is not None:
    raise ValueError("....")
```

zyaoj (approved):
LGTM!

PR #926, opened 2024-12-19
Expose weight initialization bounds for LayerNorm, Projection, Conv2D, Conv3D

antoine-tran:

Most of fairseq2 Modules runs the standard (Xavier) initialization function, or uniform / constant weights initialization. 

Sometimes users want to experiment with different algorithms too, for example when they want to manually set the boundaries of the weights. This was the case for the JEPA model.

This PR add parameters `init_fn` to the common Module (Projection, TransfomerEncoderLayer, LayerNorm)

Fixes #{issue number}

**Check list:**

### DIFF
--- a/src/fairseq2/models/jepa/factory.py
+++ b/src/fairseq2/models/jepa/factory.py
@@ -11,14 +11,8 @@ from dataclasses import dataclass, field
 from functools import partial
 from typing import Final, cast
 
-<<<<<<< HEAD
-import torch
-from torch.nn import GELU
-=======
-from torch import Tensor
 import torch
 from torch.nn import GELU, Module
->>>>>>> a09ad4fe (Can's comments)
 
 from fairseq2.config_registry import ConfigRegistry
 from fairseq2.models.jepa.model import JepaModel
--- a/src/fairseq2/models/jepa/factory.py
+++ b/src/fairseq2/models/jepa/factory.py
@@ -6,12 +6,12 @@
 
 from __future__ import annotations
 
-from collections.abc import Callable
+import math
 from dataclasses import dataclass, field
 from functools import partial
-import math
 from typing import Final, cast
 
+import torch
 from torch.nn import GELU
 
 from fairseq2.config_registry import ConfigRegistry
@@ -44,7 +44,9 @@ from fairseq2.nn.transformer import (
     create_default_sdpa,
 )
 from fairseq2.nn.transformer.residual import DropPathResidualConnect
-from fairseq2.nn.utils.module import init_truncated_uniforma_weights_and_bias as init_module
+from fairseq2.nn.utils.module import (
+    init_truncated_uniforma_weights_and_bias as init_module,
+)
 from fairseq2.typing import DataType, Device
 
 JEPA_FAMILY: Final = "jepa"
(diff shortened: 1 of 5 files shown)


src/fairseq2/models/jepa/factory.py
```diff
@@ -265,60 +291,108 @@ def build_encoder(self) -> TransformerEncoder:
...
+    def build_mha_output_projection(self, layer_idx: int) -> Linear:
         config = self._config
 
+        init_std = config.init_std
+
+        def init_projection(proj: Linear) -> None:
+            init_module(proj, std=init_std)
+
+            with torch.no_grad():
+                proj.weight.div_(math.sqrt(2.0 * layer_idx))
```
cbalioglu, line 341: Is this accurate? In the reference implementation, I see that the scaling is done with `layer_idx + 1` instead of `layer_idx`. https://github.com/facebookresearch/jepa/blob/main/src/models/vision_transformer.py#L150

src/fairseq2/nn/utils/module.py
```diff
@@ -570,3 +571,44 @@ def get_module_size(module: Module) -> ModuleSizeInfo:
         info.total_size_bytes += size_bytes
 
     return info
+
+
+def normalize_truncate(
```
cbalioglu, line 576: I think we can use PyTorch's [trunc_normal_](https://pytorch.org/docs/stable/nn.init.html#torch.nn.init.trunc_normal_) instead of this function. This is also noted in the reference implementation here: https://github.com/facebookresearch/jepa/blob/main/src/utils/tensors.py#L18-L19

src/fairseq2/nn/utils/module.py
```diff
@@ -570,3 +571,44 @@ def get_module_size(module: Module) -> ModuleSizeInfo:
...
+    tensor.uniform_(2 * lower - 1, 2 * upper - 1)
+    tensor.erfinv_()
+
+    tensor.mul_(std * math.sqrt(2.0))
+    tensor.add_(mean)
+
+    tensor.clamp_(min=a, max=b)
+
+
+def init_truncated_uniforma_weights_and_bias(
```
cbalioglu, line 600: I would prefer this function to be within JEPA's factory.py instead of this file which is meant typically for much more generic module helper functions.

src/fairseq2/models/jepa/factory.py
```diff
@@ -265,60 +291,108 @@ def build_encoder(self) -> TransformerEncoder:
...
+    def build_ffn(self, layer_idx: int) -> FeedForwardNetwork:
+        config = self._config
+
+        init_std = config.init_std
+
+        def init_projection(proj: Linear) -> None:
+            init_module(proj, std=init_std)
+
+            with torch.no_grad():
+                proj.weight.div_(math.sqrt(2.0 * layer_idx))
```
cbalioglu, line 361: Note that I performed this change to make sure that secondary `reset_parameters()` calls result in identical weight initialization.

src/fairseq2/models/jepa/factory.py
antoine-tran: good catch , yes this was the mistake, thanks @cbalioglu

src/fairseq2/models/jepa/factory.py
```diff
@@ -373,6 +370,13 @@ def init_projection(proj: Linear) -> None:
             dtype=self._dtype,
         )
 
+        # rescale the last layer
```
cbalioglu, line 373: Remnant from old commits?

cbalioglu (approved):

PR #936, opened 2024-12-23
Add script to write Llama's HF-formatted config.json for vLLM

MartinGleize:
Add script to write HF-format `config.json` for Llama. This `config.json` is then used to load fairseq2 Llama checkpoints directly in vLLM.

No

**Check list:**

### DIFF
--- a/src/fairseq2/models/llama/factory.py
+++ b/src/fairseq2/models/llama/factory.py
@@ -108,7 +108,7 @@ class RopeScaling:
     low_freq_factor: float = 1.0
     """Factor used to define low frequencies."""
 
-    high_freq_factor: float = 5.0
+    high_freq_factor: float = 4.0
     """Factor used to define high frequencies."""
 
     original_context_length: int = 8192
@@ -243,10 +243,7 @@ class LLaMABuilder:
             if self._config.rope_scaling is not None:
                 freqs_init_fn = functools.partial(
                     self._init_scaled_freqs,
-                    scale_factor=self._config.rope_scaling.factor,
-                    l_freq_factor=self._config.rope_scaling.low_freq_factor,
-                    h_freq_factor=self._config.rope_scaling.high_freq_factor,
-                    old_context_len=self._config.rope_scaling.original_context_length,
+                    rope_scaling=self._config.rope_scaling,
                 )
             else:
                 freqs_init_fn = None
@@ -295,13 +292,13 @@ class LLaMABuilder:
 
     @staticmethod
     def _init_scaled_freqs(
-        pos_encoder: RotaryEncoder,
-        scale_factor: float,
-        l_freq_factor: float,
-        h_freq_factor: float,
-        old_context_len: int,
+        pos_encoder: RotaryEncoder, rope_scaling: RopeScaling
     ) -> Tensor:
         device = pos_encoder.freqs.device
+        scale_factor = rope_scaling.factor
+        l_freq_factor = rope_scaling.low_freq_factor
+        h_freq_factor = rope_scaling.high_freq_factor
+        old_context_len = rope_scaling.original_context_length
 
         # (E / 2)
         indices = torch.arange(
--- a/src/fairseq2/models/llama/factory.py
+++ b/src/fairseq2/models/llama/factory.py
@@ -6,9 +6,10 @@
 
 from __future__ import annotations
 
-import math
 from dataclasses import dataclass, field
-from typing import Final
+import functools
+import math
+from typing import final, Final
 
 import torch
 from torch import Tensor
@@ -85,11 +86,33 @@ class LLaMAConfig:
     rope_theta: float = 10_000.0
     """The coefficient of the long-term decay of the Rotary position encoder."""
 
-    use_scaled_rope: bool = False
-    """If ``True``, scales Rotary encoding frequencies to LLaMA 3.1 context length."""
+    rope_scaling: RopeScaling | None = None
+    """If specified, provides scaling parameters for RoPE frequencies,
+    aiming to increase the context length."""
 
     dropout_p: float = 0.1
     """The dropout probability on outputs of Transformer layers."""
+
+
+@final
+@dataclass
+class RopeScaling:
+    """Holds the configuration for RoPE (Rotary Position Embedding)
+    scaling in Llama 3 models.
+    """
+
+    factor: float = 8.0
+    """Ratio between the intended max context length and the model’s
+    original max context length."""
+
+    low_freq_factor: float = 1.0
+    """Factor used to define low frequencies."""
+
+    high_freq_factor: float = 5.0
+    """Factor used to define high frequencies."""
+
+    original_context_length: int = 8192
+    """Original context length. Defaults to LLaMA 3's context length."""
 
 
 llama_archs = ConfigRegistry[LLaMAConfig]()
@@ -217,8 +240,14 @@ class LLaMABuilder:
         sdpa = create_default_sdpa(attn_dropout_p=self._config.dropout_p)
 
         if self._pos_encoder is None:
-            if self._config.use_scaled_rope:
-                freqs_init_fn = self._init_scaled_freqs
+            if self._config.rope_scaling is not None:
+                freqs_init_fn = functools.partial(
+                    self._init_scaled_freqs,
+                    scale_factor=self._config.rope_scaling.factor,
+                    l_freq_factor=self._config.rope_scaling.low_freq_factor,
+                    h_freq_factor=self._config.rope_scaling.high_freq_factor,
+                    old_context_len=self._config.rope_scaling.original_context_length,
+                )
             else:
                 freqs_init_fn = None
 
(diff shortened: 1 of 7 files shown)


src/fairseq2/models/llama/integ.py
```diff
@@ -38,3 +53,55 @@ def convert_to_reference_checkpoint(checkpoint: dict[str, Any]) -> dict[str, Any
...
+    else:
+        # mgleize: not sure of the json.dump behavior if rope_scaling is None
+        rope_scaling = None
+
+    # we only specify the parameters made explicit in the Huggingface converter
+    # https://github.com/huggingface/transformers/blob/93aafdc620d39b9ec714ffecf015a085ea221282/src/transformers/models/llama/convert_llama_weights_to_hf.py#L384
+    return {
+        "architectures": ["Fairseq2LlamaForCausalLM"],
+        "bos_token_id": config.vocab_info.bos_idx,
+        "eos_token_id": config.vocab_info.eos_idx,
```
cbalioglu, line 90: Have you seen any mention of EOT (end-of-turn) token in HG config.json? LLaMA 3 is a bit peculiar and mainly uses EOT for instruction models instead of EOS to separate prompts and responses. Not sure if that is relevant for config.json

src/fairseq2/models/llama/integ.py
```diff
@@ -38,3 +53,55 @@ def convert_to_reference_checkpoint(checkpoint: dict[str, Any]) -> dict[str, Any
...
+        n: int, ffn_dim_multiplier: float = 1, multiple_of: int = 256
+    ) -> int:
+        """From: https://github.com/huggingface/transformers/blob/82fcac0a7e40dc6cc5e3121d714b9b16775293ad/src/transformers/models/llama/convert_llama_weights_to_hf.py#L171"""
+        return multiple_of * (
+            (int(ffn_dim_multiplier * int(8 * n / 3)) + multiple_of - 1) // multiple_of
+        )
+
+    def is_llama_3_2(arch: str) -> bool:
+        # TODO: this seems too britle
+        return "llama3_2_" in arch
```
cbalioglu, line 71: Yeah, agree that this looks a bit hacky, but I think it is fine for now.

src/fairseq2/recipes/llama/__init__.py
```diff
@@ -18,3 +19,9 @@ def _setup_llama_cli(cli: Cli) -> None:
         handler=ConvertCheckpointCommandHandler(),
         help="convert fairseq2 LLaMA checkpoints to reference checkpoints",
     )
+
+    group.add_command(
```
cbalioglu, line 23: I think this is fine, but would be great if we could also have a way to automatically dump this config.json in LLM finetuning recipes.

cbalioglu (approved):
LGTM, just left two nit comments.

src/fairseq2/models/llama/integ.py
MartinGleize: Thank you for pointing this out to me. After checking there is definitely mention of an instruct-specific eos_token_id behavior in the Huggingface llama conversion script, and it definitely impacts the config.json:

```
if instruct:
    eos_token_id = [128001, 128008, 128009]
else:
    eos_token_id = 128001
```


`instruct` is a command-line flag argument of this script, is there an equivalent in fairseq2? I saw we configure special token ids in VocabularyInfo so the user could set eos_idx to the EOT token?

cbalioglu: Ok, my bad. Just realized that you get `model_config` based on the model name. Then this should be good to go. For instruct models, we use their model card to make sure that they set the right EOS. You can check it out [here](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/assets/cards/models/llama.yaml#L76-L80).

cbalioglu (approved):
At least the content of the PR looks good to me. Assuming that the generated config.json works with vLLM, I think it is good to g.

src/fairseq2/models/llama/integ.py
MartinGleize: To preface this, this PR yields exact parity with the current conversion process from our checkpoints to vLLM-compatible HF checkpoints (so: same config.json produced, and I've checked for generation equality on a dataset of prompts with vLLM).


But I think me needing the hack made me find a small inconsistency present on main right now:

RoPE scaling params in the conversion script depend on the Llama version (32.0 factor if Llama version 3.2 for example) but fairseq2 hardcodes these differently (factor is always 8.0):

https://github.com/facebookresearch/fairseq2/blob/03511c2d08fcae5946638a900138d3bc5070370b/src/fairseq2/models/llama/factory.py#L283C9-L286C26

So if we follow the intended workflow (either the current double conversion or simply writing the config.json), we end up with different RoPE parameters in vLLM inference than in training.

src/fairseq2/recipes/llama/__init__.py
MartinGleize: Agreed!

uralik:
looking great! @MartinGleize We might need to test this when we have experiment that started from some other experiment checkpoint (using resume from ckpt dir argument). In this case the model name might be just "checkpoint_step_N", and smth could be not retrieved as planned into the final config. We have a bunch of such examples on our side, can help with testing in this branch if needed.

cbalioglu:
> looking great! @MartinGleize We might need to test this when we have experiment that started from some other experiment checkpoint (using resume from ckpt dir argument). In this case the model name might be just "checkpoint_step_N", and smth could be not retrieved as planned into the final config. We have a bunch of such examples on our side, can help with testing in this branch if needed.

Let me merge AssetCard.flatten() tomorrow morning. It has been ready for a long time. I was aiming to merge it with recipe refactoring, but can do it separately for this PR.

src/fairseq2/models/llama/integ.py
cbalioglu: Interesting. When I check the reference implementation, I see that the scale factor is constant for all 3.x (https://github.com/meta-llama/llama-models/blob/main/models/llama3/reference_impl/model.py#L47). Do you know where that scale factor originates from?

cbalioglu:
Ok, looks like a bug in the reference implementation: https://github.com/pytorch/executorch/issues/7265

src/fairseq2/models/llama/integ.py
uralik: good catch @MartinGleize ! indeed 3.2 has scale factor 32 [2] , but 3.3 has scale factor 8 again! [1]

as Can pointed out below, llama-recipes ref impl also has factor 8 for all models there, but we need to fix this on the arch side for sure! 
@cbalioglu , shall we move the rope scaling to be part of the llama config? 

[1] https://huggingface.co/meta-llama/Llama-3.3-70B-Instruct/blob/main/config.json#L26-L32
[2] https://huggingface.co/meta-llama/Llama-3.2-3B-Instruct/blob/main/config.json#L26-L32

cbalioglu: @uralik yeah considering different archs have different scaling, we have to move it to the config.

uralik:
> Ok, looks like a bug in the reference implementation: [pytorch/executorch#7265](https://github.com/pytorch/executorch/issues/7265)

also discussed here (and not resolved looks like): https://github.com/meta-llama/llama-models/issues/241

cbalioglu:
[`flatten()`](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/assets/card.py#L128). fyi.

```
card = card.flatten()
```

will produce an asset card with all base metadata merged into one.

cbalioglu:
@MartinGleize Let me know if you need a hand with making RoPE scaling factor configurable. I believe it is a good nit exercise to work on, but happy to help if you need bandwidth for other parts of the project.

MartinGleize:
> @MartinGleize Let me know if you need a hand with making RoPE scaling factor configurable. I believe it is a good nit exercise to work on, but happy to help if you need bandwidth for other parts of the project.

Should be no problem for me!

src/fairseq2/models/llama/integ.py
MartinGleize: @cbalioglu @uralik Just added the configurable scaling.

Also, we currently use high_freq_factor of 5, while Huggingface and the ref impl use 4. Was it intended? I kept it at 5 for now.

src/fairseq2/models/llama/factory.py
```diff
@@ -217,8 +240,14 @@ def build_attention(
         sdpa = create_default_sdpa(attn_dropout_p=self._config.dropout_p)
 
         if self._pos_encoder is None:
-            if self._config.use_scaled_rope:
-                freqs_init_fn = self._init_scaled_freqs
+            if self._config.rope_scaling is not None:
+                freqs_init_fn = functools.partial(
```
cbalioglu, line 244: Nit: how about `_init_scaled_freqs(pos_encoder: RotaryEncoder, scaling: RopeScaling)` to make the signature shorter?

src/fairseq2/models/llama/integ.py
cbalioglu: There might have been an update on that as well. I can't find it right now, but we pretty much copy pasted the scaling code from the original reference implementation. I suggest changing that to 4 as it is apparently the right value.

cbalioglu (approved):
Nice, looks good to me.

MartinGleize:
I'm still wondering about tie_word_embeddings=true in the config.json (was trying to get rid of the last use of "is_llama_3_2"). Our impl and the ref impl don't have any mention of this or code (as far as I can see), but on HG they indeed have this setting to true in their config. Am I missing something or should we hard-code it to false on our side? That seems pretty important.

cbalioglu:
We definitely do not use tied weights for LLaMA. I am not sure how HG came up with that setting as I don't see any weight tying in the reference implementation either. I believe we have to set it to `False`, since our checkpoints will have different weights for the embeddings and final projection layer.

PR #958, opened 2025-01-08
Training DPO with pre-computed reference scores

uralik:

For large model finetuning it is beneficial to be able to precompute reference scores offline and avoid using 2nd LLM online to extract the same scores for training data every epoch.

We implement this by allowing to set DPO reference model to null. If this is the case, then finetune unit will try to use scores from the data batch. If the data batch scores are not available, training will crash. We could add some error handling here. 

TODOs:
* Error handling when reference model is not set, and scores do not exist in the batch.
* Improve the docstring around reference model in DPO config to highlight this feature.

## How scores look like in the data

we add 2 columns to json that contain floats:
* reference_score_rejected
* reference_score_chosen

## How to test

Example config:
```yaml
dataset: dpo_offline_score
optimizer_config:
  lr: 1e-7
max_num_data_epochs: 100
max_num_steps: 500
max_num_tokens: 8192
max_seq_len: 8192
publish_metrics_every_n_steps: 1
keep_last_n_models: 20
checkpoint_every_n_steps: 50
batch_size: 1
model: llama3_3_70b_instruct
tensor_parallel_size: 8
criterion: dpo
_add_:
  criterion_config:
      reference_model: null  # to use offline scores
      beta: 0.1
      nll_scale: 0.0
wandb_project: dpo_offline_score
wandb_run_name: weizhe_run
```

Example run command on h100-2:

`FAIRSEQ2_USER_ASSET_DIR=/opt/hpcaas/.mounts/fs-0df31b178aa4037ac/home/kulikov/code/RAM/fairseq2_ram/src/fairseq2_ram/cards srun fairseq2 lm preference_finetune $(realpath .)/run --no-sweep-dir --config-file ./train_config.yaml`

Note there is `FAIRSEQ2_USER_ASSET_DIR` variable pointing to the dataset card

### DIFF
--- a/src/fairseq2/datasets/preference.py
+++ b/src/fairseq2/datasets/preference.py
@@ -48,6 +48,8 @@ class PreferenceOptimizationBatch:
 
     chosen: SequenceBatch
     rejected: SequenceBatch
+    reference_score_chosen: torch.Tensor | None
+    reference_score_rejected: torch.Tensor | None
 
 
 class PreferenceOptimizationDataset(ABC):
@@ -232,6 +234,10 @@ class GenericPreferenceOptimizationDataset(PreferenceOptimizationDataset):
                 "indices_prompt": source_indices,
                 "indices_chosen": indices_chosen,
                 "indices_rejected": indices_rejected,
+                "reference_score_chosen": example.get("reference_score_chosen", None),
+                "reference_score_rejected": example.get(
+                    "reference_score_rejected", None
+                ),
                 "target_mask_chosen": target_mask_chosen,
                 "target_mask_rejected": target_mask_rejected,
                 "total_tokens": total_tokens,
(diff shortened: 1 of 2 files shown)


PingYu-iris (approved):
LGTM!

jacklanchantin:
Is there a reason for calling them `reference_score_chosen` instead of `reference_logp_chosen`? Other than that, looks good!

jacklanchantin:
> > Is there a reason for calling them `reference_score_chosen` instead of `reference_logp_chosen`? Other than that, looks good!
> 
> I thought it would be good if we can use this in any other loss that can take whatever reference scores i.e. some reward model score as long as it its passed in the dataset maybe I was overthinking that though!

When I first read `reference_score_chosen`, I assumed it was for a reward score, that's why I was asking. If it can be applied to both cases, then I think it makes sense.

xianxl:
Would it be useful to add a reference script that computes the scores?

uralik:
> Would it be useful to add a reference script that computes the scores?

Not in fairseq2 as it will be too specific given our setups. 
Ping and Weizhe will prepare reference scripts given our projects needs (in RAM repo)!

cbalioglu (approved):

PR #980, opened 2025-01-20
Non-deterministic map operator option in data loading

MartinGleize:

Add the "deterministic" boolean parameter to the map operator in data_pipeline. It follows similar semantics as the tensorflow equivalent: https://www.tensorflow.org/api_docs/python/tf/data/Dataset#map , i.e. it trades the matching of output and input order with some execution speed.

No

**Check list:**


MartinGleize:
Tested on toy example:

```
num_parallel_calls = 10

def fn(d: int) -> int:
    time.sleep(d)
    return d

pipeline = (
    read_sequence(list(range(num_parallel_calls, 0, -1)))
        .map(fn, num_parallel_calls=num_parallel_calls, deterministic=False)
        .and_return()
)
```

"=True" returns 10, 9, 8, ... in a quick burst after a 10s wait, while "=False" returns 1, 2, 3, ..., one every second.

native/src/fairseq2n/detail/thread_pool.h
```diff
@@ -0,0 +1,97 @@
...
+                        queued_condition_.wait(lock, [this] {
+                            return stop_ || !tasks_.empty();
+                        });
+                        
+                        if (stop_ && tasks_.empty()) {
+                            return;
+                        }
+
+                        task = std::move(tasks_.front());
+                        tasks_.pop();
```
artemru, line 41: do we track somehow the submission time here ? (to make sure that among finished tasks we return tasks according to their submission order).

native/src/fairseq2n/data/map_data_source.cc
```diff
@@ -128,6 +187,74 @@ map_data_source::fill_buffer()
...
+    {
+        return tasks_in_flight_ == 0 || exception_ptr_;
+    });
+
+    if (exception_ptr_)
+        std::rethrow_exception(exception_ptr_);
+}
+
+bool
+map_data_source::fill_buffer_async()
```
artemru, line 219: I'm right to say that the max buffer size will be `num_parallel_calls` in any case ?

tests/unit/data/data_pipeline/test_map.py
```diff
@@ -337,3 +345,60 @@ def fn(d: int) -> int:
 
         with pytest.raises(StopIteration):
             next(iter(pipeline))
+
+    @pytest.mark.parametrize("num_parallel_calls", [1, 4, 20])
+    def test_op_saves_and_restores_its_state_non_deterministic(self, num_parallel_calls: int) -> None:  # fmt: skip
+
+        def fn(d: int) -> int:
+            return d
```
artemru, line 353: it would be important to add a test where different tasks have different sleep(time) and to check the return order

native/src/fairseq2n/data/map_data_source.cc
MartinGleize: Yes. This was intended, I wondered about providing the option to have it higher, would that be interesting? Maybe for slightly better throughput.

native/src/fairseq2n/detail/thread_pool.h
MartinGleize: We do not, that's a good point (they will come out sorted by completion order). Is the option of preserving submission order as much as feasible still desirable in the deterministic=False setting?

PR #987, opened 2025-01-24
Use logprob scores in sampling generator

cbalioglu:
A small PR that changes the step scores produced by the sampling sequence generator to logprobs as it is more conventional than regular probs.


uralik (approved):
looks good!

PR #1007, opened 2025-02-10
Ensure that the model is in train mode

cbalioglu:
This PR ensures that the model is in train mode before starting the training. The expectation is to always pass the model with `train()` set before, but we do this to avoid any surprises in case the caller forgets the call.


pa-dqn (approved):
Perfect, thanks!

pa-dqn (approved):

PR #1009, opened 2025-02-11
extras example with keep_jsonl_keys option

uralik:

We have extras field for data reader that users can use in their extension modules. One common use case is to allow some jsonl columns to be passed in the example in addition to those that are expected by preference recipe. 

Now we show as a reference code how to utilize extras so that we can specify list of keys that will be kept as part of the example in the batch object.


src/fairseq2/datasets/preference.py
```diff
@@ -216,6 +216,16 @@ def cat_source_and_target(example: dict[str, Any]) -> dict[str, Any]:
                 + len(target_indices_rejected)
             )
 
+            # below is an example of using extras field of data reader options
+            if "keep_jsonl_keys" in options.extras:
+                jsonl_keys = options.extras["keep_jsonl_keys"]
+                assert isinstance(jsonl_keys, list) and all(
```
cbalioglu, line 222: Nit: can you please change this assertion to a `raise ValueError()`. Otherwise, looks good to me!

cbalioglu (approved):

PR #1072, opened 2025-02-27
Instantiate metric recorders on rank 0 only

cbalioglu:
This PR ensures that `MetricRecorder` instances are constructed only on rank 0. Some recorders such as `WandbRecorder` has the side effect of setting up their environment during construction which causes redundant/duplicate artifacts to be generated. The actual metric records are not affected though since we always write them on rank 0.


zyaoj (approved):
LGTM!

PR #1073, opened 2025-02-27
updated end_to_end_fine_tuning.rst

eduardosanchezg:
adding --gpus-per-node=8 argument to srun multi-node example


cbalioglu (approved):
Thanks for the fix!

PR #1096, opened 2025-03-28
Add RayClusterHandler to enable fairseq2 run on Ray

dongwang218:
Allow fairseq2 to run inside ray cluster.

**Check list:**
- [N ] Was the content of this PR **discussed and approved** via a GitHub issue? (no need for typos or documentation improvements)
- [ Y] Did you read the [contributor guideline](https://github.com/facebookresearch/fairseq2/blob/main/CONTRIBUTING.md)?
- [ Y] Did you make sure that your **PR does only one thing** instead of bundling different changes together?
- [ Y] Did you make sure to **update the documentation** with your changes? (if necessary)
- [ N] Did you write any **new necessary tests**?
- [ Y] Did you verify new and **existing tests pass** locally with your changes?
- [ N] Did you **update the [CHANGELOG](https://github.com/facebookresearch/fairseq2/blob/main/CHANGELOG.md)**? (no need for typos, documentation, or minor internal changes)

### DIFF
--- a/src/fairseq2/cluster.py
+++ b/src/fairseq2/cluster.py
@@ -6,14 +6,18 @@
 
 from __future__ import annotations
 
+import socket
 import subprocess
+import time
 from abc import ABC, abstractmethod
 from collections.abc import Collection, Mapping, MutableMapping
 from random import Random
 from typing import final
 
+import ray
 from typing_extensions import override
 
+from fairseq2.logging import log
 from fairseq2.registry import Provider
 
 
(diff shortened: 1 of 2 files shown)


src/fairseq2/cluster.py
```diff
@@ -182,3 +186,163 @@ def supports_current_cluster(self) -> bool:
...
+        self.placement_group_ids = placement_group_ids
+
+    def register_worker(self, worker_id, hostname, node_id, placement_group_id, gpu_id):
+        """Register a worker with its placement group ID and GPU ID"""
+        assert (
+            placement_group_id in self.placement_group_ids
+        ), f"Placement group id {placement_group_id} should in {self.placement_group_ids}"
+        assert 0 <= gpu_id <= 7, f"Bad gpu_id {gpu_id}"
+        node_index = self.placement_group_ids.index(placement_group_id)
+        global_rank = node_index * self.gpus_per_node + gpu_id
```
artemru, line 215: would it be working if ray cluster does not take a full node ?

src/fairseq2/cluster.py
```diff
@@ -182,3 +186,163 @@ def supports_current_cluster(self) -> bool:
...
+        assert coordinator_name
+        self.coordinator = ray.get_actor(*coordinator_name.split(":"))
+
+        accelerator_ids = ray_context.get_accelerator_ids()
+        if (
+            not accelerator_ids
+            or "GPU" not in accelerator_ids
+            or not accelerator_ids["GPU"]
+        ):
+            raise ValueError("No GPU accelerator ID available")
```
artemru, line 285: can we just skip in that case ? to avoid ray cluster with mixed type of deployment ?

artemru:
nice contribution !
can you add a simple integration test and some doc so everyone could start using this integration with Ray ?

src/fairseq2/cluster.py
dongwang218: this version is not going to work as it assumes gpu 0 is the leader. Just updated the code to passing in the rank, where rank==0 is the leader. So this will enable we asking for not all the gpus of a node.

src/fairseq2/cluster.py
dongwang218: no longer need to detect whether gpu == 0.

src/fairseq2/cluster.py
```diff
@@ -182,3 +186,163 @@ def supports_current_cluster(self) -> bool:
     @override
     def supported_cluster(self) -> str:
         return "none"
+
+
+@ray.remote(num_cpus=0)
```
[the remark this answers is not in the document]
dongwang218: ray head sometimes are launch with num_cpu == 0 to avoid cpu jobs running on it. In this case, we want the coordinator to run in the head, which force us to set its required cpu to be 0 to enable the scheduling, other `ray.exceptions.ActorUnschedulableError: The actor is not schedulable`

setup.py
```diff
@@ -69,6 +69,7 @@
         # listed as optional in tiktoken's pyproject.toml
         # (https://github.com/openai/tiktoken/blob/main/pyproject.toml#L9)
         "blobfile~=3.0.0",
+        "ray~=2.40",
```
cbalioglu, line 72: I believe this should be optional. In fact, instead of making it optional, we can just do a quick import check in the cluster implementation, and raise an error advising the use to install Ray themselves. For majority of use cases, people won't be using Ray. See [this](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/nn/_normalization.py#L21-L34) as an example.

src/fairseq2/cluster.py
```diff
@@ -182,3 +187,108 @@ def supports_current_cluster(self) -> bool:
...
+    _env: MutableMapping[str, str]
+
+    def __init__(self, env: MutableMapping[str, str]) -> None:
+        self._env = env
+
+    @override
+    def set_torch_distributed_variables(self) -> None:
+        env = self._env
+
+        rank_str = env.get("RANK")
```
cbalioglu, line 242: You can (ideally should) use the `get_rank`, `get_local_rank`, `get_world_size`, and `get_local_world_size` utility functions defined here: https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/utils/env.py#L85

They also raise an appropriate exception that causes the process to terminate gracefully.

setup.py
dongwang218: made ray as extra. also make sure the test and existing code works without install ray.

src/fairseq2/cluster.py
dongwang218: thanks, changed to use utility functions.

artemru:
I would be interested in reanimating this work!

PR #1109, opened 2025-04-15
Update sweep tag and add doc

zyaoj:

Update the default sweep tags following the refactoring of our configs.

***Can's edit***
The `sweep_keys` and `extra_sweep_keys` were helpful in the v0.3 days when we did not have the `--sweep-format` option and the sweep directory format was hard-coded. Since our recipe configurations are significantly bigger and dynamic now, it is not possible to list every possible key (e.g. every model, optimizer, lr scheduler configuration). In this PR, we entirely remove the `sweep_keys` and `extra_sweep_keys`, and use the entire configuration when computing the sweep hash. It is now user's responsibility if they want to have a specific directory format via `--sweep-format`. 

N/A

**Check list:**


src/fairseq2/recipes/utils/sweep_tag.py
```diff
@@ -244,39 +244,127 @@ def get_sweep_keys(extra_sweep_keys: Set[Hashable] | None) -> Set[Hashable]:
...
+        "lr_mul",
+        "start_lr",
+        "final_lr",
+        "final_lr_scale",
+        # Regime configuration
+        "num_steps",
+        "num_data_epochs",
+        "score_metric",
+        "lower_score_better",
+        "validate_after_n_steps",
```
cbalioglu, line 327: No blocker since you are on leave. Just a note though that some of the newly added keys are not meant to be used in sweep names. For instance the `xxx_after_n_steps` and `xxx_every_n_steps` options do not affect the actual job, so they should not be really part of the output name. I will filter out them out and push to your PR before merging it.

cbalioglu (approved):

PR #1126, opened 2025-04-18
adding integration with fsspec FileSystem

artemru:

Fixes #{issue number}

**Check list:**

### DIFF
--- a/src/fairseq2/utils/file.py
+++ b/src/fairseq2/utils/file.py
@@ -17,7 +17,9 @@ from pickle import PickleError
 from shutil import copytree, rmtree
 from typing import BinaryIO, TextIO, TypeAlias, cast, final
 
+import fsspec
 import torch
+from fsspec.implementations.local import LocalFileSystem as fsspec_LocalFileSystem
 from torch import Tensor
 from typing_extensions import override
 
@@ -79,7 +81,7 @@ class FileSystem(ABC):
 
 
 @final
-class LocalFileSystem(FileSystem):
+class NativeLocalFileSystem(FileSystem):
     @override
     def is_file(self, path: Path) -> bool:
         return path.is_file()
@@ -168,6 +170,115 @@ class LocalFileSystem(FileSystem):
     @override
     def is_local(self) -> bool:
         return True
+
+
+class FSspecFileSystem(FileSystem):
+    """
+    Wrapper around fsspec to provide a FileSystem interface.
+    >>> from s3fs import S3FileSystem
+    >>> fs = S3FileSystem()
+    >>> fs = FSspecFileSystem(fs)
+    """
+
+    __fs: fsspec.AbstractFileSystem
+
+    def __init__(self, fs: fsspec.AbstractFileSystem) -> None:
+        self.__fs = fs
+
+    @override
+    def is_file(self, path: Path) -> bool:
+        return self.__fs.isfile(str(path))
+
+    @override
+    def is_dir(self, path: Path) -> bool:
+        return self.__fs.isdir(str(path))
+
+    @override
+    def exists(self, path: Path) -> bool:
+        return self.__fs.exists(str(path))
+
+    @override
+    def open(self, path: Path, mode: FileMode = FileMode.READ) -> BinaryIO:
+        mode_str = "rb"
+        if mode == FileMode.WRITE:
+            mode_str = "wb"
+        elif mode == FileMode.APPEND:
+            mode_str = "ab"
+
+        return cast(BinaryIO, self.__fs.open(str(path), mode_str))
+
+    @override
+    def open_text(self, path: Path, mode: FileMode = FileMode.READ) -> TextIO:
+        mode_str = "r"
+        if mode == FileMode.WRITE:
+            mode_str = "w"
+        elif mode == FileMode.APPEND:
+            mode_str = "a"
+
+        return cast(TextIO, self.__fs.open(str(path), mode_str, encoding="utf-8"))
+
+    @override
+    def move(self, old_path: Path, new_path: Path) -> None:
+        self.__fs.mv(str(old_path), str(new_path), recursive=True)
+
+    @override
+    def remove(self, path: Path) -> None:
+        # only one file !
+        self.__fs.rm(str(path), recursive=False)
+
+    @override
+    def make_directory(self, path: Path) -> None:
+        self.__fs.makedirs(str(path), exist_ok=True)
+
+    @override
+    def copy_directory(self, source_path: Path, target_path: Path) -> None:
+        self.__fs.copy(str(source_path), str(target_path), recursive=True)
+
+    @override
+    def remove_directory(self, path: Path) -> None:
+        self.__fs.rmdir(str(path))
+
+    @override
+    def glob(self, path: Path, pattern: str) -> Iterable[Path]:
+        paths = self.__fs.glob(str(path.joinpath(pattern)))
+        return [Path(p) for p in paths]
+
+    @override
+    def walk_directory(
+        self, path: Path, *, on_error: Callable[[OSError], None] | None
+    ) -> Iterable[tuple[str, Sequence[str]]]:
+        results = self.__fs.walk(str(path), on_error=on_error)  # type: ignore[arg-type]
+        for dir_pathname, _, filenames in results:
+            yield str(dir_pathname), filenames  # type: ignore[return-value]
+
+    @override
+    def resolve(self, path: Path) -> Path:
+        # Use expanduser if available, otherwise just convert to string and back
+        if self.is_local:
+            return path.expanduser().resolve()
+        raise NotImplementedError("resolve is not implemented for this filesystem")
+
+    @property
+    @override
+    def is_local(self) -> bool:
+        try:
+            return isinstance(self.__fs, fsspec_LocalFileSystem)
+        except ImportError:
+            return False
+
+
+@final
+class LocalFileSystem(FSspecFileSystem):
+    def __init__(self) -> None:
+        super().__init__(fsspec_LocalFileSystem())
+
+
+@final
+class S3FileSystem(FSspecFileSystem):
+    def __init__(self, *arg, **kwargs) -> None:
+        from s3fs import S3FileSystem
+
+        super().__init__(S3FileSystem(*arg, **kwargs))
 
 
 MapLocation: TypeAlias = (


src/fairseq2/setup/_root.py
```diff
@@ -46,7 +46,7 @@
 def setup_library(progress_reporter: ProgressReporter | None = None) -> RuntimeContext:
     env = os.environ
 
-    file_system = LocalFileSystem()
+    file_system = GlobalFileSystem()
```
artemru, line 49: hmm, should we move that after `register_filesystems(...)` call ?

src/fairseq2/file_system.py
```diff
@@ -39,6 +46,11 @@ def is_dir(self, path: Path) -> bool: ...
     @abstractmethod
     def open(self, path: Path, mode: FileMode = FileMode.READ) -> BinaryIO: ...
 
+    @abstractmethod
+    def cat(self, path: Path) -> bytes:
```
artemru, line 50: it's useful for faster files downloading

src/fairseq2/assets/download_manager.py
```diff
@@ -243,7 +248,25 @@ def supported_schemes(self) -> Set[str]:
 
 @final
 class LocalAssetDownloadManager(AssetDownloadManager):
-    _SCHEMES: Final = {"file"}
+    """
+    Asset download manager for local files and fsspec-compatible URI schemes.
+
+    For local files (file:// scheme), it returns the local path directly.
+    For fsspec schemes (s3://, gs://, etc.), it returns the URI as a Path,
```
cbalioglu: Not really sure whether we should treat schemes beyond `file://` as "local" in a download manager. It defeats the purpose of download manager if we do not download and cache a remote file.

src/fairseq2/checkpoint/manager.py
```diff
@@ -485,21 +485,20 @@ def _commit_checkpoint(self, step_nr: int) -> None:
         except GangError as ex:
             raise_operational_gang_error(ex)
 
-    def _sync_nfs_cache(self) -> None:
-        if not self._file_system.is_local:
-            return
-
+    def _sync_distributed_writes(self) -> None:
```
cbalioglu: This whole function can become noop in case the file system is not local (I assume fsspec guarantees some form of strong consistency, if not, this function will likely become tricky). We do not need a redundant barrier call.

src/fairseq2/file_system.py
```diff
@@ -76,11 +88,426 @@ def resolve(self, path: Path) -> Path: ...
...
+    def is_local_path(self, path: Path) -> bool:  # noqa: ARG002
+        """Check if a specific path is on a local filesystem.
+
+        This method can be overridden by filesystems that handle both local
+        and remote paths. The default implementation returns self.is_local.
+        """
+        return self.is_local
+
+
+ExtenedPath: TypeAlias = str | Path | Sequence[str | Path]
```
cbalioglu: Why `Sequence`?

src/fairseq2/file_system.py
```diff
@@ -76,11 +88,426 @@ def resolve(self, path: Path) -> Path: ...
...
+            scheme = self.__prefix.split("://")[0]
+            # Check if path looks like a URI with single slash (normalized by pathlib)
+            if path_str.startswith(f"{scheme}:/") and not path_str.startswith(
+                f"{scheme}://"
+            ):
+                # Restore the double slash
+                return f"{scheme}://{path_str[len(scheme) + 2:]}"
+        return path_str
+
+    def get_short_uri(self, path: ExtenedPath) -> str | list[str]:
```
cbalioglu: I think naming this type alias `ExtendedPath` although it can also be a sequence of paths is highly misleading as a name.

src/fairseq2/file_system.py
```diff
@@ -76,11 +88,426 @@ def resolve(self, path: Path) -> Path: ...
...
+@final
+class FSspecFileSystem(FileSystem):
+    """
+    Wrapper around fsspec to provide a FileSystem interface.
+    >>> from s3fs import S3FileSystem
+    >>> fs = S3FileSystem()
+    >>> fs = FSspecFileSystem(fs)
+    """
+
+    __fsspec: fsspec.AbstractFileSystem
```
cbalioglu: Since this class is final, no need for double underscores.

src/fairseq2/file_system.py
```diff
@@ -76,11 +88,426 @@ def resolve(self, path: Path) -> Path: ...
...
+def _register_filesystems(context: Any) -> None:
+    # Protocols to skip - either problematic or not needed
+    # ftp: Has cleanup issues (__del__ AttributeError)
+    # sftp: Requires SSH connection
+    # ssh: Requires SSH connection
+    # smb: Requires SMB/CIFS connection
+    # hdfs: Requires Hadoop
+    # arrow_hdfs: Requires Hadoop
+    # webhdfs: Requires Hadoop
+    SKIP_PROTOCOLS = {
```
cbalioglu: I would much prefer to have a "allowed" protocols list instead of a "skip" list. Better safe than sorry.

src/fairseq2/assets/download_manager.py
artemru: what is the exact purpose of that class btw ? 
or it's a pure naming issue LocalAssetDownloadManager -> GlobalAssetDownloadManager ?

src/fairseq2/checkpoint/manager.py
artemru: just wanted to confirm that

artemru: yes, i think s3 should get some transactional consistency indeed

caciolai: https://aws.amazon.com/blogs/aws/amazon-s3-update-strong-read-after-write-consistency/

src/fairseq2/checkpoint/manager.py
```diff
@@ -387,7 +387,7 @@ def _do_save_checkpoint(
         saved_callback: CheckpointSavedCallback | None,
         blocking: bool,
     ) -> None:
-        self._sync_nfs_cache()
+        self._sync_distributed_writes()
```
caciolai: beside this change is anything else needed to make checkpointing work on any fsspec-compliant filesystem like s3fs?

src/fairseq2/file_system.py
caciolai: nit: typo `ExtendedPath`

src/fairseq2/file_system.py
caciolai: as far as I could tell, all the references to this functions seem to pass in just one single path

src/fairseq2/file_system.py
artemru: sounds good

src/fairseq2/checkpoint/manager.py
artemru: not to my knowledge, i can get the renaming back

src/fairseq2/file_system.py
artemru: yes, i was doing with mypy issues at some point with it, - can simplify it

src/fairseq2/file_system.py
artemru: ok, why not

src/fairseq2/checkpoint/__init__.py
```diff
@@ -44,3 +47,15 @@
 from fairseq2.checkpoint.model_metadata import (
     StandardModelMetadataLoader as StandardModelMetadataLoader,
 )
+
+
+@final
+class CheckpointDir:
```
artemru, line 53: what is the right place for this dubby holder class ? maybe we can even remove it

src/fairseq2/checkpoint/hg.py
```diff
@@ -121,18 +126,33 @@ def do_export() -> Callable[[], None]:
...
+
+                        err_fp = exit_stack.enter_context(
+                            self._file_system.open_text(err_file, mode=FileMode.WRITE)
+                        )
+
+                        result = self._process_runner.run_text(
+                            cmd, stdout=out_fp, stderr=err_fp
+                        )
+                else:
+                    result = self._process_runner.run_text(cmd, capture_output=True)
```
artemru, line 145: generally speaking, we cannot pass self._file_system into subprocess, i dont know that is the best option here ?

src/fairseq2/file_system.py
```diff
@@ -76,11 +100,454 @@ def resolve(self, path: Path) -> Path: ...
...
+                str(local_path),
+                normalized_uri,
+                recursive=local_path.is_dir(),
+            )
+
+    @override
+    def remove_directory(self, path: Path) -> None:
+        # For S3/remote filesystems, rmdir only removes empty directories
+        # Use rm with recursive=True to delete directory and all contents
+        self._fsspec.rm(self.get_short_uri(path), recursive=True)
```
artemru, line 294: there's an issue with phantoms s3 directories that cannot be removed...

src/fairseq2/recipe/internal/metric_recorders.py
```diff
@@ -167,19 +167,12 @@ def get_id(self) -> str:
 
         run_id_file = self._save_dir.joinpath("wandb_run_id")
 
-        fp = None
-
         try:
-            fp = self._file_system.open_text(run_id_file)
-
-            return fp.read()
+            return self._file_system.cat(run_id_file).decode("utf-8")
```
artemru, line 171: i still dont understand why we have filesystem abstraction here if we cannot really use other remote fs for the metrics logging...

src/fairseq2/recipe/internal/train_model.py
```diff
@@ -327,19 +327,25 @@ def save(self, family_name: str, config: object) -> None: ...
 @final
 class _StandardTrainModelMetadataSaver(_TrainModelMetadataSaver):
     def __init__(
-        self, metadata_dumper: ModelMetadataDumper, output_dir: Path, gangs: Gangs
+        self,
+        metadata_dumper: ModelMetadataDumper,
+        output_dir: Path,
+        gangs: Gangs,
+        checkpoint_dir: Path | None = None,
```
artemru, line 334: i dont know where else we need to propagate checkpoint_dir ...

src/fairseq2/checkpoint/hg.py
```diff
@@ -121,18 +126,33 @@ def do_export() -> Callable[[], None]:
...
-                    out_fp = exit_stack.enter_context(
-                        self._file_system.open_text(out_file, mode=FileMode.WRITE)
-                    )
-
-                    err_fp = exit_stack.enter_context(
-                        self._file_system.open_text(err_file, mode=FileMode.WRITE)
-                    )
-
-                    result = self._process_runner.run_text(
-                        cmd, stdout=out_fp, stderr=err_fp, env={}
```
artemru, line 134: do you know why env={} was necessary here ?

src/fairseq2/checkpoint/hg.py
artemru: i cannot pass env={} here become varenv can be useful to reinit s3 filesystem from aws related varenvs

cirquit:
> I believe one major thing missing right now is some benchmark numbers as well as verification that the new file system implementation works end-to-end (i.e. a real world training run, asset metadata loading, Hugging Face model exports, and any other place where we used regular file system calls work without any regressions). Have you been able to verify those?

Taking this on from @artemru for the final push.
* Added fsspec as explicit dependecy given that we use it in `file_system.py` (93be990d1)
* Added some tests for GlobalFileSystem (60338fee)
* `pathlib.Path` normalizes `s3://` to `s3:/` which caused `FileSystemRegistry` to fall through to local filesystem instead of using S3. Match both URI forms in pattern check (a6577da). This is ultimately a hotfix and not a good patch given that this mangling comes from the use of `pathlib.Path` which simply does not behave in line with the way we access S3 with `s3fs`. Should be reworked in its own PR.

--- 

### Small smoke test

A quick S3 checkpointing benchmark (Llama 3.1 8B, 8x H100, FSDP):
  - Add checkpoint: `"hg://meta-llama/Llama-3.1-8B"` and `tokenizer: llama3_1_8b` asset card `llama.yaml` (also something for the future to be reworked)
  - `pip install s3fs`, configure AWS profile in `~/.aws/config`

```bash
AWS_PROFILE=YOUR_PROFILE_RW torchrun --nproc-per-node=8 -m recipes.lm.sft "/tmp/fsspec_s3_test" \
    --checkpoint-dir s3://YOUR_BUCKET/path/ \
    --config-file recipes/lm/sft/configs/llama3_2_1b_instruct_gsm8k.yaml \
    --config model.name=llama3_1_8b --config tokenizer.name=llama3_1_8b \
    --config dataset.chat_mode=false --config dataset.max_seq_len=2048 --config dataset.max_num_tokens=2048 \
    --config trainer.data_parallelism=fsdp --config regime.num_steps=400 \
    --config regime.checkpoint_every_n_steps=200 --config regime.validate_every_n_steps=200
```

Disable HF export with `--config regime.save_model_only=true`. Resume by re-running the same command.

Actual data sizes (from S3):
  - Model shard: 4.0 GB × 8 ranks = **29.9 GB**
  - Optimizer shard: 8.0 GB × 8 ranks = **59.8 GB**
  - **Total checkpoint: ~90 GB/step**
  - HF export: 7 safetensors files = **32.1 GB**

  | Operation | Wall clock | Data | Throughput |
  |---|---|---|---|
  | Checkpoint save | ~2m| ~90 GB (8 ranks) | ~87 MB/s/rank |
  | Resume: model load | ~34s | ~30 GB (8 ranks) | ~110 MB/s/rank |
  | Resume: optimizer load | ~65s | ~60 GB (8 ranks) | ~115 MB/s/rank |
  | **Resume total** | **~1m 40s** | **~90 GB** | |
  | HF export | ~5-7m | ~32 GB | ~76 MB/s single stream (single-threaded export, bottleneck) |
  | **Save+export (2 steps)** | **16m 12s** | | |
  | **Save only (2 steps)** | **~4m 14s** | | `save_model_only=true` |

Per-rank S3 throughput is ~85-115 MB/s. Checkpoint saves are async during training. HF export is the bottleneck — single-stream from rank 0.

src/fairseq2/checkpoint/__init__.py
cirquit: The holder is needed because the DI container resolves by type - without it, `checkpoint_dir` and `output_dir` are both `Path` and can't be distinguished. Not that its super clean here, but its fine I think.

src/fairseq2/file_system.py
cirquit: Hm, `remove_directory` is called by `delete_checkpoint`  in `manager.py` when we need to delete old checkpoints. 

I've verified this to work when running a training where I have:
```bash
--config regime.checkpoint_every_n_steps=100 \
--config regime.keep_last_n_checkpoints=2 \
--config regime.num_steps=300 \
```

Deletion seems to work for me when I check the s3 bucket contents (only step 200 and 300 exist).  

The only issue that if the "checkpointing time" < "step time until the next checkpoint" (i.e. happens during frequent checkpointing) we wait before training further. This should not be a problem in real training, but can be a pain when seeing this by accident in smaller runs where checkpointing is enabled. Might be a good forcing function to only use S3 checkpointing if really useful/fix the checkpointing frequency. 

Might be worthwhile to allow parallel async checkpointing, but definitely not for this PR.

src/fairseq2/recipe/internal/metric_recorders.py
cirquit: The issue is that metric loggers (wandb, tensorboard, jsonl writers) write incrementally. They append lines, keep file handles open, etc. and S3 is an object store, not a filesystem so it does not support this.

src/fairseq2/recipe/internal/train_model.py
cirquit: Not clean, but we'll take it for now - this should be optimally refactored to be in `_StandardTrainModelMetadataSaver` so every child (`OutOfProcHuggingFaceExporter` and `StandardCheckpointManager`) gets this without the `| None = None` signature.

src/fairseq2/checkpoint/hg.py
cirquit: Yea, this was added in #1438 - exporting with WORLD_SIZE and other environment variables makes huggingface exporting think its part of a distributed run, so it behaves strangely. We must make it think its on its own.

src/fairseq2/checkpoint/hg.py
cirquit: As explained below - hg must have this parameter be empty to function correctly.

src/fairseq2/assets/download_manager.py
cirquit: I think the `LocalAssetDownloadManager` should essentially not do this if its just forwarding the S3 path to the `CheckpointManager`. Should be revisited later to see if we can separate the concerns a clean `FileSystem` implementation and an `AssetDownloadManager` that returns the path and whether a local copy exists (active caching ownership).

This makes file:// a no-op, for s3:// or hg:// it downloads on demand and so on. The concept of local will then be owned instead of implied by the class name.

PR #1155, opened 2025-04-30
Fix bug in delete_stale_checkpoints which was causing no chekpoints t…

mattsetz:
Fix bug in delete_stale_checkpoints which was causing no chekpoints t…

Before bugfix none of our model ckpts were being deleted during training, so the `regime.keep_last_n_chekpoints` and `regime.keep_best_n_chekpoints` were effectively being ignored. After bugfix these configs work as expected.


src/fairseq2/checkpoint/_manager.py
```diff
@@ -1042,4 +1044,4 @@ class CheckpointDeleteError(CheckpointError):
     def __init__(self, step_nr: int, message: str) -> None:
         super().__init__(message)
 
-        self.step_nr = step_nr
+        self.step_nr = step_nr
```
mattsetz, line 1047: i dont see a difference here?

cbalioglu (approved):
Thanks for the fix!

PR #1172, opened 2025-05-09
allow extra metrics from compute_loss, fs2 side

gilkeren1:
We may want to report more metrics from our model forward / compute_loss. This diff allows this (for scalars), and those will be reported in wandb automatically. See next PRs for examples how to use.

### DIFF
--- a/src/fairseq2/models/asr/_model.py
+++ b/src/fairseq2/models/asr/_model.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
+from typing import Dict, Tuple
 
 import torch
 from torch import Tensor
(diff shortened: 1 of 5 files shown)


mattsetz (approved):

src/fairseq2/models/asr/_model.py
```diff
@@ -8,15 +8,16 @@
 
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
+from typing import Dict, Tuple
 
 import torch
+
+from fairseq2.models.sequence import SequenceBatch
+from fairseq2.nn.padding import get_seq_lens, pad_seqs, PaddingMask
```
mattsetz, line 16: the linter cmds should change the import re-ordering back as desired.

eventually we'll wanna merge our `main_w2v2_pretraining` branch into fs2 `main` and at that point we'll need to make sure we're lint-safe. so, don't spend a lot of time on it, but if easy with the lint-fix commands, this may make our lives easier down the road.

PR #1173, opened 2025-05-09
Jacklanchantin/verifiable nonverifiable

jacklanchantin:

Fixes #{issue number}

right now, we pass all prompts to the LLM reward model for synchronization. need to optimize it. e.g.:

in SuperReward:
1. collect all prompts and task types on rank 0
2. depending on task type submit all prompts to their rewards
3. carefully scatter that back to their ranks

the biggest mismatch is that non-LLM rewards all run in parallel, but with LLM ones we have to do collective call, so here we have to unify that

**Check list:**


src/fairseq2/recipes/lm/_online_finetune/_rewards.py
```diff
@@ -264,16 +279,29 @@ def process_rollouts(
...
+                vllm_inputs.append(vllm_input)
+
             batch_text.append(rollouts_text)
             batch_tokens.append(rollouts_tokens)
             batch_rewards.append(rollouts_rewards)
 
+        # dummy vllm call for syncronization # FIXME
+        # if self.vllm_math_reward_model is not None:
+        log.info("MathVerifyVerifir generate_rewards()")
+        dummy_rewards = generate_rewards(
```
jacklanchantin, line 292: if i check for `if self.vllm_math_reward_model is not None` which it will be None for non combination runs, then i get an nccl error. not sure what's happening there

PR #1183, opened 2025-05-22
Expose load_from_path in `ModelHub`

cbalioglu:
This PR exposes `load_from_path` method of `ModelHandler` in `ModelHub` for easier use of ad-hoc models without requiring an asset card.

### DIFF
--- a/src/fairseq2/models/_hub.py
+++ b/src/fairseq2/models/_hub.py
@@ -6,6 +6,7 @@
 
 from __future__ import annotations
 
+from pathlib import Path
 from typing import Generic, TypeVar, cast, final
 
 import torch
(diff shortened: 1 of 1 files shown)


src/fairseq2/models/_hub.py
```diff
@@ -163,6 +164,57 @@ def load(
 
         return cast(ModelT, model)
 
+    def load_from_path(
+        self,
+        path: Path,
+        config: ModelConfigT,
```
mattsetz, line 170: so to use this method, we will need a ckpt path as well as a model config specifying the hps of the model? is there a mechanism to load a model solely with ckpt path (and just infer the hps from the ckpt) ?

mattsetz (approved):

src/fairseq2/models/_hub.py
cbalioglu: Unfortunately there is no way to infer hyperparameters from checkpoint. Most of the time, you need more than the checkpoint to infer them. For instance, you cannot know the number attention heads solely by inferring the checkpoint.

PR #1185, opened 2025-05-23
stateful rollout bag in online recipe, microbatching in GRPO

uralik:

* grpo unit has rollout bag that is reused when we have multiple accumulated train steps
* non-backward compatible config change: arguments in grpo loss config changed: we have group_size and forward_group_size. Changing these arguments will trigger microbatching when forward_group_size <= group_size.

DPO does not use rollout bag now since we usually use only one pair per rollout round, we can add rollout bag to GroupDPO maybe?


src/fairseq2/recipes/lm/_online_finetune/_common.py
```diff
@@ -598,3 +600,36 @@ def log_rollouts(prompt_batch: PromptBatch, rollouts, split_name, num_rollouts=1
...
+        self.reward_outputs: List = []
+
+    def maybe_reset_bag(self, trainer_step):
+        # this is called every train step to see if we need to reset the bag
+        if self._trainer_step != trainer_step:
+            # new trainer step, reset bag and counters
+            self.rollouts = []
+            self.reward_outputs = []
+            self.bag_step = 0
+            self._trainer_step = trainer_step
```
jacklanchantin, line 620: does it make sense to increment self.bag_step in an `else` block here?

PR #1194, opened 2025-06-02
Support for pairwise judges in online training

swarnaHub:
Adds support for online training with any pairwise LLM-as-a-Judge that generates real-valued scores for responses. 

Fixes #{issue number}

**Check list:**


src/fairseq2/recipes/lm/_online_finetune/_grpo.py
```diff
@@ -89,6 +88,7 @@ class GrpoFinetuneUnit(TrainUnit[SequenceBatch]):
     _sync_vllm_model_every_n_steps: int
     _sync_ref_model_every_n_steps: int
     _reward: VLLMOutputReward
+    _reward_name: str
```
uralik: can we make reward_name to be reward object attribute? this way we wont need to have +1 attribute in the finetuning unit

src/fairseq2/recipes/lm/_online_finetune/_grpo.py
```diff
@@ -161,7 +163,8 @@ def maybe_sync_models(self, force_sync_vllm=False):
     def validate_reward(self, prompt_batch: PromptBatch) -> tuple[Tensor, int]:
         if self._gangs.dp.rank == 0:
             policy_sampling_params = copy(self._vllm_model.sampling_params)
-            policy_sampling_params.n = 1
+            # For a pairwise RM, need to sample at least two judgments
+            policy_sampling_params.n = 2 if self._reward_name == "generative_pairwise_verifier" else 1
```
uralik: we have this validation sampling params where we can control this n if needed. Would it be fine if we add safety check in the grpo unit creation where if reward is pairwise one, then we update the n there?

https://github.com/facebookresearch/fairseq2/blob/online_training/src/fairseq2/recipes/lm/_online_finetune/_grpo.py#L508-L509

src/fairseq2/recipes/lm/_online_finetune/_remote_model.py
```diff
@@ -263,42 +264,76 @@ def reward_from_model(self, prompt_list, batch_size=64):
         rewards = [o.outputs.data.item() for o in ray_outputs_flat]
         return rewards
 
-    def reward_from_generative_model(self, prompt_list):
+    def reward_from_generative_model(self, prompt_list, is_pointwise):
 
-        def extract_score(output):
+        def extract_score_single(output):
```
uralik: this looks to be very prompt specific, can we somehow make this importable from _generative_prompts.py ?

src/fairseq2/recipes/lm/_online_finetune/_grpo.py
swarnaHub: done!

src/fairseq2/recipes/lm/_online_finetune/_remote_model.py
swarnaHub: as discussed offline, moved all such utils to a separate _generative_judge.py file.

src/fairseq2/recipes/lm/_online_finetune/_grpo.py
swarnaHub: Since this is only for validation, may be it's fine to hardcode n to 2 (even if users set it to some other value)? I think that was already happening with n being hardcoded to 1 for all other verifiers, so I just extended it to the pairwise case.

src/fairseq2/recipes/lm/__init__.py
```diff
@@ -171,15 +171,41 @@
...
     GenerativePointwiseVerifierHandler as GenerativePointwiseVerifierHandler,
 )
 
+from fairseq2.recipes.lm._online_finetune._rewards import (
+    GenerativePairwiseVerifier as GenerativePairwiseVerifier,
+)
+from fairseq2.recipes.lm._online_finetune._rewards import (
+    GenerativePairwiseVerifierHandler as GenerativePairwiseVerifierHandler,
+)
+
```
swarnaHub, line 172: These are generic generative-judge reward classes, which internally will call the specific extractors (that users will have to define) as follows in the reward config:

```
reward:
    name: "generative_pairwise_verifier"
    config:
        prompt_key: prompt_raw
        tokenizer: /datasets/pretrained-llms/Llama-3.1-8B-Instruct
        judgment_extractor: "j1_pairwise_score_extractor"
```
  
For scalar RMs, the "judgment_extractor" will be empty or ignored.

swarnaHub: @jacklanchantin FYI

PR #1199, opened 2025-06-08
safetensor loader speedup

uralik:

existing safetensor loader has some issues with loading shards from network filesystems. More discussion here: https://github.com/huggingface/safetensors/issues/562

So we will follow the standard fix approach where we first load shard to RAM entirely and then operate as we wish. From testing this substantially speeds up ckpt loading time.


cbalioglu:
As discussed offline, this should be the default behavior considering our compute nodes have plenty of host memory, but ideally we should reintroduce `mmap` parameter to make this optional for environments with less memory resources. We can follow-up on that later next week.

cbalioglu (approved):
LGTM! Thanks for the fixes!

PR #1200, opened 2025-06-08
qwen: fix export keymap by reusing and reversing existing importing key map

uralik:

exporting logic key map had some missing maps. Now we make one source of truth keymap in checkpoint and then reverse it in the export using the reverse function from common ckpt utils.

For now this is only in QWEN, but can be adopted in llama as well if needed?


cbalioglu (approved):
LGTM!

PR #1207, opened 2025-06-13
Introduce pack data pipeline operation

cbalioglu:
This is the first version of the `pack()` data pipeline operation used in LM pretraining parity checks. For now the API is limited to handling single tensor inputs. Later iterations should extend its feature set, in particular around handling of arbitrary target masking.


src/fairseq2/data/_data_pipeline.py
```diff
@@ -317,6 +317,16 @@ def map(
                 Can yield gains in execution speed when used with ``num_parallel_calls > 1``.
             """
 
+        def pack(
```
artemru, line 320: @cbalioglu : let's add some doc string examples to explain what happens here !

tests/unit/data/data_pipeline/test_pack.py
```diff
@@ -0,0 +1,158 @@
...
+            torch.tensor([], dtype=torch.int64),
+            torch.tensor([4, 1, 0]),
+        ]
+
+        pipeline = (
+            read_sequence(seqs).pack(num_elements=7, max_seq_len=32, truncate=False, drop_remainder=False).and_return()  # fmt: skip
+        )
+
+        expected_examples = [
+            (torch.tensor([1, 2, 3, 0, 5, 6, 0]), [4, 3]),
```
artemru, line 93: why dont we represent seq_lens as torch/numpy but just as a python list ?

PR #1208, opened 2025-06-13
Support corpus x langauage upsampling and SpecAugment across all ASR data

mattsetz:
see https://github.com/fairinternal/mms/pull/36

### DIFF
--- a/src/fairseq2/recipes/wav2vec2/asr/_train.py
+++ b/src/fairseq2/recipes/wav2vec2/asr/_train.py
@@ -312,25 +312,7 @@ def load_wav2vec2_asr_trainer(
             ) from ex
 
     # Load a full model from checkpoint
-    if model.is_empty_initialized and config.pretrained_model.name:
-        tp = (
-            AsrModel
-            if "ctc" in config.pretrained_model.name
-            or "best_checkpoint" in config.pretrained_model.name
-            else Wav2Vec2Model
-        )
-        pt_model = load_reference_model(
-            tp,
-            context,
-            config.pretrained_model,
-            gangs,
-            config.trainer.dtype,
-            mp=config.trainer.mixed_precision != "off",
-        )
-        pt_module = cast(tp, pt_model.module)
-        share_parameters(pt_module, module)
-        del pt_model
-    elif config.pretrained_model_full.name:
+    if config.pretrained_model_full.name:
         assert not config.pretrained_model.name
 
         pt_model = load_reference_model(


src/fairseq2/datasets/speech.py
```diff
@@ -39,9 +40,90 @@
...
+
+
+# SpecAugment
+@torch.no_grad()
+def freq_mask(spec: torch.Tensor, freq_mask_param: int = 80) -> torch.Tensor:
+    """Apply frequency masking to the spectrogram. Maximum mask length is set by freq_mask_param."""
+    n_freq = spec.size(-2)
+
+    assert freq_mask_param < n_freq
+    fmask_len = random.randint(20, freq_mask_param)
```
artemru, line 59: let's use numpy or torch random generators since we have already them to avoid to introduce an extra randomization system that we need to control (which does not seem to be the case here)

src/fairseq2/datasets/speech.py
```diff
@@ -359,6 +465,7 @@ def add_bucketing_pipeline(
         if isinstance(batching, LengthBatching):
             # Bucket by the audio length.
             max_num_elements = batching.max_num_elements
+            log.info(f"SETZLER. Using max_num_elements={max_num_elements}!")
```
artemru, line 468: you want to be famous)

src/fairseq2/datasets/speech.py
```diff
@@ -39,9 +40,90 @@
...
+    tmask_len = random.randint(0, time_mask_param)
+    tmask_i = random.randint(0, (n_t - tmask_len - 1))
+
+    masked_spec = spec.clone()
+    masked_spec[..., tmask_i : (tmask_i + tmask_len)] = 0.0
+    return masked_spec
+
+
+@torch.no_grad()
+def _apply_spec_augment(
```
artemru, line 83: how this is related to lang x corpus upsampling (which in the PR description) ?

src/fairseq2/datasets/speech.py
```diff
@@ -39,9 +40,90 @@
...
+    waveform: Tensor,
+    n_fft: int = 400,
+    win_len: int | None = None,
+    hop_len: int | None = None,
+    power: int | None = None,
+    freq_mask_param: int = 80,
+    time_mask_param: int = 80,
+) -> Tensor:
+    log.info(
```
artemru, line 92: I dont think we need to log here at every call ...

src/fairseq2/datasets/speech.py
artemru: you can reuse AudioCropping example (with numpy generator)

src/fairseq2/datasets/asr_parquet.py
```diff
@@ -52,9 +52,6 @@ def build_example_reading_frontend(
     ) -> Tuple[SpeechReadOptions, DataPipelineBuilder]:
         assert min_audio_len <= max_audio_len, "min_audio_len must be <= max_audio_len"
 
-        if split not in self._splits:
```
artemru, line 55: why removing this ?

src/fairseq2/datasets/speech.py
artemru: can you add some doc or reference for this method and when it should be used ?

src/fairseq2/recipes/wav2vec2/asr/_train.py
```diff
@@ -166,6 +166,22 @@ class Wav2Vec2AsrTrainDatasetSection(DatasetSection):
...
+    # SpecAugment
+    spec_aug_p: float | None = None
+    """Probability of applying SpecAugment per row."""
+    spec_aug_freq_mask_param: int = 80
+    """Maximum frequency mask length."""
+    spec_aug_time_mask_param: int = 80
+    """Maximum time mask length."""
+
+    always_read_tsv: bool = False
+    """If ``True``, always read the TSV manifest, regardless of whether parquet datasets exist."""
```
artemru, line 183: i think this's not clear in the context of fs2 trainer, should not we create our own mms trainer at some point ?

PR #1214, opened 2025-06-23
zero-shot to commit

gilkeren1:
Zero-shot LLM-style ASR, the fs2 side.

black, flake8, isort, mypy cleared all files. 
Running on RSC to check everything still works:
CTC test: mms2_7b_mms1_m2c2v1v2_spc_jhu_400k_bib1143_seed2_test_zs5 (10444855)
llm-style test: mms2_7b_mms1_m2c2v1v2_spc_jhu_400k_bib1143_llama_zs_seed2_test_zs4 (10444848)

### DIFF
--- a/src/fairseq2/datasets/speech.py
+++ b/src/fairseq2/datasets/speech.py
@@ -147,7 +147,6 @@ def postprocess(
 
 
 class AudioCropper:
-
     audio_feature: str = "audio_feature"
 
     def __init__(
@@ -177,7 +176,6 @@ class AudioCropper:
 
 @dataclass(kw_only=True)
 class SpeechReadOptions(DataReadOptions):
-
     dtype: DataType = torch.float32
     """The data type of the decoded audio sequences."""
 
@@ -207,6 +205,17 @@ class SpeechReadOptions(DataReadOptions):
     """Maximum frequency mask length."""
     spec_aug_time_mask_param: int = 80
     """Maximum time mask length."""
+
+    # Zero-shot args
+    n_context_examples: int = 0
+    """The number of context examples to use when providing context."""
+
+    bucket_size: int = 2000
+    """Minimum size of pool for choosing context examples."""
+
+    deterministic_context: bool = False
+    """If ``True``, the context examples will be selected deterministically from the \
+    audio path. Should be True for eval sets and False for train sets."""
 
 
 class SpeechDataset(ABC):
@@ -247,7 +256,6 @@ get_speech_dataset_hub = DatasetHubAccessor(SpeechDataset)
 
 
 class ManifestDatasetInterface:
-
     _name: str
     _manifest_dir: Path
     _splits: set[str]
@@ -311,7 +319,10 @@ class GenericSpeechDataset(ManifestDatasetInterface, SpeechDataset):
 
     @staticmethod
     def add_audio_decoding(
-        builder: DataPipelineBuilder, options: SpeechReadOptions, audio_dir: Path | None
+        builder: DataPipelineBuilder,
+        options: SpeechReadOptions,
+        audio_dir: Path | None,
+        selector: str = "[*].audio",
     ) -> DataPipelineBuilder:
         # Memory map audio files.
         cached_fd_count = options.extras.get("cached_fd_count", 1000)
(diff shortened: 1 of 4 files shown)


src/fairseq2/models/seq2seq.py
```diff
@@ -100,6 +100,36 @@ def to(self, device: Device) -> None:
...
+                )
+                self.example["context_text"][i]["seq_lens"] = self.example[  # type: ignore
+                    "context_text"
+                ][
+                    i
+                ][
+                    "seq_lens"
+                ].to(
+                    device
+                )
```
artemru: can you create a small dedicated function for this transformation to better structure this code (and to not modify the orginal fs2 function used by others) ?
also adding some comments will be helpful.

src/fairseq2/recipes/wav2vec2/asr/_train.py
mattsetz, line 1: Can we also land the code-snippet to support seeding with a wav2vec2_asr model ? Something like...

```python
    if model.is_empty_initialized and config.pretrained_model.name:
        tp = (
            AsrModel
            # TODO (gilkeren): find a nicer way to do this
            if "ctc" in config.pretrained_model.name
            or "best_checkpoint" in config.pretrained_model.name
            else Wav2Vec2Model
        )
        pt_model = load_reference_model(
            tp,
            context,
            config.pretrained_model,
            gangs,
            config.trainer.dtype,
            mp=config.trainer.mixed_precision != "off",
        )
        pt_module = cast(tp, pt_model.module)
    elif config.pretrained_model_full.name:
        assert not config.pretrained_model.name

        pt_model = load_reference_model(
            type(model.module),
            context,
            config.pretrained_model_full,
            gangs,
            config.trainer.dtype,
            mp=config.trainer.mixed_precision != "off",
        )
        pt_module = pt_model.module  # type: ignore[assignment]
        share_parameters(pt_module, module)
        del pt_model

        # Make sure that the final projection layer is instantiated along with
        # the pretrained parameters if it was on the meta device.
        if gangs.dp.rank == 0:
            to_device(module, gangs.root.device)

        try:
            gangs.root.barrier()
        except GangError as ex:
            raise RecipeError(
                "The collective barrier after the pretrained model load operation has failed. See the nested exception for details."
            ) from ex
```

gilkeren1: done

src/fairseq2/models/seq2seq.py
gilkeren1: done

mattsetz (approved):

src/fairseq2/datasets/speech.py
```diff
@@ -208,6 +206,17 @@ class SpeechReadOptions(DataReadOptions):
...
     """Maximum time mask length."""
 
+    # Zero-shot args
+    n_context_examples: int = 0
+    """The number of context examples to use when providing context."""
+
+    bucket_size: int = 2000
+    """Minimum size of pool for choosing context examples."""
+
+    deterministic_context: bool = False
```
artemru, line 216: it's not clear is it can be deterministic random generator here ...

src/fairseq2/recipes/wav2vec2/asr/_train.py
```diff
@@ -416,6 +422,9 @@ def load_wav2vec2_asr_trainer(
         spec_aug_p=config.dataset.spec_aug_p,
         spec_aug_freq_mask_param=config.dataset.spec_aug_freq_mask_param,
         spec_aug_time_mask_param=config.dataset.spec_aug_time_mask_param,
+        n_context_examples=config.dataset.n_context_examples,
+        bucket_size=2000,
```
artemru, line 426: why hard coding this ?

src/fairseq2/recipes/wav2vec2/asr/_train.py
```diff
@@ -459,6 +468,9 @@ def load_wav2vec2_asr_trainer(
             num_prefetch=config.dataset.num_prefetch,
             seed=seed,
             extras=config.dataset.extras,
+            n_context_examples=config.dataset.n_context_examples,
+            bucket_size=30,
```
artemru, line 472: and why this value ?

src/fairseq2/recipes/wav2vec2/asr/_train.py
gilkeren1: because I don't expect this to even change, so preferred not overloading our config with such values. 
I know it's generally "frowned upon" in the coding world to have numeric values in the code but I think in certain cases, like this one, it could be ok. If this bothers you I can move this to the config though

src/fairseq2/datasets/speech.py
gilkeren1: not sure exactly what you are asking
but to clarify for dev sets we want to always choose the same context examples, so we set this flag to True, and choose the "random" context examples with a seed computed deterministically from the path of the audio sample.

PR #1216, opened 2025-06-24
Introduce a basic jsonl dataset

cbalioglu:
This PR introduces a basic packed JSONL dataset used in LLaMA pretraining parity checks.


src/fairseq2/data/_data_pipeline.py
```diff
@@ -325,7 +325,46 @@ def pack(
             truncate: bool = True,
             drop_remainder: bool = True,
             pinned_memory: bool = False,
-        ) -> Self: ...
+        ) -> Self:
+            """
+            Packs input sequences into a batch by concataneting them till the
```
artemru, line 330: what happens to some elements that are not tensors ?
how to indicate on which of elements apply it ?
will it work one has a sequence of embs instead of tokens ?

there's some hidden hypothesis about the elements structure here that makes this operations very specific.

src/fairseq2/datasets/jsonl.py
```diff
@@ -0,0 +1,137 @@
...
+
+    @staticmethod
+    def from_path(path: Path, name: str) -> JsonlDataset:
+        path = path.expanduser().resolve()
+
+        if not path.is_dir():
+            files = [path]
+        else:
+            try:
+                files = [f for f in path.glob("**/*.chunk.*.jsonl") if not f.is_dir()]
```
artemru, line 59: why this pattern ? *.chunk.*.jsonl` what if my files have a different names ?

src/fairseq2/datasets/jsonl.py
```diff
@@ -0,0 +1,137 @@
...
+    _name: str
+    _files: Sequence[Path]
+
+    def __init__(self, name: str, files: Sequence[Path]) -> None:
+        self._name = name
+        self._files = files
+
+    @staticmethod
+    def from_path(path: Path, name: str) -> JsonlDataset:
+        path = path.expanduser().resolve()
```
artemru, line 53: this seems to be NFS specific, why not using filesytem abstraction ?

src/fairseq2/datasets/jsonl.py
```diff
@@ -0,0 +1,137 @@
...
+
+        if len(self._files) < file_world_size:
+            raise NotSupportedError(
+                "The number of dataset files must be greater than or equal to the number of world size."
+            )
+
+        builder = read_sequence(self._files)
+
+        if file_world_size > 1:
+            builder.shard(file_rank, file_world_size, allow_uneven=True)
```
artemru, line 94: so in case of unbalanced json files it can be ill balanced between ranks.
we need at least to notify a user about it

src/fairseq2/datasets/jsonl.py
```diff
@@ -0,0 +1,137 @@
...
+
+        def read_file(file: Path) -> DataPipeline:
+            return read_text(file).map(json.loads, num_parallel_calls=1).and_return()
+
+        builder.yield_from(read_file)
+
+        # Tokenize.
+        def encode(example: dict[str, Any]) -> Tensor:
+            return text_encoder(example["text"])
```
artemru, line 103: why text is hard coded here, and if it's called transcription ?

src/fairseq2/datasets/jsonl.py
```diff
@@ -0,0 +1,137 @@
...
+                "The number of dataset files must be greater than or equal to the number of world size."
+            )
+
+        builder = read_sequence(self._files)
+
+        if file_world_size > 1:
+            builder.shard(file_rank, file_world_size, allow_uneven=True)
+
+        def read_file(file: Path) -> DataPipeline:
+            return read_text(file).map(json.loads, num_parallel_calls=1).and_return()
```
artemru, line 97: can we restrict it only to the subset of useful columns ?

src/fairseq2/datasets/jsonl.py
```diff
@@ -0,0 +1,137 @@
...
+            builder.take(options.max_num_batches)
+
+        # Prefetch `num_prefetch` batches in background.
+        builder.prefetch(options.num_prefetch)
+
+        # Convert to `SequenceBatch`.
+        def to_batch(example: dict[str, Any]) -> SequenceBatch:
+            seqs, seq_lens = example["seqs"], example["seq_lens"]
+
+            return SequenceBatch(seqs, seq_lens, packed=True)
```
artemru, line 131: so, we cannot propagate any other data from batch here ?

src/fairseq2/recipes/lm/_train.py
```diff
@@ -215,8 +203,6 @@ def load_clm_trainer(
 
     read_options = TextReadOptions(
         batching=batching,
-        example_shuffle_window=config.dataset.example_shuffle_window,
```
artemru, line 218: why removing this ones ? why not supporting the example shuffling before packing ?
for me it still makes sense, and if one dont want it can deactivate it with example_shuffle_window param, no ?

src/fairseq2/datasets/jsonl.py
```diff
@@ -0,0 +1,137 @@
...
+    LengthBatching,
+    SequenceBatch,
+)
+from fairseq2.datasets.text import TextDataset, TextReadOptions
+from fairseq2.error import NotSupportedError
+from fairseq2.gang import Gang
+from fairseq2.nn import BatchLayout
+
+# TODO: FIX, INFER
+npc = 10
```
artemru, line 36: this's not used at all

src/fairseq2/datasets/jsonl.py
```diff
@@ -0,0 +1,137 @@
...
+        options: TextReadOptions | None = None,
+    ) -> DataReader[SequenceBatch]:
+        if options is None:
+            options = TextReadOptions()
+
+        file_rank = gang.rank
+
+        file_world_size = gang.size
+
+        if len(self._files) < file_world_size:
```
artemru, line 86: in theory, we could shard some jsonl files as well ...

src/fairseq2/data/_data_pipeline.py
cbalioglu: The current implementation only supports single tensor input (with chosen/rejected and target masking support soon). This is not really a general purpose pack operation (which would be substantially more complex to implemented), but specifically tuned towards LM pretraining/finetuning/preference optimization.

src/fairseq2/datasets/jsonl.py
cbalioglu: yeah, right now none of the dataset implementations use the `FileSystem` abstraction. We should gradually move them to use it.

src/fairseq2/datasets/jsonl.py
cbalioglu: As mentioned offline, this dataset is implemented specifically for LM pretraining recipe and will be moved to its directory soon with the recipe authoring PR.

src/fairseq2/datasets/jsonl.py
cbalioglu: Since `DataReader` always guarantees the same number of batches are seen in each rank, this is mostly safe. Also since each jsonl shard contains the same number of examples, ranks will stop once they reach the end of first file in their queue.

src/fairseq2/datasets/jsonl.py
cbalioglu: This assumes that the jsonl files contain only a single `text` field (or `src`, `tgt`, `tgt_chosen`, `tgt_rejected` once finetuning support lands)

src/fairseq2/datasets/jsonl.py
cbalioglu: Again, as mentioned above. This dataset implementation is meant only for LM pretraining recipe. Hopefully we will move it soon to the recipe directory.

src/fairseq2/datasets/jsonl.py
cbalioglu: same as above

src/fairseq2/datasets/jsonl.py
artemru: `each jsonl shard contains the same number of examples` what guarantees that ?

src/fairseq2/datasets/jsonl.py
artemru: *chunk* is very particular naming pattern, I dont see any reason it to be so

src/fairseq2/datasets/jsonl.py
artemru: let included those requirement explicitly

PR #1220, opened 2025-06-30
Main parquet text loader

artemru:
* Enable text dataloading from parquet dataset (shared at row groups levels)
* adding split to text and jsonl dataloader reader to unify the interface
* fixing npc at other different dataloader

No BC, only new feature

**Check list:**

### DIFF
--- a/src/fairseq2/datasets/instruction.py
+++ b/src/fairseq2/datasets/instruction.py
@@ -127,10 +127,6 @@ class InstructionDataset(ABC):
         """Return the set of splits."""
 
 
-# TODO: FIX, INFER
-npc = 10
-
-
 GENERIC_INSTRUCTION_DATASET_FAMILY: Final = "generic_instruction"
 
 
@@ -238,8 +234,8 @@ class GenericInstructionDataset(InstructionDataset):
         source_encoder = tokenizer.create_encoder(mode=options.source_encode_mode)
         target_encoder = tokenizer.create_encoder(mode=options.target_encode_mode)
 
-        builder.map(source_encoder, selector="src", num_parallel_calls=npc)
-        builder.map(target_encoder, selector="tgt", num_parallel_calls=npc)
+        builder.map(source_encoder, selector="src", num_parallel_calls=1)
+        builder.map(target_encoder, selector="tgt", num_parallel_calls=1)
 
         def cat_source_and_target(example: dict[str, Any]) -> dict[str, Any]:
             id_ = example.get("id")
@@ -253,7 +249,7 @@ class GenericInstructionDataset(InstructionDataset):
 
             return {"id": id_, "indices": indices, "target_mask": target_mask}
 
-        builder.map(cat_source_and_target, num_parallel_calls=npc)
+        builder.map(cat_source_and_target, num_parallel_calls=1)
 
         batching = options.batching
 
@@ -301,7 +297,7 @@ class GenericInstructionDataset(InstructionDataset):
             pad_value=tokenizer.vocab_info.pad_idx, overrides=[target_mask_collate_opts]
         )
 
-        builder.map(collater, num_parallel_calls=npc)
+        builder.map(collater, num_parallel_calls=options.npc)
 
         # Return only the first `max_num_batches`.
         if options.max_num_batches is not None:
@@ -373,7 +369,7 @@ class GenericInstructionDataset(InstructionDataset):
 
             return {"id": id_, "prompt": source, "indices": indices}
 
-        builder.map(encode, num_parallel_calls=npc)
+        builder.map(encode, num_parallel_calls=1)
 
         # Filter out long examples.
         def skip(example: dict[str, Any]) -> bool:
@@ -394,7 +390,7 @@ class GenericInstructionDataset(InstructionDataset):
         # Collate bucketed examples into a batch.
         collater = Collater(pad_value=tokenizer.vocab_info.pad_idx or 0)
 
-        builder.map(collater, num_parallel_calls=npc)
+        builder.map(collater, num_parallel_calls=options.npc)
 
         # Prefetch `num_prefetch` batches in background.
         builder.prefetch(options.num_prefetch)
@@ -421,7 +417,7 @@ class GenericInstructionDataset(InstructionDataset):
             for line in fp:
                 lines.append(line)
 
-        return read_sequence(lines).map(json.loads, num_parallel_calls=npc)
+        return read_sequence(lines).map(json.loads, num_parallel_calls=1)
 
     @override
     def splits(self) -> set[str]:
--- a/src/fairseq2/datasets/jsonl.py
+++ b/src/fairseq2/datasets/jsonl.py
@@ -14,10 +14,7 @@ from typing import Any, Final, final
 from torch import Tensor
 from typing_extensions import override
 
-from fairseq2.data import (
-    DataPipeline,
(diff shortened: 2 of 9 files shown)


src/fairseq2/datasets/jsonl.py
```diff
@@ -110,28 +144,26 @@ def encode(example: dict[str, Any]) -> Tensor:
             max_num_elements = batching.max_num_elements
 
             builder.pack(
-                max_num_elements + 1, max_seq_len, truncate=True, pinned_memory=True
+                max_num_elements + 1,
+                max_seq_len,
+                pad_value=pad_idx,
```
artemru, line 158: @cbalioglu : pad_value is not documented, I hope I used it correctly

src/fairseq2/datasets/text.py
```diff
@@ -123,16 +149,20 @@ def create_reader(
         min_seq_len: int,
         max_seq_len: int,
         options: TextReadOptions | None = None,
+        split: str | None = None,
```
artemru, line 165: this will keep the previous behaviour by default

src/fairseq2/datasets/text_parquet.py
```diff
@@ -0,0 +1,222 @@
...
+        if split is not None:
+            fragment_config = fragment_config.add_partition_filter(
+                pa.compute.field("split") == split
+            )
+        fragement_builder = ParquetFragmentStreamer(
+            config=fragment_config
+        ).build_pipeline(rank=rank, world_size=world_size)
+
+        num_parallel_fragments = options.extras.get("num_parallel_fragments", npc)
+        assert isinstance(num_parallel_fragments, int)
```
zyaoj: For the `assert`s, could we add more error messages to make it as instructive as possible?

src/fairseq2/datasets/text_parquet.py
```diff
@@ -0,0 +1,222 @@
...
+        )
+        options.seed += gang.rank
+
+        pipeline = JsonlDataset.build_pipeline_backend(
+            builder,
+            options,
+            text_encoder,
+            pad_idx=pad_idx,
+            max_seq_len=max_seq_len,
+            text_column_name="text",
```
zyaoj, line 225: This hardcode the text column name to "text" instead of allowing dynamic configuration of the target. For example say if I want to load `sampled_sonar_wiki_lingua.parquet`, whose `source` and `target` columns are text columns.  Shall we set `text_column_name` as an input argument to this functional interface?

src/fairseq2/datasets/instruction.py
```diff
@@ -238,8 +234,8 @@ def create_reader(
         source_encoder = tokenizer.create_encoder(mode=options.source_encode_mode)
         target_encoder = tokenizer.create_encoder(mode=options.target_encode_mode)
 
-        builder.map(source_encoder, selector="src", num_parallel_calls=npc)
-        builder.map(target_encoder, selector="tgt", num_parallel_calls=npc)
+        builder.map(source_encoder, selector="src", num_parallel_calls=1)
```
zyaoj: shall we change this implicit default to an `options` field for early-stage/light work?

src/fairseq2/datasets/text.py
```diff
@@ -114,6 +115,31 @@ def from_path(path: Path, name: str) -> GenericTextDataset:
...
+        """Filter the dataset files by split.
+
+        :param split:
+            The split to filter by.
+        """
+        if split is None:
+            return files
+
+        # filtering based on two following patterns:
+        pattern1 = f"**/*{split}*.{extention}"
```
zyaoj: is this going to fail for some cases like `constraint.parquet`?

src/fairseq2/datasets/text_parquet.py
zyaoj: After hardcoding the column to `source` I still got `KeyError: 'source'`. Is there something I'm missing?

src/fairseq2/datasets/text.py
artemru: this's only for jsonl or text examples

src/fairseq2/datasets/text_parquet.py
artemru: you need to change the value of `DefaultTextSchema.text = "source" ! note that this class is universally responsible of renaming input columns into the target schema.
That being said we can use only add an extra param here.

artemru: see the doc string for DefaultTextSchema

src/fairseq2/datasets/instruction.py
artemru: I dont completely get the idea, but here in any case, putting num_parallel_calls>1 will only result in the slowdown since the fn is GIL strongly bound.

zyaoj: I think the idea overall is to avoid magic numbers. There could be 2 options:
- add some comments there to note the choice of this config
- set it as a field in the options as something like `npc_gil_bound_job`

artemru: ok, we can rm num_parallel_calls=1 here at all as an arg to avoid an extra param at all

src/fairseq2/datasets/text.py
zyaoj: Yeah I think the point here is not about the file extension, but more about the regex pattern for file names. This might be a nit but there could be some edge cases like:
- when `split = "train"`, if the file is named as `wiki-constraint-test.jsonl` or `wiki-constraint-test.*`, though it's probably a test split dataset, it contains `train` in `constraint` which might accidentally include a test split file into the train split. Wdyt?

src/fairseq2/datasets/text_parquet.py
zyaoj: Ahah I indeed missed the `DefaultTextSchema` and this works like a charm.

src/fairseq2/datasets/text.py
artemru: ah, I see, ok what about the strict check ? so only train.jsonl will be accepted ?

artemru: ( I went for most the strict schema), somehow for text files there's no universal conversion for partitioning unlikely for parquet)

zyaoj: Yeah I guess we can put a strict check as a default and add an input argument that accepts split patterns, something like:

```py
@staticmethod
def filter_split(
    files: Sequence[Path],
    split: str | None,
    extention: str,
    *,
    split_pattern: str | Sequence[str] | None = None,  # ← change here
) -> Sequence[Path]:
    if split is None:
        return list(files)

    # Prepare pattern list
    if split_pattern is None:
        patterns: list[str] = [
            f"**/{split}.{extention}",
            f"{split}/*.{extention}",
        ]
    else:
        # Accept str, list[str], tuple[str], … and interpolate placeholders
        raw = [split_pattern] if isinstance(split_pattern, str) else list(split_pattern)
        patterns = [pattern.format(split=split, ext=extention) for pattern in raw]

    # Filter paths 
    return [path for path in files if any(fnmatch.fnmatch(str(path), pattern) for pattern in patterns)]
```

artemru: good idea indeed !

artemru: I took it

zyaoj (approved):
LGTM! Thanks for addressing the comments.

PR #1222, opened 2025-07-02
Filtering empty text examples

artemru:
Removing empty text examples in ASR dataloaders
- This requires to move tokenization before bucketing!

**Check list:**

### DIFF
--- a/src/fairseq2/datasets/asr.py
+++ b/src/fairseq2/datasets/asr.py
@@ -165,6 +165,7 @@ class GenericAsrDataset(ManifestDatasetInterface, AsrDataset):
         builder: DataPipelineBuilder,
         tokenizer: TextTokenizer,
     ) -> DataPipelineBuilder:
+        # Tokenize target text.
         text_encoder = tokenizer.create_encoder()
         builder.map(text_encoder, selector="text")
 
--- a/src/fairseq2/datasets/asr.py
+++ b/src/fairseq2/datasets/asr.py
@@ -8,17 +8,19 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from functools import partial
-from typing import Any, cast, Dict, Final, Tuple
+from typing import Any, Dict, Final, Tuple, cast
+
+from typing_extensions import override
 
 from fairseq2.data import (
     CollateOptionsOverride,
     Collater,
     DataPipeline,
     DataPipelineBuilder,
-    read_sequence,
     SequenceData,
+    read_sequence,
 )
-from fairseq2.data.text import read_text, StrSplitter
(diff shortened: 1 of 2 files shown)


src/fairseq2/datasets/asr.py
```diff
@@ -160,6 +160,28 @@ def create_reader(
             self._name, split, builder.and_return(), gang, options
         )
 
+    @staticmethod
+    def add_tokenization_pipeline(
+        builder: DataPipelineBuilder,
+        tokenizer: TextTokenizer,
```
mattsetz, line 166: we can add another arg here `remove_unk` to optionally remove all UNKs tokens from transcript. we should default this to False but it can be something that we experiment with.

src/fairseq2/datasets/asr.py
```diff
@@ -160,6 +160,28 @@ def create_reader(
...
+
+        # to avoid to tokenize empty text, we filter out them out first
+        builder = builder.filter(lambda x: bool(len(x["text"]) > 0))
+
+        builder.map(text_encoder, selector="text")
+
+        unk_idx = tokenizer.vocab_info.unk_idx
+
+        def empty_text(example: Dict[str, Any]) -> bool:
+            return bool((example["text"] != unk_idx).sum().item() > 0)
```
mattsetz, line 180: nit: `example["text"] != unk_idx).numel()` ?

src/fairseq2/datasets/asr.py
mattsetz: also, maybe we can log the number of UNKs (or % of `num_unks/total_tokens` within a sample/batch) in weights and biases? its useful for us to know if our dataset contains too many UNKs.

src/fairseq2/datasets/asr.py
artemru: no, it's not the same https://docs.pytorch.org/docs/stable/generated/torch.numel.html#torch-numel
example["text"] != unk_idx is boolean mask if we need to count True elements and all elements

src/fairseq2/datasets/asr.py
artemru: > also, maybe we can log the number of UNKs (or % of `num_unks/total_tokens` within a sample/batch) in weights and biases? its useful for us to know if our dataset contains too many UNKs.

it's possible (requires a bit to propagate some metric bags to data reader) but for now we can just do in a adhoc analysis

artemru: > we can add another arg here `remove_unk` to optionally remove all UNKs tokens from transcript. we should default this to False but it can be something that we experiment with.

done !

PR #1228, opened 2025-07-15
chat template overwriting, chat mode in SFT dataset

uralik:

We now can use chat option in LLM recipes

### DIFF
--- a/src/fairseq2/data/text/tokenizers/llama.py
+++ b/src/fairseq2/data/text/tokenizers/llama.py
@@ -37,6 +37,21 @@ from fairseq2.data.text.tokenizers.tiktoken import (
 )
 from fairseq2.device import Device
 
+# llama3 chat template with assistant mask support, see https://github.com/huggingface/transformers/issues/28950
+LLAMA3_CHAT_TEMPLATE = """{{- bos_token }}
+{%- for message in messages %}
+    {%- if message.role == 'user' or message.role == 'system' %}
+        {{- '<|start_header_id|>' + message['role'] + '<|end_header_id|>\\n\\n' + message['content'] | trim + '<|eot_id|>' }}
+    {%- else %}
+        {%- generation %}
+        {{- '<|start_header_id|>assistant<|end_header_id|>\\n\\n' + message['content'] | trim + '<|eot_id|>' }}
+        {%- endgeneration %}
+    {%- endif %}
+{%- endfor %}
+{%- if add_generation_prompt %}
+    {{- '<|start_header_id|>assistant<|end_header_id|>\\n\\n' }}
+{%- endif %}"""
+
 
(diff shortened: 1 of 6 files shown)


src/fairseq2/data/text/tokenizers/llama.py
```diff
@@ -37,6 +37,21 @@
 )
 from fairseq2.device import Device
 
+# llama3 chat template with assistant mask support, see https://github.com/huggingface/transformers/issues/28950
```
jacklanchantin, line 40: is this template correct? looks different than https://huggingface.co/meta-llama/Meta-Llama-3-8B-Instruct/blob/main/tokenizer_config.json#L2053

uralik: yes it looks different, because original llama template is not correct w.r.t. assistant masks, it just doesnt support it. And I removed all tool related things which are not used within fairseq2 anyway

src/fairseq2/datasets/preference.py
```diff
@@ -201,66 +204,142 @@ def create_reader(
...
-            )
+                indices_rejected = encoded_output_rejected["input_ids"][0]
+                target_mask_rejected = encoded_output_rejected["assistant_masks"][
+                    0
+                ].bool()
+
+                if not options.mask_source_tokens:
+                    # no source masking i.e. mask has all 1s
+                    target_mask_chosen._fill(True)
+                    target_mask_rejected._fill(True)
```
jacklanchantin: this isn't in instruction.py encoding_chat() function. is that correct?

uralik: yes, SFT doesnt have src masking support, it only exists in DPO since one team figured it helped them 😅

src/fairseq2/datasets/preference.py
```diff
@@ -201,66 +204,142 @@ def create_reader(
...
+                    "reference_score_chosen": example.get(
+                        "reference_score_chosen", None
+                    ),
+                    "reference_score_rejected": example.get(
+                        "reference_score_rejected", None
+                    ),
+                    "target_mask_chosen": target_mask_chosen,
+                    "target_mask_rejected": target_mask_rejected,
+                    "total_tokens": total_tokens,
+                    "keep_jsonl_keys": jsonl_content,
```
jacklanchantin, line 342: these keys are also not returned in instruction.py. not sure if we want those

jacklanchantin (approved):

src/fairseq2/datasets/preference.py
uralik: yep we use total tokens in some of the metrics in preference recipe, and keep_jsonl_keys is used when we pass reference scores precomputed offline

PR #1229, opened 2025-07-16
Add a basic flatten operator for data pipelines

MartinGleize:
Adds an operator turning a pipeline of lists into a pipeline of their flattened elements.

None

**Check list:**

### DIFF
--- a/native/src/fairseq2n/data/flatten_data_source.cc
+++ b/native/src/fairseq2n/data/flatten_data_source.cc
@@ -1,0 +1,122 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+// All rights reserved.
+//
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+
+#include "fairseq2n/data/flatten_data_source.h"
+
+#include <queue>
+#include <string>
+#include <utility>
+
+#include "fairseq2n/data/data.h"
+#include "fairseq2n/data/detail/exception.h"
+
+namespace fairseq2n::detail {
+
+std::optional<data>
+flatten_data_source::next()
+{
+    // If we have elements in the queue, return the next one
+    if (!elements_queue_.empty()) {
+        data element = std::move(elements_queue_.front());
+        elements_queue_.pop();
+        return element;
+    }
+
+    // Get the next example from the inner data source
+    std::optional<data> maybe_example = inner_->next();
+    if (!maybe_example)
+        return std::nullopt;
+
+    // Extract elements from the example and add them to the queue
+    elements_queue_ = extract_elements(*maybe_example);
+
+    // If the queue is empty (no elements could be extracted), recursively call next()
+    if (elements_queue_.empty())
+        return next();
+
+    // Return the first element from the queue
+    data element = std::move(elements_queue_.front());
+    elements_queue_.pop();
+    return element;
+}
+
+void
+flatten_data_source::reset(bool reset_rng)
+{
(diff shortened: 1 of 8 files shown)


native/src/fairseq2n/data/flatten_data_source.h
```diff
@@ -0,0 +1,57 @@
...
+private:
+    // Extracts elements from a data object based on the selector.
+    // Returns a queue of elements to be returned by next().
+    std::queue<data>
+    extract_elements(const data &example);
+
+    std::unique_ptr<data_source> inner_;
+    std::optional<std::string> selector_;
+    std::queue<data> elements_queue_;
+    //std::exception_ptr exception_ptr_{};
```
cbalioglu, line 54: Not used?

native/src/fairseq2n/data/flatten_data_source.cc
```diff
@@ -0,0 +1,122 @@
...
+std::queue<data>
+flatten_data_source::extract_elements(const data &example)
+{
+    std::queue<data> elements;
+
+    // If selector is provided, use it to extract elements
+    if (selector_) {
+        std::string error_msg = "Cannot use selector with flatten.";
+        std::cerr << error_msg << std::endl;
+        std::terminate();  // Force immediate program termination
```
cbalioglu, line 108: I would prefer to raise a "not implemented exception" here, instead of forcibly terminating the process. In data_pipeline.cc:93 we mark the pipeline as broken if we catch an unhandled exception.

native/src/fairseq2n/data/flatten_data_source.cc
```diff
@@ -0,0 +1,122 @@
...
+    }
+    // If no selector is provided, assume the example itself is a list
+    else if (example.is_list()) {
+        auto example_list = example.as_list();
+        for (size_t i = 0; i < example_list.size(); ++i) {
```
cbalioglu, line 113: My understanding when looking at `next()` implementation, we can actually use move semantics here. So the signature of `extract_elements` can be:

```
extract_elements(data &&example);
```

If that works, you can instead use `element.emplace(std::move(example_list[i]))`.

native/src/fairseq2n/data/flatten_data_source.cc
```diff
@@ -0,0 +1,122 @@
...
+            elements.push(example_list.at(i));
+        }
+    }
+    // If no valid elements were found, the queue remains empty
```
cbalioglu, line 117: Does this mean we simply discard/drop the `example` we just read from the underlying data source?

native/src/fairseq2n/data/flatten_data_source.cc
```diff
@@ -0,0 +1,122 @@
...
+    std::optional<data> maybe_example = inner_->next();
+    if (!maybe_example)
+        return std::nullopt;
+
+    // Extract elements from the example and add them to the queue
+    elements_queue_ = extract_elements(*maybe_example);
+
+    // If the queue is empty (no elements could be extracted), recursively call next()
+    if (elements_queue_.empty())
+        return next();
```
cbalioglu, line 38: Wouldn't be cleaner if we had this:

```
while (elements_queue_.empty()) {
  std::optional<data> maybe_example = ...
  elements_queue = extract_elements(*maybe_example);
 ```

cbalioglu (commented):
Thanks for the PR! Left a couple comments.

artemru:
i think i can be achieved with yield_from pattern ?

PR #1232, opened 2025-07-22
skywork and some qwrn changes

swarnaHub:

Fixes #{issue number}

**Check list:**


src/fairseq2/recipes/lm/_online_finetune/_online_dpo.py
```diff
@@ -142,12 +134,16 @@ def validate_reward(
...
         if self._gangs.dp.rank == 0:
             policy_sampling_params = copy(self._vllm_model.sampling_params)
-            policy_sampling_params.n = 1
             for (
                 k,
                 v,
             ) in self._config.loss_config.validation_vllm_sampling_params.items():
                 policy_sampling_params.__setattr__(k, v)
+
+            # For a pairwise RM, need to sample at least two judgments
```
jacklanchantin: this is 2 rollouts per prompt, but i assume there are 2 copies of the prompt (2 different orders)? so shouldn't you only need 1 rollout for each order?

swarnaHub: sorry, this is a typo! I meant 2 "rollouts" instead of "judgments". 

How many judgment prompts are created out of those two rollouts (i.e., 2 with (a,b) and (b,a) in both orders) will be automatically handled with the "all_pairs" setting in the reward config. So, if it's all_pairs, we'll have N * (N-1) judgments and N = 2, so we'll have 2 judgments. Let me know if this makes sense!

jacklanchantin: yep, makes sense now

PR #1235, opened 2025-07-25
Gilkeren/debug2

gilkeren1:
Not for landing


src/fairseq2/datasets/asr_parquet.py
```diff
@@ -154,6 +154,11 @@ def build_parquet_audio_text_reading(
             builder, options, GenericSpeechDataset.rename_feature
         )
 
+        # Tokenizer langauge name as well
```
gilkeren1: This file for tokenizing lang name

PR #1241, opened 2025-08-05
adding bleu computation for S2tt asr training

artemru:

Fixes #{issue number}

**Check list:**

### DIFF
--- a/src/fairseq2/models/wav2vec2/asr/_config.py
+++ b/src/fairseq2/models/wav2vec2/asr/_config.py
@@ -10,8 +10,7 @@ from dataclasses import dataclass, field
 from typing import Final
 
 from fairseq2.context import RuntimeContext
-from fairseq2.data import VocabularyInfo
-from fairseq2.models.wav2vec2 import Wav2Vec2EncoderConfig
+from fairseq2.models.wav2vec2 import Wav2Vec2Config, Wav2Vec2EncoderConfig
 
 WAV2VEC2_ASR_MODEL_FAMILY: Final = "wav2vec2_asr"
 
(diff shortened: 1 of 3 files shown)


src/fairseq2/nn/utils/module.py
```diff
@@ -464,6 +464,21 @@ def load_state_dict(
     ``state_dict`` does not contain any keys corresponding to descendants that are set to ``None``
     via :meth:`Module.register_module()`.
     """
+    # Key mapping
```
gilkeren1: Is this section here on purpose? I think it got here from one of my local diffs and should not be landed

src/fairseq2/models/wav2vec2/asr/_config.py
```diff
@@ -160,7 +160,7 @@ def bib61_300m() -> Wav2Vec2AsrConfig:
     @wav2vec2_asr_arch("300m_bib1143")
     def bib1143_300m() -> Wav2Vec2AsrConfig:
         config = bib61_300m()
-        config.vocab_info.size = 3335
+        config.vocab_info.size = 3292
```
gilkeren1: same comment as below

src/fairseq2/nn/utils/module.py
zyaoj: +1 this seems to be irrelevant to this PR and might be specific to special cases?

src/fairseq2/models/wav2vec2/asr/_config.py
artemru: thanks

PR #1246, opened 2025-08-12
Fix drop_remainder bug in pack op

cbalioglu:
This PR fixes an edge case bug where the `drop_remainder` option does not get respected by the `pack` operator if the current batch does not get filled entirely in truncate mode.


uralik (approved):
thanks!!!!

PR #1250, opened 2025-08-19
Bleu memory leak

artemru:

Fixes #{issue number}

**Check list:**


src/fairseq2/metrics/text/_bleu.py
```diff
@@ -65,14 +76,16 @@ def update(self, refs: Sequence[str], hyps: Sequence[str]) -> Self:
...
         self.sys_len += bleu.sys_len
         self.ref_len += bleu.ref_len
 
-        self.valid_ngrams += torch.tensor(bleu.counts, device=self.device)
-        self.total_ngrams += torch.tensor(bleu.totals, device=self.device)
-
+        self.valid_ngrams += torch.tensor(bleu.counts, device=self.device).clone()
+        self.total_ngrams += torch.tensor(bleu.totals, device=self.device).clone()
+        self._steps += 1
+        if self._steps % 20 == 0:
```
artemru: 20 here is arbitrary (too often will create too much time overhead)

cbalioglu: Why do we still need to call GC if we have fixed the memory leak? Having said that I think 20 is still too low. In the main training loop, by default ,we call GC every 1000 step (which is configurable via recipe). I think anything less than ~400-500 might start to affect the runtime performance depending on the memory allocation pattern of the recipe

src/fairseq2/metrics/text/_bleu.py
```diff
@@ -65,14 +76,16 @@ def update(self, refs: Sequence[str], hyps: Sequence[str]) -> Self:
...
-        bleu = corpus_bleu(hyps, [refs], tokenize=self.tokenizer)
+        bleu = self._metric_class.corpus_score(hyps, [refs])
 
         self.sys_len += bleu.sys_len
         self.ref_len += bleu.ref_len
 
-        self.valid_ngrams += torch.tensor(bleu.counts, device=self.device)
-        self.total_ngrams += torch.tensor(bleu.totals, device=self.device)
-
+        self.valid_ngrams += torch.tensor(bleu.counts, device=self.device).clone()
```
cbalioglu: Why do we clone here an already newly initialized tensor? Seems redundant?

src/fairseq2/metrics/text/_bleu.py
```diff
@@ -57,6 +59,15 @@ def __init__(
 
         self._add_state("valid_ngrams", valid_ngrams)
         self._add_state("total_ngrams", total_ngrams)
+        self._metric_class = BLEU(
```
artemru, line 59: pls, check if I got all the params right

src/fairseq2/metrics/text/_bleu.py
```diff
@@ -16,6 +17,7 @@
 from torcheval.metrics import Metric
 from typing_extensions import Self, override
 
+from fairseq2.logging import log
```
artemru: rm

src/fairseq2/metrics/text/_bleu.py
```diff
@@ -65,14 +76,16 @@ def update(self, refs: Sequence[str], hyps: Sequence[str]) -> Self:
...
+        bleu = self._metric_class.corpus_score(hyps, [refs])
 
         self.sys_len += bleu.sys_len
         self.ref_len += bleu.ref_len
 
-        self.valid_ngrams += torch.tensor(bleu.counts, device=self.device)
-        self.total_ngrams += torch.tensor(bleu.totals, device=self.device)
-
+        self.valid_ngrams += torch.tensor(bleu.counts, device=self.device).clone()
+        self.total_ngrams += torch.tensor(bleu.totals, device=self.device).clone()
```
artemru: no needed

src/fairseq2/metrics/text/_bleu.py
```diff
@@ -65,14 +76,16 @@ def update(self, refs: Sequence[str], hyps: Sequence[str]) -> Self:
...
+        self._steps += 1
+        if self._steps % 500 == 0:
```
cbalioglu: Why do we still need to call GC here if we have fixed the memory leak? Having said that, we call GC every 1000 step in the trainer (which is configurable here: https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/recipes/config.py#L171-L172)

I think a metric implementation should not have this kind of "hidden" behavior that might impact the overall job throughput. Depending on the memory allocation pattern of the job, a GC can become quite expensive.

artemru: I can try without it and see if it helps or not!
do we have some gc call after each eval run ?

src/fairseq2/metrics/text/_bleu.py
artemru: already removed

src/fairseq2/metrics/text/_bleu.py
artemru: it runs ok without it

artemru: @cbalioglu : does `gc_every_n_steps` acts also for validation steps ?

artemru: actually, no - it's better with an internal gc.collect() ...

cbalioglu: We do a CUDA cache allocator clean-up, but not GC: https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/recipes/_trainer.py#L1167-L1168

We can also do GC there though.

> actually, no - it's better with an internal gc.collect()

Do you mean you still get a leak when you don't call gc.collect() even after the refactoring of the corpus call?

src/fairseq2/metrics/text/_bleu.py
gilkeren1: hmmm I personally can't really say

PR #1251, opened 2025-08-20
Multi split eval

artemru:
Adding a possibility to generate several eval datasets by dividing given split/corpus by an extra `readers_partition_columns` .


iamyuanchung (approved):
lgtm

PR #1261, opened 2025-08-23
startup fixes in v0.5.0a1

uralik:

* fixing environment resolver call (that links to cluster resolver)
* cluster resolver slurm variable name fix

### DIFF
--- a/src/fairseq2/assets/store.py
+++ b/src/fairseq2/assets/store.py
@@ -202,7 +202,7 @@ class AssetEnvironmentDetector:
 
     def detect(self) -> str | None:
         for env_resolver in self._env_resolvers:
-            env = env_resolver()
+            env = env_resolver(self._resolver)
             if env is not None:
                 return env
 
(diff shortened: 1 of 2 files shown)


src/fairseq2/cluster.py
```diff
@@ -177,7 +177,7 @@ def _get_master_port(self, job_id: int) -> str:
 
     @override
     def supports_current_cluster(self) -> bool:
-        return self._env.has("SLURM_PROC_ID")
+        return self._env.has("SLURM_PROCID")
```
zyaoj, line 180: thanks for the fix!

src/fairseq2/assets/store.py
```diff
@@ -202,7 +202,7 @@ def __init__(
 
     def detect(self) -> str | None:
         for env_resolver in self._env_resolvers:
-            env = env_resolver(self._resolver)
+            env = env_resolver()
```
zyaoj: This is placed here to follow the protocol `AssetEnvironmentResolver`. Please refer to the updated fs2-ext [here](https://github.com/fairinternal/fairseq2-ext/pull/132).

PR #1262, opened 2025-08-27
parallel vllm worker init, online dpo ref score fix when bs>1 and reward is binary

uralik:
as title says

tested with GRPO and online DPO
spinning up 16 vllm actors over 2 Ray nodes take couple mins now. thanks @a-paulus for first implementation and testing this out :) 

> online dpo ref score fix

our online dpo ref score was buggy when we filtered out some items from batch due to no preference signal. This got introduced during refactor in the first week of June and only show itself when we use math-verify (i.e. 0/1 reward alone) and batch size > 1. Otherwise (batch size = 1), entire batch loss is zeroed out anyway.

also we added _presets.py where we show one example of a preset done programmatically

### DIFF
--- a/src/fairseq2/recipes/lm/_online_finetune/_online_dpo.py
+++ b/src/fairseq2/recipes/lm/_online_finetune/_online_dpo.py
@@ -264,12 +264,18 @@ class OnlineDpoFinetuneUnit(TrainUnit[SequenceBatch]):
         )
         update_logit_entropy(metric_bag, tgt_logit_entropy)
 
+        prompt_lengths = (
+            prompt_batch.prompt_lengths
+            if is_bad_batch
+            else reward_output["prompt_lengths"]
+        )
+
         token_ref_chosen_logps = compute_reference_logps(
             self._gangs,
             self._reference_model,
             chosen_input_batch_seqs,
             chosen_input_batch_layout,
-            prompt_batch.prompt_lengths,
+            prompt_lengths,
         )
 
         token_ref_rejected_logps = compute_reference_logps(
@@ -277,7 +283,7 @@ class OnlineDpoFinetuneUnit(TrainUnit[SequenceBatch]):
             self._reference_model,
             rejected_input_batch_seqs,
             rejected_input_batch_layout,
-            prompt_batch.prompt_lengths,
+            prompt_lengths,
         )
 
         ref_average_chosen_logps = token_ref_chosen_logps.mean(dim=-1)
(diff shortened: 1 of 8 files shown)


src/fairseq2/recipes/lm/_online_finetune/_presets.py
```diff
@@ -0,0 +1,178 @@
...
+
+    @preset("qwen3_4b_wildchat_grpo")
+    def qwen3_4b_wildchat_grpo() -> OnlineFinetuneConfig:
+
+        model_config = ModelSection(name="qwen3_4b")
+        tokenizer_config = TextTokenizerSection(name="qwen3_4b")
+
+        dataset_config = OnlineFinetuneDatasetSection(
+            name="wildchat",
+            path="/checkpoint/ram/chenxwh/data/online_dpo/wildchat_1k",
```
jacklanchantin: do you want this path here?

src/fairseq2/recipes/lm/_online_finetune/_presets.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+
+        policy_vllm_config = VllmRayActorConfig(
+            ray_actor_name="vllm_policy",
+            backend="vllm",
+            num_replicas=2,
+            init_update_process_group=True,
+            vllm_engine_args=VllmEngineArgs(
+                model="/datasets/pretrained-llms/Qwen3-4B/",
+                tokenizer="/datasets/pretrained-llms/Qwen3-4B/",
```
jacklanchantin: also these paths?

src/fairseq2/recipes/lm/_online_finetune/_online_dpo.py
```diff
@@ -264,20 +264,26 @@ def __call__(
         )
         update_logit_entropy(metric_bag, tgt_logit_entropy)
 
+        prompt_lengths = (
+            prompt_batch.prompt_lengths
+            if is_bad_batch
+            else reward_output["prompt_lengths"]
+        )
```
jacklanchantin, line 272: what's the difference between `prompt_batch.prompt_lengths` and `reward_output["prompt_lengths"]`?

src/fairseq2/recipes/lm/_online_finetune/_presets.py
uralik: will clean up!

src/fairseq2/recipes/lm/_online_finetune/_presets.py
uralik: oh yes here we can use hg:// paths

src/fairseq2/recipes/lm/_online_finetune/_online_dpo.py
uralik: reward_output["prompt_lengths"] excludes lengths that correspond to examples that we filter out due to no preference signal. This was the issue that in such situation (when smth is filtered out) the compute ref score was using wrong prompt length for masking 

and if is_bad_batch is True, then entire batch is dummy, so we use original prompt lengths

jacklanchantin (approved):
tested with online DPO. looks good after fixing the hardcoded RAM paths

PR #1265, opened 2025-09-01
V0.5.0a1: fix checkpointing

avidale:
This change fixes the problem of saving checkpoints that started occurring after the major recipe refactoring (https://github.com/facebookresearch/fairseq2/commit/f2a7454e7a91287d1eb8b26873f7355cc6f014a6):
1. Trainer metrics were not serialized properly during checkpoint saving. 
To fix it, I just start calling `state_dict()` on each metric.
2. Checkpoint manager was trying to remove an checkpoint directory that does not exist and failing with a subsequent OS error.
To fix it, I am simply ignoring `FileNotFoundError` when removing the old checkpoint.
3.  The `_check_safe_state_dict` method of the checkpoint manager was raising errors every time it encountered a list or a dict.
To fix this apparent bug, I am just adding a missing return statement after inspecting all elements of a dict or list.
Additionally, to simplify debugging in the future, I am making the error message more informative by indicating the exact location of the non-compliant object.

None

**Check list:**


src/fairseq2/checkpoint/manager.py
```diff
@@ -433,18 +433,23 @@ def check_item(item: object) -> None:
...
                 return
+            
+            if path is None:
+                path = []
 
             if isinstance(item, Mapping):
                 for k, v in item.items():
-                    check_item(k)
-                    check_item(v)
+                    check_item(k, path=path + [str(k)])
```
artemru, line 442: will it be back or forward compatible ?

artemru (approved):
look good, but I'm not sure about the back compatibility (reloading older chkps)

src/fairseq2/checkpoint/manager.py
avidale: Yes, it is fully compatible; this particular change affects only the error message that gets printed in case if the checkpoint cannot be saved.

src/fairseq2/metrics/bag.py
```diff
@@ -93,7 +93,7 @@ def state_dict(self) -> dict[str, object]:
         state_dict: dict[str, object] = {}
 
         for name, metric in self._metrics.items():
-            state_dict[name] = metric
+            state_dict[name] = metric.state_dict()
```
avidale: Apparently, the information about the metric type gets lost here, so we'll need to save some more information in order to make it deserializeable.

src/fairseq2/checkpoint/manager.py
```diff
@@ -424,7 +424,7 @@ def commit() -> None:
                 raise OperationalError("A thread pool queue operation failed.") from ex
 
     def _check_safe_state_dict(self, kind: str, state_dict: dict[str, object]) -> None:
-        def check_item(item: object) -> None:
+        def check_item(item: object, path: Optional[List[str]] = None) -> None:
```
zyaoj: Could we apply `path: List[str] | None = None` ([PEP 604](https://peps.python.org/pep-0604/) for python>=3.10) or directly `path: List[str] = []` so that we can skip the L437-438 below?

src/fairseq2/metrics/bag.py
zyaoj: Great catch! Could we update to check whether the metric is `Stateful` first, similar to [this](https://github.com/facebookresearch/fairseq2/blob/v0.5.0a1/src/fairseq2/trainer.py#L1167-L1170)? For your case, `torcheval.metrics.Mean/Sum/...` are `Stateful`. But this might not apply to other cases.

```py
>>> from fairseq2.checkpoint import Stateful
>>> metric
<torcheval.metrics.aggregation.mean.Mean object at 0x74976b7d9960>
>>> isinstance(metric, Stateful)
True
```

zyaoj (changes requested):
Thanks for the fix! Great catch and thanks for enhancing the debug logging. Some nit changes requested.

src/fairseq2/metrics/bag.py
avidale: If the metric is not Stateful, why would we try saving it? The checkpoint manager type safety check will fail anyway.

src/fairseq2/checkpoint/manager.py
avidale: I was taught that mutable default argument values are a bad practice 🤷

zyaoj: yeah you are right you got a better teacher

zyaoj (approved):
LGTM!

PR #1270, opened 2025-09-04
Upstream MMS specific code

cirquit:

Upstreams mms's specific code that lived in main_w2v2_pretraining and can not be extracted to their repository (duplicate of #1256).

### DIFF
--- a/src/fairseq2/assets/cards/datasets/librilight.yaml
+++ b/src/fairseq2/assets/cards/datasets/librilight.yaml
@@ -5,5 +5,4 @@
 # LICENSE file in the root directory of this source tree.
 
 name: librilight_asr_10h
-#dataset_family: generic_asr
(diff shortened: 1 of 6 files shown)


src/fairseq2/assets/cards/datasets/librilight.yaml
```diff
@@ -5,5 +5,4 @@
 # LICENSE file in the root directory of this source tree.
 
 name: librilight_asr_10h
-#dataset_family: generic_asr
```
artemru, line 8: should we really keep it ?

cbalioglu: I consider this a temporary solution till we have the recipes ready. Ideally these asset YAMLs now should live within recipes.

PR #1271, opened 2025-09-04
Add SFT Recipe

jacklanchantin:

- Adds SFT recipe for masked input training.
- Adapts old instruction_finetune recipe for to fit within v0.5 `train` recipe format.

Tested 4 configs using named assets:
```
dataset=alpaca, model=llama_3_2_1b
dataset=alpaca, model=llama_3_2_1b_instruct, chat_mode=true
dataset=gsm8k, model=llama_3_2_1b
dataset=gsm8k, model=llama_3_2_1b_instruct, chat_mode=true
```

Run using:
```
python -m sft /output/dir/ --config-file sft/configs/llama3_2_1b_gsm8k.yaml
````

**Check list:**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to fairseq2 are documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [0.6.0] - TBD
+- The recipe command line interface now accepts a new `--no-exit-on-error` flag
+  to allow post-mortem debugging of recipe processes. [More info](https://github.com/facebookresearch/fairseq2/pull/1337).
 - The optimizer and learning rate scheduler recipe configurations now support
   multiple parameter groups. This is in particular convenient for models that
   require more than one learning rate to train (e.g. GAN models). Check out
--- a/src/fairseq2/recipe/cli.py
+++ b/src/fairseq2/recipe/cli.py
@@ -7,12 +7,18 @@
 from __future__ import annotations
 
 import sys
-from argparse import OPTIONAL, ArgumentError, ArgumentParser, Namespace
+from argparse import (
+    OPTIONAL,
+    ArgumentError,
+    ArgumentParser,
+    BooleanOptionalAction,
+    Namespace,
+)
 from collections.abc import Iterator, Mapping
 from contextlib import contextmanager
 from pathlib import Path
 from signal import SIG_DFL, SIGINT, raise_signal, signal
-from typing import Any, Protocol, TextIO, TypeVar, final, runtime_checkable
+from typing import Any, NoReturn, Protocol, TextIO, TypeVar, final, runtime_checkable
 
 import torch
 from torch.cuda import OutOfMemoryError
@@ -120,7 +126,7 @@ def train_main(recipe: TrainRecipe) -> None:
 
     container = DependencyContainer()
 
-    with _handle_errors(container):
+    with _handle_errors(container, args.exit_on_error):
         with _swap_default_resolver(container):
             _register_library(container)
 
@@ -141,7 +147,7 @@ def eval_main(recipe: EvalRecipe) -> None:
 
     container = DependencyContainer()
 
-    with _handle_errors(container):
+    with _handle_errors(container, args.exit_on_error):
         with _swap_default_resolver(container):
             _register_library(container)
 
@@ -162,7 +168,7 @@ def generate_main(recipe: GenerationRecipe) -> None:
 
     container = DependencyContainer()
 
-    with _handle_errors(container):
+    with _handle_errors(container, args.exit_on_error):
         with _swap_default_resolver(container):
             _register_library(container)
 
@@ -246,6 +252,13 @@ def _parse_args() -> Namespace:
         help="dump the configuration in mergeable format to standard output",
     )
 
+    parser.add_argument(
+        "--exit-on-error",
+        default=True,
+        action=BooleanOptionalAction,
+        help="whether to gracefully exit in case of an error",
+    )
+
     output_dir_action = parser.add_argument(
         "output_dir",
         type=Path,
@@ -264,7 +277,13 @@ def _parse_args() -> Namespace:
 
 
 @contextmanager
-def _handle_errors(resolver: DependencyResolver) -> Iterator[None]:
+def _handle_errors(resolver: DependencyResolver, exit_on_error: bool) -> Iterator[None]:
+    def maybe_exit(status: int) -> NoReturn:
+        if exit_on_error:
+            sys.exit(status)
+
+        raise
+
     try:
         yield
     except TaskStopException:
@@ -272,11 +291,11 @@ def _handle_errors(resolver: DependencyResolver) -> Iterator[None]:
     except RecipeConfigParseError:
         log.exception("Recipe configuration cannot be parsed. See logged stack trace for details.")  # fmt: skip
 
-        sys.exit(2)
+        maybe_exit(2)
     except ValidationError as ex:
         log.error(str(ex))
 
-        sys.exit(2)
+        maybe_exit(2)
     except RecipeError as ex:
         msg = str(ex)
 
@@ -285,25 +304,28 @@ def _handle_errors(resolver: DependencyResolver) -> Iterator[None]:
         else:
             log.exception(msg)
 
-        sys.exit(2)
+        maybe_exit(2)
     except OperationalError:
         log.exception("Recipe failed due to an operational error. See logged stack trace for details.")  # fmt: skip
 
-        sys.exit(1)
+        maybe_exit(1)
     except ExtensionError as ex:
         log.exception("{} extension failed to initialize. See logged stack trace for details.", ex.entry_point)  # fmt: skip
 
-        sys.exit(1)
+        maybe_exit(1)
     except OutOfMemoryError:
         s = torch.cuda.memory_summary()
 
         log.exception("CUDA out of memory. See logged memory stats.\n{}", s)  # fmt: skip
 
-        sys.exit(1)
+        maybe_exit(1)
     except KeyboardInterrupt:
-        signal(SIGINT, SIG_DFL)
+        if exit_on_error:
+            signal(SIGINT, SIG_DFL)
 
-        raise_signal(SIGINT)
+            raise_signal(SIGINT)
+
+        raise
     except Exception as ex:
         handler: ExceptionHandler[Any] | None
 
@@ -319,7 +341,7 @@ def _handle_errors(resolver: DependencyResolver) -> Iterator[None]:
         else:
             ret_code = handler(ex)
 
-        sys.exit(ret_code)
+        maybe_exit(ret_code)
 
 
 @final

### CODE REVIEW

File: recipes/lm/sft/default_config.py [hunk 1]
```diff
@@ -0,0 +1,161 @@
...
+            compile_options=CompileOptions(fullgraph=True, dynamic=False),
+        )
+    )
+
+    # tokenizer: TokenizerSection = field(
+    #     default_factory=lambda: TokenizerSection(
+    #         config_overrides=LLaMATokenizerConfig(impl="tiktoken", use_eot=True),
+    #         # config_overrides=LLaMATokenizerConfig(impl="hg")
+    #     )
+    # )
```
jacklanchantin: If we specify a local tokenizer path, we have to use impl=tiktoken. but maybe we want to just keep everything as hf?

recipes/lm/sft/configs/llama3_2_1b_instruct_alpaca.yaml
```diff
@@ -0,0 +1,32 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+dataset:
+  valid_split: "valid"
+  batch_size: 16
+  chat_mode: true
+  config_overrides:
+    path: "/checkpoint/ram/jacklanchantin/data/alpaca/"
```
jacklanchantin: need to remove these paths. how should we set default? new dataset card?

recipes/lm/sft/dataset.py
```diff
@@ -0,0 +1,374 @@
...
+    proportion to their weights.
+    """
+
+    source_encode_mode: str = "prompt"
+    """The tokenizer mode to encode the source text."""
+
+    target_encode_mode: str = "prompt_response"
+    """The tokenizer mode to encode the target text."""
+
+    chat_mode: bool = False
```
jacklanchantin: maybe we don't need this class anymore and it's all just in [Class DataReadOptions](https://github.com/facebookresearch/fairseq2/pull/1271/files#diff-fc9871ae1e62201b728f966210cf78f11734a2d773bc51133a3bc955e20a6f98R48)?

recipes/lm/sft/dataset.py
```diff
@@ -0,0 +1,374 @@
...
+
+LM_SFT_DATASET: Final = "lm_sft"
+
+
+@dataclass(kw_only=True)
+class DataReadOptions:
+    batching: Batching = field(default_factory=lambda: StaticBatching(1))
+    """The batching strategy for returned examples."""
+
+    example_shuffle_window: int = 1
```
uralik: lets do true shuffling by default?

recipes/lm/sft/dataset.py
uralik: yep I think this can all go there within standalone recipe now

recipes/lm/sft/config.py
uralik, line 1: shall this be just 'config.py' ?

recipes/lm/sft/default_config.py
uralik: yes, default impl=hf makes sense, since this is how we would expect usage with chat mode, + usage with other model like qwen

recipes/lm/sft/dataset.py
```diff
@@ -0,0 +1,374 @@
...
+    LengthBatching,
+    StaticBatching,
+    load_files_and_weights,
+)
+
+# TODO: FIX, INFER
+npc = 10
+
+
+LM_SFT_DATASET: Final = "lm_sft"
```
jacklanchantin: change to LM_SFT_PADDED_DATASET "lm_sft_padded"

recipes/lm/sft/dataset.py
```diff
@@ -0,0 +1,374 @@
...
+
+        for split, (files, weights) in splits.items():
+            if len(files) != len(weights):
+                raise ValueError(
+                    f"The lengths of the file and weight lists of the '{split}' split must match, but they are {len(files)} and {len(weights)} instead."
+                )
+
+        self._splits = splits
+
+    def _read_jsonl(self, path: Path, tokenizer: Tokenizer) -> DataPipelineBuilder:
```
jacklanchantin, line 145: change to how read_jsonl is done in train recipe

recipes/lm/sft/default_config.py
```diff
@@ -0,0 +1,160 @@
...
+    """If True, dataset jsonl must have 'chat' field with openai-like messages List[Dict] entries"""
+
+    min_seq_len: int = 1
+    """The minimum sequence length."""
+
+    max_seq_len: int = 8192
+    """The maximum sequence length."""
+
+    max_num_tokens: int = 8192 * 2
+    """The maximum number of tokens per batch."""
```
jacklanchantin: specify that this is excluding padding tokens

recipes/lm/sft/dataset.py
jacklanchantin: changed both of these:

```
example_shuffle_window: int = 0
batch_shuffle_window: int = 0
```

recipes/lm/sft/config.py
jacklanchantin: i was basing it off of this https://github.com/facebookresearch/fairseq2/tree/balioglu/wav2vec2_example/recipes/wav2vec2/asr

recipes/lm/sft/dataset.py
```diff
@@ -0,0 +1,371 @@
...
+
+        # file_world_size = gangs.dp.size
+
+        # if len(files) < file_world_size:
+        #     raise ValueError(
+        #         "The number of dataset files must be greater than or equal to the number of world size."
+        #     )
+
+        # if file_world_size > 1:
+        #     builder.shard(file_rank, file_world_size, allow_uneven=True)
```
jacklanchantin: do we need this? it was used in train recipe, but i don't think we need to enforce data chunks with sft

recipes/lm/sft/dataset.py
```diff
@@ -0,0 +1,378 @@
...
+from .utils import (
+    Batching,
+    DatasetLoadError,
+    LengthBatching,
+    StaticBatching,
+    load_files_and_weights,
+)
+
+# TODO: FIX, INFER
+npc = 10
```
uralik: we have npc in the config below, so this can be removed if we use npc from dataset config

recipes/lm/sft/dataset.py
uralik: yes we dont need it, the code
`builder.shard(gangs.dp.rank, gangs.dp.size, allow_uneven=True)` below executes the sharding logic across data workers

recipes/lm/sft/config.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+
+    max_num_valid_tokens: int | None = None
+    """The maximum number of tokens per validation batch."""
+
+    example_shuffle_window: int = 10_000
+    """The size of the sliding window for shuffling examples."""
+
+    batch_shuffle_window: int = 1000
+    """The size of the sliding window for shuffling batches."""
```
uralik, line 71: we have different defaults here comparing to dataset, lets also set these to 0?

recipes/lm/sft/default_config.py
```diff
@@ -0,0 +1,152 @@
...
+    """The encode mode for the target, determines what special tokens to add."""
+
+    chat_mode: bool = False
+    """If True, dataset jsonl must have 'chat' field with openai-like messages List[Dict] entries"""
+
+    min_seq_len: int = 1
+    """The minimum sequence length."""
+
+    max_seq_len: int = 8192
+    """The maximum sequence length."""
```
uralik:
```suggestion
    """The maximum sequence length. NOTE: longer sequences are dropped from the training."""
```

recipes/lm/sft/config.py
```diff
@@ -0,0 +1,152 @@
...
+    )
+
+    lr_scheduler: LRSchedulerSection | None = field(
+        default_factory=lambda: LRSchedulerSection(
+            name=COSINE_ANNEALING_LR,
+            config=CosineAnnealingLRConfig(
+                num_warmup_steps=2000, start_lr=1e-30, final_lr_scale=0.01
+            ),
+        ),
+    )
```
uralik, line 120: we didnt do such warmups in SFT before right? probably we should disable warmup by default?

recipes/lm/sft/config.py
```diff
@@ -0,0 +1,152 @@
...
+    regime: RegimeSection = field(
+        default_factory=lambda: RegimeSection(
+            num_steps=48_000,
+            validate_every_n_steps=4000,
+            checkpoint_every_n_steps=1000,
+            keep_last_n_checkpoints=3,
+            keep_checkpoint_every_n_steps=4000,
+            publish_metrics_every_n_steps=10,
+        )
+    )
```
uralik, line 133: lets set args that we typically use? to save HF model, and also this new arg that allows to keep only the last step full train state, but not for older ckpts

recipes/lm/sft/default_config.py
```diff
@@ -0,0 +1,152 @@
...
+
+@dataclass(kw_only=True)
+class InstructionFinetuneDatasetSection(DatasetSection):
+    # name: str = "foo"  # FIXME: change!
+
+    # family: str = GENERIC_INSTRUCTION_DATASET_FAMILY
+
+    path: str | None = None
+
+    train_split: str = "train"  # FIXME (jacklanchantin)not sure what this should be
```
uralik: can remove comments now and we can set default names here for gsm8k dataset?

recipes/lm/sft/default_config.py
```diff
@@ -0,0 +1,152 @@
...
+    ModelSection,
+    OptimizerSection,
+    RegimeSection,
+    TokenizerSection,
+    TorchConfig,
+    TrainerSection,
+)
+from .dataset import LM_SFT_PADDED_DATASET, LMSFTDatasetConfig
+
+GENERIC_INSTRUCTION_DATASET_FAMILY: Final = "generic_instruction"
```
uralik: this is not used right? we have other family name in the dataset.py

uralik: the LM_SFT_PADDED_DATASET

recipes/lm/sft/recipe.py
```diff
@@ -0,0 +1,227 @@
...
+
+        unit = LMSFTUnit(context.model)
+
+        dataset = context.default_dataset.as_(LMSFTDataset)
+
+        seed = config.common.seed
+        device = get_default_device()
+
+        if device.type == "cuda":
+            torch.cuda.set_device(device)
```
cbalioglu: This is not really needed. fairseq2 already sets the device before calling `create_trainer`. Any particular reason for calling this here?

recipes/lm/sft/recipe.py
```diff
@@ -0,0 +1,227 @@
...
+
+        rng_bag = RngBag.from_device_defaults(CPU, device)
+
+        rng_bag.manual_seed(seed)
```
cbalioglu: This is also not needed. fairseq2 calls `manual_seed(config.common.seed)` before calling `create_trainer`.

recipes/lm/sft/recipe.py
```diff
@@ -0,0 +1,227 @@
...
+    @override
+    def create_trainer(self, context: RecipeContext) -> Trainer:
+        config = context.config.as_(LMSFTConfig)
+
+        unit = LMSFTUnit(context.model)
+
+        dataset = context.default_dataset.as_(LMSFTDataset)
+
+        seed = config.common.seed
+        device = get_default_device()
```
cbalioglu: I think you don't need `device` (see my comments below), but in case needed, a cleaner way is `context.gangs.root.device`.

recipes/lm/sft/default_config.py
```diff
@@ -0,0 +1,145 @@
...
+    TokenizerSection,
+    TorchConfig,
+    TrainerSection,
+)
+
+from .dataset import LM_SFT_PADDED_DATASET, LMSFTDatasetConfig
+
+
+@dataclass(kw_only=True)
+class InstructionFinetuneDatasetSection(DatasetSection):
```
cbalioglu: Should we name this `LMSFTDatasetSection`?

cbalioglu (approved):
Left a few mostly nit comments. Overall looks good to me. Thanks a lot for updating this recipe!

I think one follow-up improvement we should make is revising the dataset implementation to a packed variant. This is on me (or one of the fs2 team members) though.

I am stamping the PR, but would be great if you could check/address the few comments I left.

recipes/lm/sft/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+```bash
+# Run with a pre-configured Llama setup
+python -m fairseq2.recipes.lm.sft --config-name recipes/lm/sft/configs/llama3_2_1b_gsm8k.yaml
+
+# Run with a pre-configured Qwen setup
+python -m fairseq2.recipes.lm.sft --config-name recipes/lm/sft/configs/qwen3_4b_gsm8k.yaml
+
+# Run with custom config file")
+python -m fairseq2.recipes.lm.sft --config-file path/to/your/config.yaml
```
cbalioglu: I think you don't need "fairseq2" here (and in other examples) right? It ideally should work with

```
python -m recipes.lm.sft ...
```

src/fairseq2/models/llama/tokenizer.py
```diff
@@ -191,7 +191,7 @@ def vocab_info(self) -> VocabularyInfo:
 
 @dataclass
 class LLaMATokenizerConfig:
-    impl: Literal["sp", "tiktoken", "hg"] = "sp"
+    impl: Literal["sp", "tiktoken", "hg"] = "hg"
```
cbalioglu: Just catching this up. Do you mind keeping this as `sp` so that we don't inadvertently break some dependent libraries? I think as long as we set "hg" as a "model_config" in the base LLaMA-3 asset card, this is not really needed anyways.

src/fairseq2/data/__init__.py
```diff
@@ -0,0 +1,7 @@
+from fairseq2.data.data_pipeline import CollateOptionsOverride as CollateOptionsOverride
+from fairseq2.data.data_pipeline import Collater as Collater
+from fairseq2.data.data_pipeline import DataPipelineBuilder as DataPipelineBuilder
+from fairseq2.data.data_pipeline import SequenceData as SequenceData
+from fairseq2.data.data_pipeline import create_bucket_sizes as create_bucket_sizes
+
+__all__: list[str] = []
```
cbalioglu: Nit: no need for `__all__` here. This can be removed.

PR #1280, opened 2025-09-05
Respect content-disposition while downloading assets

cbalioglu:
This PR improves filename handling when downloading files via redirection. In particular Hugging Face Hub recently started returning the actual filename in `Content-Disposition` header instead of within the URL which causes checkpoint loading to fail.


cirquit (approved):
LGTM!

PR #1282, opened 2025-09-07
Improve CLI error messaging for unknown assets

cbalioglu:
Per user feedback, a small PR that slightly improves the error message for unknown assets. This should be further improved once we have the proper documentation in place.


cbalioglu:
> Thanks for improving UX!

Thanks for the feedback!

PR #1283, opened 2025-09-07
Remove safetensors import check

cbalioglu:
A quick follow-up to PR #1281. Since safetensors is now a regular dependency of fairseq2, we do not need to have an import check for it anymore.


zyaoj (approved):
Good catch!

PR #1285, opened 2025-09-08
[documentation] update doc for v0.5

zyaoj:

- Add basic doc on `assets`
- Bring back (updated) installation instructions
- Bring back (updated) add model tutorial + pudb tutorial
- Introduce `doctest`
- Update some API references

N/A

**Check list:**


doc/source/basics/building_recipes.rst
```diff
@@ -370,11 +370,13 @@ Once you've created these four files, running your recipe is straightforward:
         --config-file /path/to/config.yaml \
         --config model.name=llama3_2_1b_instruct regime.num_steps=20 lr_scheduler.config.num_warmup_steps=10
 
+You can also specify the asset store to use with the config override ``--config common.asset.extra_paths=/path/to/asset/yaml/dir`` option.
```
cbalioglu: `extra_paths` is not a list. So the correct CLI syntax would be:

```
--config common.asset.extra_paths="['/path/to/assets/dir', '/path/to/yet_other_assets/dir']"
```

doc/source/basics/runtime_extension.rst
```diff
@@ -0,0 +1,123 @@
...
+
+* Register new models and model cards
+* Add custom asset providers and validators
+* Extend the runtime context
+* And more...
+
+Basic Usage
+-----------
+
+Before using any fairseq2 APIs, you must initialize the framework with :meth:`fairseq2.init_fairseq2`:
```
cbalioglu: This is not really needed anymore. Internal APIs call `init_fairseq2` if it has not been called so far. Calling `init_fairseq2` explicitly is mainly useful if you have a standalone non-recipe Python script where you mainly use fairseq2 and want to extend its dependency graph with some extra objects. Then at the beginning of your script you can do something like:

```
def register_my_script_objects(container: DependencyContainer) -> None:
   container.register(...)

if __name__ == "__main__":
  init_fairseq2(extras=register_my_script_objects)

  ...
```

Otherwise, as a user you can mostly forget about `init_fairseq2`.

doc/source/basics/runtime_extension.rst
```diff
@@ -0,0 +1,123 @@
...
+    from fairseq2.composition import register_package_assets
+    from fairseq2.runtime.dependency import DependencyContainer
+
+
+    def setup_my_extension(container: DependencyContainer) -> None:
+
+        # Register package metadata provider for cards
+        register_package_assets(container, package="my_package.cards")
+
+        def resolve_cluster_name(resolver: DependencyResolver) -> str:
```
cbalioglu: I think as an example, it would be nicer to show a simpler concept like simply registering a model like:

```
register_model_family(container, name=xyz, factory=create_my_model, ...)
```

I think for most users cluster name resolution is a fairly edge/advanced case. Even in FAIR, this is handled by fairseq2-ext, so most users don't care about how clusters are determined.

doc/source/basics/runtime_extension.rst
```diff
@@ -0,0 +1,123 @@
...
+
+
+Error Handling
+--------------
+
+The extension system includes error handling to maintain system stability:
+
+* Failed extensions log warnings by default
+* Set ``FAIRSEQ2_EXTENSION_TRACE`` environment variable for detailed error traces
+* Invalid extension functions raise ``RuntimeError``
```
cbalioglu: Nit, they raise a `fairseq2.composition.ExtensionError`.

doc/source/reference/api/models/index.rst
```diff
@@ -0,0 +1,77 @@
...
+    >>> print(f"Architecture: {card.field('model_arch').as_(str)}")
+    Architecture: qwen25_7b
+
+Supported Model Families
+-------------------------
+
+Language Models
+~~~~~~~~~~~~~~~
+
+- **Qwen**: Qwen and Qwen2.5 language models
```
cbalioglu: Qwen 2.5 and Qwen 3

doc/source/reference/api/models/index.rst
```diff
@@ -0,0 +1,77 @@
...
+- **LLaMA**: LLaMA and LLaMA 2 language models
```
cbalioglu: LLaMA 3 as well.

doc/source/reference/api/models/index.rst
```diff
@@ -0,0 +1,77 @@
...
+
+    >>> from fairseq2.assets import get_asset_store
+    >>> asset_store = get_asset_store()
+    >>> card = asset_store.retrieve_card("qwen25_7b")
+    >>> print(f"Model family: {card.field('model_family').as_(str)}")
+    Model family: qwen
+    >>> print(f"Architecture: {card.field('model_arch').as_(str)}")
+    Architecture: qwen25_7b
+
+Supported Model Families
```
cbalioglu, line 63: Maybe we can skip this section entirely. This requires us to keep maintaining this list separately. We can simply advise users to check `src/fairseq2/models` directory and/or `python -m fairseq2.assets list --kind models` to see the list of available models.

doc/source/tutorials/add_model.rst
```diff
@@ -0,0 +1,327 @@
...
+
+Step-by-Step Guide: Adding a Model to Existing Family
+-----------------------------------------------------
+
+Let's walk through adding ``qwen25_3b_instruct`` to the existing Qwen family.
+
+Step 1: Add Model Architecture Configuration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Model architectures are defined as configuration presets that specify the structural parameters of your model.
```
cbalioglu, line 89: If we are aiming for external users, I believe it would make more sense to show how to do this without needing to touch fairseq2 code base.

```python
def setup_my_fairseq2_extension(container: DependencyContainer) -> None:
  arch = ConfigRegistrar(container, QwenConfig)

  @arch("my_qwen_arch")
  def my_qwen_arch() -> QwenConfig:
     config = QwenConfig()
     ...
     return config
```

Either register `setup_my_fairseq2_extension` if this code resides in a Python package, or use `init_fairseq2(extras=setup_my_fairseq2_extension)` if this lives inside a plain Python script.

doc/source/tutorials/add_model.rst
```diff
@@ -0,0 +1,327 @@
...
+   - The function name should match or describe the architecture
+   - Use the appropriate config class for the model family (``QwenConfig`` for Qwen models)
+   - Set all necessary parameters for your specific model variant
+
+Step 2: Create Asset Card
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Asset cards are YAML files that tell fairseq2 where to find your model checkpoints and how to load them.
+
+1. **Navigate to the model family's asset card file:**
```
cbalioglu, line 125: Similar to above, ideally we should show how users can do this without touching fairseq2:

```python
def setup_my_fairseq2_extension(container: DependencyContainer) -> None:
  register_package_assets("my_package.assets")
  # or
  # register_file_assets("path/to/assets")
```

And put the asset card to the corresponding location.

doc/source/tutorials/add_model.rst
```diff
@@ -0,0 +1,327 @@
...
+2. **Test model loading:**
+
+   .. code-block:: python
+
+       import fairseq2
+       from fairseq2.models.hub import load_model
+
+       # Test loading your model
+       try:
+           model = load_model("qwen25_3b_instruct")
```
cbalioglu, line 178: Not sure if we should mention it here, but we also have "model hubs" for each model family that exposes methods (like load from path, create a newly initialized, list architectures so on) beyond loading. Maybe makes sense to refer to that in another part of the documentation though.

doc/source/tutorials/add_model.rst
```diff
@@ -0,0 +1,327 @@
...
+
+.. code-block:: yaml
+
+    checkpoint: "https://example.com/models/my_model.pt"
+
+
+Common Model Parameters
+-----------------------
+
+When creating new architecture configurations, here are the most common parameters you'll need to set:
```
cbalioglu: We should mention that these parameter names are "conventions" used by most of our models, but not necessarily standard to all models (e.g. wav2vec2 does not have a `vocab_size` or NLLB does not have `num_key_value_heads`).

cbalioglu (commented):
Thanks for last minute improvements to the documentation! Left a few comments, mostly nit ones.

doc/source/tutorials/add_model.rst
zyaoj: Good point! I'll add a short section here and create a bit more detailed doc in `/referece/api/modesl/hub.rst`

zyaoj: Also added a `/reference/api/models/qwen.rst` since teams might be working on that shortly.

doc/source/basics/assets.rst
```diff
@@ -0,0 +1,484 @@
...
+    tokenizer: "hg://meta-llama/Llama-3-8b"
+    tokenizer_family: llama
+
+**Required Fields for Models:**
+
+- ``name``: Unique identifier for the model
+- ``model_family``: The model family (e.g., 'llama', 'qwen', 'mistral')
+- ``model_arch``: The model architecture configuration to use
+- ``checkpoint``: Path or URI to the model checkpoint
+- ``tokenizer``: Path or URI to the tokenizer
```
cbalioglu: Nit. I would not say anything beyond `name`, `model_family`, and `checkpoint` are required to define model. In particular, `tokenizer` is specific only to text-based models.

doc/source/basics/assets.rst
```diff
@@ -0,0 +1,484 @@
...
+
+Environment-Specific Assets
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Assets can have environment-specific configurations using the ``@environment`` syntax:
+
+.. code-block:: yaml
+
+    # Base configuration
+    name: my_model@
```
cbalioglu: Nit. For the base card the "@" at the end is optional while writing asset YAMLs. If it is not there in the YAML file, fairseq2 internally appends it.

doc/source/basics/assets.rst
```diff
@@ -0,0 +1,484 @@
...
+   :widths: 20 40 40
+
+   * - Store Type
+     - Default Path
+     - Environment Variable
+   * - Built-in
+     - ``fairseq2/assets/cards/``
+     - N/A (package resources)
+   * - System
+     - ``/etc/fairseq2/assets/``
```
cbalioglu: Again, nit, but a small mistake: `FAIRSEQ2_ASSET_DIR` overrides `/etc/fairseq2/assets` (similarly the user variant as well)

doc/source/basics/assets.rst
```diff
@@ -0,0 +1,484 @@
...
+     - ``FAIRSEQ2_ASSET_DIR``
+   * - User
+     - ``~/.config/fairseq2/assets/``
+     - ``FAIRSEQ2_USER_ASSET_DIR``
+   * - :ref:`basics-runtime-extension`
+     - e.g. ``my_package/cards/``
+     - N/A (package resources)
+
+
+If you are working with recipe, you can also specify the asset store to use with the config override ``--config common.asset.extra_paths=/path/to/asset/yaml/dir`` option.
```
cbalioglu: As I mentioned in the first review, `extra_paths` is now a list instead of a string.

doc/source/basics/assets.rst
```diff
@@ -0,0 +1,484 @@
...
+.. code-block:: python
+
+    from pathlib import Path
+    from fairseq2.assets import (
+        get_asset_store,
+        StandardFileAssetMetadataLoader,
+        StandardAssetStore
+    )
+
+    def register_custom_assets():
```
cbalioglu: Unfortunately this is not the right way and will likely cause issues. The correct way to add new assets is via `register_file_assets()`.

```
from fairseq2.composition import register_file_assets

def setup_my_fairseq2(container: DependencyContainer) -> None:
  register_file_assets("/path/to/my/assets")

init_fairseq2(extras=setup_my_fairseq2)

# or register via setuptools entry_point.
# or via `Recipe.register()`
```

doc/source/basics/assets.rst
```diff
@@ -0,0 +1,484 @@
...
+        "tokenizer_family": "llama",
+    }
+
+    # Load it into the asset system
+    provider = load_in_memory_asset_metadata("dynamic_source", [card_info])
+
+    # Register with the asset store
+    asset_store = get_asset_store()
+    if isinstance(asset_store, StandardAssetStore):
+        asset_store._metadata_providers.append(provider)
```
cbalioglu: Likewise, this is going to fail in subtle ways because AssetStore treats its providers as immutable after being constructed and caches some information. I also realized that we miss a helper utility here for this. Let me quickly send a PR for that. Eventually this will look like below:

```
def setup_my_fairseq2(container: DependencyContainer) -> None:
  register_in_memory_assets(container, "my_in_mem_source", [...])
```

doc/source/basics/assets.rst
```diff
@@ -0,0 +1,484 @@
...
+   * - ``tokenizer``
+     - Tokenizer location
+     - Models only
+   * - ``tokenizer_family``
+     - Tokenizer family name
+     - Models only
+   * - ``dataset_family``
+     - Dataset loader family (for datasets)
+     - Datasets only
+   * - ``dataset_config.data``
```
cbalioglu: This is not true anymore. We now expect the dataset implementation to extra any relevant paths from its own config.

doc/source/reference/api/models/qwen.rst
```diff
@@ -0,0 +1,347 @@
...
+        config.num_layers = 32
+        config.num_attn_heads = 32
+        config.max_seq_len = 16384
+
+        # Or get pre-defined architecture
+        from fairseq2.models.qwen import get_qwen_model_hub
+        hub = get_qwen_model_hub()
+        config = hub.get_arch_config("qwen25_7b")
+
+register_qwen_configs
```
cbalioglu: Nit, I don't think it is really necessary to document this function. It is mostly internal plumbing and very unlikely to be used directly by the users.

doc/source/basics/assets.rst
zyaoj: I remember that I removed it at some point haha. Maybe I forgot to save.

doc/source/basics/assets.rst
zyaoj: Forgot to fix it here! Thanks for catching this!

doc/source/basics/assets.rst
zyaoj: Sounds good! Updated to `register_file_assets(container, Path("/path/to/my/assets"))`.

doc/source/basics/assets.rst
zyaoj: Yeah I wanted to indicate in this table that the default value will be override by the env var. I'll make it more explicit.

doc/source/basics/assets.rst
zyaoj: Oh I was under the impression that we were updating the mms dataset cards into this ways. Could you elaborate a bit more? Is this [the field](https://github.com/facebookresearch/fairseq2/blob/balioglu/wav2vec2_example/recipes/wav2vec2/ssl/data/dataset.py#L48-L49) that will be picked up as the extra path?

cbalioglu: I actually fixed in the wav2vec2_example branch for Alexander.  That `data` field is removed and instead the same information is now included here:

```
https://github.com/facebookresearch/fairseq2/blob/balioglu/wav2vec2_example/recipes/wav2vec2/asr/data/dataset.py#L416-L428
```

doc/source/basics/assets.rst
zyaoj: This has been updated per #1288

PR #1298, opened 2025-09-10
Introduce a new 'all_but_last' option for `regime.save_model_only`

cbalioglu:
This PR extends the `regime.save_model_only` training recipe config with an additional `all_but_last` option. When set to `all_but_last`, only the last checkpoint will have its full state saved while all previous checkpoints will contain only their model state. This is helpful if the user does not plan to branch off the training from a previous checkpoint since it avoids unnecessary disk space use for trainer, optimizer, and data reader states.

Tested in all three settings (`False`, `True`/`all`, `all_but_last`) with LM pretraining and wav2vec2 ASR finetuning recipes including with and without Hugging Face export.


src/fairseq2/recipe/trainer.py
```diff
@@ -1004,6 +1012,18 @@ def _maybe_complete_checkpoint_save_operation(self) -> None:
...
+        step_nrs = self._checkpoint_manager.get_step_numbers()
+        if len(step_nrs) <= 1:
+            return
+
+        log.info("Deleting non-model state of previous checkpoints.")
+
+        for step_nr in step_nrs[:-1]:  # always keep the last checkpoint.
+            self._checkpoint_manager.delete_checkpoint(step_nr, keep_model=True)
+
+        log.info("Non-model state of previous checkpoints deleted.")
```
artemru, line 1027: maybe include len(step_nrs) info here

src/fairseq2/recipe/config.py
```diff
@@ -345,7 +345,23 @@ class RegimeSection(Validatable):
...
     """The data epoch interval at which to checkpoint."""
 
-    save_model_only: bool = False
+    save_model_only: bool | Literal["all", "all_but_last"] = False
+    """
+    If ``False``, the full state of the training job is saved, including the
+    trainer, model, optimizer, and data reader states. This is the conventional
+    checkpointing behavior.
+
+    If ``True`` or ``all``, only the model state is saved during checkpointing.
```
artemru:
```suggestion
    If ``True`` or ``"all"``, only the model state is saved during checkpointing.
```

src/fairseq2/recipe/config.py
```diff
@@ -345,7 +345,23 @@ class RegimeSection(Validatable):
...
+    This is beneficial for short-lived training jobs where the user does not
+    expect to resume the job but requires frequent snapshots of the model for
+    evaluation purposes. In this mode, checkpointing is faster and disk space is
+    saved by avoiding the storage of trainer, optimizer, and data reader states.
+
+    If ``all_but_last``, only the last checkpoint will have its full state saved,
```
artemru:
```suggestion
    If ``"all_but_last"``, only the last checkpoint will have its full state saved,
```

src/fairseq2/recipe/trainer.py
cbalioglu: Thanks! Addressed in the new commit along with docstr feedback.

PR #1299, opened 2025-09-10
Relax default 'resume_mode' value for wandb

cbalioglu:
This PR (1) "relaxes" the default value of the `common.metric_recorders.wandb.resume_mode` from `allow` to `None`, so that preempted jobs do not generate noisy warnings when the existing metric records are overwritten. (2) Adds docstrs to Weights & Bias configuration.

### DIFF
--- a/src/fairseq2/recipe/config.py
+++ b/src/fairseq2/recipe/config.py
@@ -583,14 +583,51 @@ WandbResumeMode: TypeAlias = Literal["allow", "never", "must", "auto"]
 
 @dataclass(kw_only=True)
 class WandbConfig:
+    """
+    Holds the configuration of the Weights & Biases metric recorder.
+
+    For more information, check out the official Weights & Biases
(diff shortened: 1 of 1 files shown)


src/fairseq2/recipe/config.py
```diff
@@ -583,14 +583,51 @@ class TensorBoardConfig:
...
     group: str | None = None
+    """
+    The group name to organize individiual jobs as part of a larger experiment.
+    """
+
     job_type: str | None = None
-    resume_mode: WandbResumeMode = "allow"
+    """The type of the job such as "train", "eval"."""
+
+    resume_mode: WandbResumeMode | None = None
```
artemru, line 626: maybe adding a simple example in doc string

cbalioglu: All recipe configuration settings will have a detailed documentation update very soon as part of another PR.

PR #1305, opened 2025-09-15
Llama 4 Scout and Maverick text support

MartinGleize:
This PR adds Llama 4 Scout and Maverick support to fairseq2.

Test plan:

No regression on Llama 3 inference recipe
Llama 4 Scout parity: verified against llama-stack on gsm8k test.
Llama 4 Scout finetuning: recipe should work but no guarantee on effectiveness (token choice routing likely needs a load balancing loss, this will be another PR).

TODO in fs2 0.5.*:
test Llama 4 Scout native inference recipe support

No.

**Check list:**

### DIFF
--- a/src/fairseq2/models/llama4/config.py
+++ b/src/fairseq2/models/llama4/config.py
@@ -1,0 +1,153 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Final
+
+from fairseq2.models.llama import LLaMAConfig
+from fairseq2.runtime.config_registry import ConfigRegistrar
+from fairseq2.runtime.dependency import DependencyContainer
+
+LLAMA4_FAMILY: Final = "llama4"
+
+
+# Not used atm
+@dataclass(kw_only=True)
+class LLaMA4VisionEncoderConfig:
+    """Holds the configuration of a Llama 4 Vision Encoder"""
+
+    # TODO: Add descriptions to all these
+
+    image_size: int = 336
+
+    patch_size: int = 14
+
+    model_dim: int = 1408
+
+    num_layers: int = 34
+
+    num_attn_heads: int = 16
+
+    mlp_ratio: float = 4.0
+
+    output_dim: int = 4096
+
+    pixel_shuffle_ratio: float = 0.5
+
+
+@dataclass(kw_only=True)
+class LLaMA4ExpertsConfig:
+    """Holds the configuration of a Llama 4 Experts"""
+
+    num_experts: int = 16
+    """The number of MoE experts."""
+
+    use_shared_expert: bool = True
+    """If ``True``, an additional single expert is used for all tokens."""
+
+    capacity_factor: float = 1.0
+    """The capacity factor of experts."""
+
+    auto_scale: float = True
+
+    top_k: int = 1
+    """MoE sends each token to the ``top_k`` top experts."""
+
+    interleave_moe_layer_step: int = 1
+    """Llama will use a MoE layer as FFN every ``interleave_moe_layer_step``th layer.
+    If equal to 1, a MoE is used for every layer."""
+
+    eval_with_saved_stats: bool = False
+
+    expert_act_threshold: float = 0.0
+
+
+@dataclass(kw_only=True)
+class Llama4MetapConfig:
+    """Llama 4 optional meta parameters"""
+
+    base_width: float = 1024.0
+    """The base width used in the output multiplier computation."""
+
+    embedding_multiplier: float = 1.0
+    """
+    The embedding multiplier. Not implemented currently,
+    since the multiplier is 1.0 in all available checkpoints.
+    """
+
+
+@dataclass(kw_only=True)
+class Llama4Config(LLaMAConfig):
+    experts: LLaMA4ExpertsConfig = field(default_factory=lambda: LLaMA4ExpertsConfig())
+    """If not ``None``, specifies the configuration of Mixture-of-Experts."""
+
+    vision_config: LLaMA4VisionEncoderConfig | None = None
+    """If not ``None``, specifies the configuration of the vision encoder."""
+
+    metap_config: Llama4MetapConfig | None = None
+    """If not ``None``, specifies the configuration of the metaparameters."""
+
+    attention_chunk_size: int = 8192
+    """The chunk size used for chunked attention biases."""
+    
+    use_qk_norm: bool = False
+    """If ``True``, applies layer normalization to the projected query and key."""
+
+    nope_layer_interval: int | None = None
+    """
+    If not ``None``, will use a NoPE layer (no positional embedding)
+    instead of a RoPE layer every ``nope_layer_interval`` layers.
+    """
+
+
+def register_llama4_configs(container: DependencyContainer) -> None:
+    arch = ConfigRegistrar(container, Llama4Config)
+
+    @arch("llama4_scout")
+    def llama4_scout() -> Llama4Config
+        config = Llama4Config()
+
+        config.model_dim = 5120
+        config.num_layers = 48
+        config.num_attn_heads = 40
+        config.num_key_value_heads = 8
+
+        config.vocab_size = 202_048
+        config.pad_idx = 200_018
+
+        config.ffn_inner_dim = config.model_dim * 4
+        config.ffn_inner_dim_multiplier = 1.2
+        config.ffn_inner_dim_multiple_of = 2048
+
+        config.use_qk_norm = True
+
+        config.experts = LLaMA4ExpertsConfig()
+        
+        # vision has not been tested yet
+        # config.vision_config = LLaMA4VisionEncoderConfig()
+
+        config.rope_theta = 500_000.0
+        config.use_scaled_rope = True
+        config.nope_layer_interval = 4
+        config.rope_scale.factor = 16.0
+        config.rope_scale.frequency_factors = (1.0, 1.0)
+
+        return config
+
+    @arch("llama4_maverick")
+    def llama4_maverick() -> Llama4Config:
+        config = llama4_scout()
+
+        config.experts.num_experts = 128
+        config.experts.interleave_moe_layer_step = 2
+
+        config.use_qk_norm = False
+
+        config.use_scaled_rope = False
+
+        return config
--- a/src/fairseq2/models/llama4/factory.py
+++ b/src/fairseq2/models/llama4/factory.py
@@ -1,0 +1,212 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from __future__ import annotations
+
+import math
+from typing_extensions import override
+
+import torch
+import torch.nn as nn
+from torch import Tensor
+
+from fairseq2.models.llama import LLaMAFactory
+from fairseq2.models.llama.factory import _init_truncated_normal
+from fairseq2.models.llama4.model.frontend import LLaMA4DecoderFrontend
+from fairseq2.models.llama4.model.moe.moe import MoE
+from fairseq2.models.llama4.config import Llama4Config
+from fairseq2.models.llama4.model.vision.embedding import VisionEmbeddings
+from fairseq2.models.transformer import (
+    CausalAttentionBias,
+    ChunkedAttentionBias,
+    FeedForwardNetwork,
+    GLUFeedForwardNetwork,
+    MultiheadAttention,
+    StandardMultiheadAttention,
+    TransformerEmbeddingFrontend,
+    TransformerFrontend,
+    TransformerNormOrder,
+    create_default_sdpa,
+)
+from fairseq2.models.transformer_lm import (
+    StandardTransformerLMDecoder,
+    StandardTransformerLMDecoderLayer,
+    TransformerLM,
+    TransformerLMDecoder,
+    TransformerLMDecoderLayer,
+)
+from fairseq2.nn import (
+    Embedding,
+    LayerNorm,
+    Linear,
+    PositionEncoder,
+    Projection,
+    RMSNorm,
+    RotaryEncoder,
+    StandardEmbedding,
+    TiedProjection,
+)
+
+
+def create_llama4_model(config: Llama4Config) -> TransformerLM:
+    return LLaMA4Factory(config).create_model()
+
+
+class LLaMA4Factory(LLaMAFactory):
+    def __init__(self, config: Llama4Config) -> None:
+        self._config = config
+
+    @override
+    def create_decoder_frontend(self, embed: Embedding) -> TransformerFrontend:
+        config = self._config
+
+        if config.vision_config:
+            vision_embed = VisionEmbeddings(config.vision_config)
+            vision_proj = Linear(
+                config.vision_config.output_dim,
+                config.model_dim,
+                bias=False,
+                init_fn=lambda x: None,
+            )
+        else:
+            vision_embed = None
+            vision_proj = None
+
+        return LLaMA4DecoderFrontend(
+            embed,
+            vision_embed,
+            vision_proj,
+            pos_encoder=None,
+            no_scale=True,
+            dropout_p=config.dropout_p,
+        )
+    
+    def create_self_attention(
+        self, layer_idx: int, pos_encoder: PositionEncoder
+    ) -> MultiheadAttention:
+        """
+        Compared to Llama 1-3, Llama 4 interleaves layers with local chunked attention,
+        the only ones to actually use RoPE. Other layers use NoPE.
+        """
+        config = self._config
+        
+        is_nope_layer = self._is_nope_layer(layer_idx)
+        use_local_attn_mask = not is_nope_layer
+        
+        if use_local_attn_mask:
+            attn_bias = ChunkedAttentionBias(
+                attn_chunk_size=config.attention_chunk_size
+            )
+        else:
+            attn_bias = CausalAttentionBias()
+
+        sdpa = create_default_sdpa(attn_bias)
+        
+        pos_encoder_maybe = pos_encoder if not is_nope_layer else None
+        
+        qk_norm = None
+        if config.use_qk_norm and not is_nope_layer:
+            head_dim = config.model_dim // config.num_attn_heads
+            qk_norm = self.create_qk_norm(head_dim)
+
+        init_std = config.init_std
+
+        std_scale_factor = self.get_std_scale_factor(layer_idx)
+
+        def init_projection(proj: Linear) -> None:
+            input_dim = proj.weight.shape[1]
+
+            std = init_std or (input_dim**-0.5)
+
+            _init_truncated_normal(proj.weight, proj.bias, std=std / std_scale_factor)
+
+        return StandardMultiheadAttention(
+            config.model_dim,
+            config.num_attn_heads,
+            sdpa,
+            num_key_value_heads=config.num_key_value_heads,
+            qkv_proj_init_fn=init_projection,
+            pos_encoder=pos_encoder_maybe,
+            output_proj_init_fn=init_projection,
+            bias=False,
+            q_norm=qk_norm,
+            k_norm=qk_norm,
+        )
+    
+    def _is_nope_layer(self, layer_idx: int) -> bool:
+        return (
+            self._config.nope_layer_interval is not None
+            and (layer_idx + 1) % self._config.nope_layer_interval == 0
+        )
+
+    @override
+    def create_ffn(self, layer_idx: int) -> FeedForwardNetwork:
+        config = self._config
+
+        init_std = config.init_std
+
+        std_scale_factor = self.get_std_scale_factor(layer_idx)
+
+        def init_projection(proj: Linear) -> None:
+            input_dim = proj.weight.shape[1]
+
+            std = init_std or (input_dim**-0.5)
+
+            _init_truncated_normal(proj.weight, proj.bias, std=std / std_scale_factor)
+
+        ffn_inner_dim = int(config.ffn_inner_dim * config.ffn_inner_dim_multiplier)
+
+        # TODO: implement interleave_moe_layer_step for Llama 4 Maverick
+
+        return MoE(
+            config.model_dim,
+            ffn_inner_dim,
+            config.experts.use_shared_expert,
+            config.experts.num_experts,
+            config.experts.capacity_factor,
+            config.experts.top_k,
+            inner_dim_scale=config.ffn_inner_dim_scale,
+            inner_dim_to_multiple=config.ffn_inner_dim_multiple_of,
+            eval_with_saved_stats=config.experts.eval_with_saved_stats,
+            expert_act_threshold=config.experts.expert_act_threshold,
+        )
+
+        # TODO: re-introduce for Maverick
+        # return GLUFeedForwardNetwork(
+        #     config.model_dim,
+        #     ffn_inner_dim,
+        #     bias=False,
+        #     inner_dim_scale=config.ffn_inner_dim_scale,
+        #     inner_dim_to_multiple=config.ffn_inner_dim_to_multiple,
+        #     inner_dropout_p=config.dropout_p,
+        #     proj_init_fn=init_projection,
+        # )
+    
+    def create_llama4_rms_norm(
+        self,
+        dim: int,
+        *,
+        elementwise_affine: bool = True,
+    ) -> LayerNorm:
+        # This is llama-stack version, we can try switching
+        return RMSNorm(
+            dim,
+            bias=False,
+            elementwise_affine=elementwise_affine,
+            eps=1e-5,
+        )
+
+    def create_layer_norm(self)-> LayerNorm:
+        return self.create_llama4_rms_norm(
+            self._config.model_dim,
+            elementwise_affine=True,
+        )
+    
+    def create_qk_norm(self, dim: int) -> LayerNorm:
+        return self.create_llama4_rms_norm(
+            dim,
+            elementwise_affine=False,
+        )
--- a/src/fairseq2/models/llama4/interop.py
+++ b/src/fairseq2/models/llama4/interop.py
@@ -1,0 +1,177 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from __future__ import annotations
+
+import re
+from typing import cast
+
+import torch
+
+from fairseq2.models.llama4.config import Llama4Config
+from fairseq2.models.utils.checkpoint import convert_state_dict
+
+
+def get_indices_to_split_wqkv(
+    config: Llama4Config, actual_wqkv_dim_0: int
+) -> tuple[tuple[int, int], tuple[int, int], tuple[int, int]]:
+    """
+    Reference Llama 4 checkpoints have merged WQKV matrices.
+    This method returns three tuples, (q_range, k_range, v_range). They can be used as:
+        - Q: [0, k_end)
+        - K: [k_start, v_start)
+        - V: [v_start, -1)
+    """
+    # compute the expected unsharded combined wqkv dimension
+    n_kv_heads = config.num_key_value_heads or config.num_attn_heads
+    n_local_heads = config.num_attn_heads
+    head_dim = config.model_dim // config.num_attn_heads
+    n_local_kv_heads = n_kv_heads
+    expected_wqkv_dim_0 = n_local_heads * head_dim + 2 * n_local_kv_heads * head_dim
+    # deduce the tensor-parallelism size
+    tp_size = expected_wqkv_dim_0 // actual_wqkv_dim_0
+    # recompute local dimensions
+    n_kv_heads = config.num_key_value_heads or config.num_attn_heads
+    n_local_heads = config.num_attn_heads // tp_size
+    head_dim = config.model_dim // config.num_attn_heads
+    n_local_kv_heads = max(1, n_kv_heads // tp_size)
+    # return sharded ranges
+    return (
+        (0, n_local_heads * head_dim),
+        (
+            n_local_heads * head_dim,
+            n_local_heads * head_dim + n_local_kv_heads * head_dim,
+        ),
+        (
+            n_local_heads * head_dim + n_local_kv_heads * head_dim,
+            n_local_heads * head_dim + 2 * n_local_kv_heads * head_dim,
+        ),
+    )
+
+
+def convert_llama4_state_dict(
+    state_dict: dict[str, object], config: Llama4Config
+) -> dict[str, object]:
+    try:
+        state_dict = cast(dict[str, object], state_dict["model"])  # legacy
+    except KeyError:
+        pass
+    
+    loaded_checkpoint = state_dict
+
+    # Check if we have a reference or Hugging Face checkpoint.
+    if "lm_head.weight" in loaded_checkpoint:  # HG
+        raise ValueError("Llama 4 Huggingface checkpoint is not supported yet.")
+    else:
+        checkpoint = {}  # type: ignore
+
+        for k, v in loaded_checkpoint.items():
+            if ".moe_w_" in k:
+                assert isinstance(v, torch.Tensor)
+                assert config.experts is not None
+
+                num_experts = config.experts.num_experts
+
+                checkpoint[k] = v.unflatten(0, (num_experts, -1))
+
+            elif re.match(r"layers\.([0-9]+)\.attention\.wqkv\.weight", k):
+                assert isinstance(v, torch.Tensor)
+                # split fused QKV weights (along the output dimension)
+                q_range, k_range, v_range = get_indices_to_split_wqkv(
+                    config, v.shape[0]
+                )
+
+                wq = v[q_range[0] : q_range[1]]
+                wk = v[k_range[0] : k_range[1]]
+                wv = v[v_range[0] : v_range[1]]
+
+                wq_key = k.replace(".wqkv.", ".wq.")
+                wk_key = k.replace(".wqkv.", ".wk.")
+                wv_key = k.replace(".wqkv.", ".wv.")
+
+                checkpoint[wq_key] = wq
+                checkpoint[wk_key] = wk
+                checkpoint[wv_key] = wv
+
+            elif "rope.freqs" in k:
+                # We do not need the pre-computed 'rope.freqs' buffers.
+                pass
+
+            elif "._extra_state" in k:
+                # This might contain FP8 quantization info
+                pass
+
+            elif k.startswith("vision_") and not config.vision_config:
+                # Skip vision layers if vision config is disabled
+                pass
+
+            elif k.endswith(".expert_activation_DE"):
+                # This doesn't seem to be used in Scout or Maverick
+                pass
+
+            else:
+                checkpoint[k] = v
+
+        key_map = {
+            # fmt: off
+            r"^layers\.([0-9]+)\.attention\.wq\.": r"decoder.layers.\1.self_attn.q_proj.",
+            r"^layers\.([0-9]+)\.attention\.wk\.": r"decoder.layers.\1.self_attn.k_proj.",
+            r"^layers\.([0-9]+)\.attention\.wv\.": r"decoder.layers.\1.self_attn.v_proj.",
+            r"^layers\.([0-9]+)\.attention\.wo\.": r"decoder.layers.\1.self_attn.output_proj.",
+            r"^layers\.([0-9]+)\.attention.wqkv.layer_norm_weight$": r"decoder.layers.\1.self_attn_layer_norm.weight",
+            r"^layers\.([0-9]+)\.feed_forward\.w1\.": r"decoder.layers.\1.ffn.gate_proj.",
+            r"^layers\.([0-9]+)\.feed_forward\.mlp\.fc2_weight$": r"decoder.layers.\1.ffn.output_proj.weight",
+            r"^layers\.([0-9]+)\.feed_forward\.w3\.": r"decoder.layers.\1.ffn.inner_proj.",
+            r"^layers\.([0-9]+)\.feed_forward\.mlp\.layer_norm_weight$": r"decoder.layers.\1.ffn_layer_norm.weight",
+            # MoE: router
+            r"^layers\.([0-9]+)\.feed_forward\.router_DE$": r"decoder.layers.\1.ffn.router.gate",
+            r"^layers\.([0-9]+)\.feed_forward\.expert_activation_DE$": r"decoder.layers.\1.ffn.router.expert_activation",
+            r"^layers\.([0-9]+)\.feed_forward\.norm\.": r"decoder.layers.\1.ffn_layer_norm.",
+            r"^layers\.([0-9]+)\.feed_forward\.running_gate_stats_3E$": r"decoder.layers.\1.ffn.running_gate_stats",
+            r"^layers\.([0-9]+)\.feed_forward\.global_gate_stats_3E$": r"decoder.layers.\1.ffn.global_gate_stats",
+            # MoE: shared expert
+            r"^layers\.([0-9]+)\.feed_forward\.w_in_shared_FD\.": r"decoder.layers.\1.ffn.shared_expert.gate_proj.",
+            r"^layers\.([0-9]+)\.feed_forward\.w_swiglu_FD\.": r"decoder.layers.\1.ffn.shared_expert.inner_proj.",
+            r"^layers\.([0-9]+)\.feed_forward\.w_out_shared_DF\.": r"decoder.layers.\1.ffn.shared_expert.output_proj.",
+            # MoE: grouped experts
+            r"^layers\.([0-9]+)\.feed_forward\.experts\.moe_w_in_eD_F$": r"decoder.layers.\1.ffn.experts.gate_proj.weight",
+            r"^layers\.([0-9]+)\.feed_forward\.experts\.moe_w_swiglu_eD_F$": r"decoder.layers.\1.ffn.experts.inner_proj.weight",
+            r"^layers\.([0-9]+)\.feed_forward\.experts\.moe_w_out_eF_D$": r"decoder.layers.\1.ffn.experts.output_proj.weight",
+            # Initial layer norm
+            r"^norm\.": r"decoder.layer_norm.",
+            # Embeddings and input projections
+            r"^tok_embeddings\.": r"decoder_frontend.embed.",
+            r"^vision_embeddings\.": r"decoder_frontend.vision_embed.",
+            r"^vision_projection\.": r"decoder_frontend.vision_proj.",
+            r"^speech_embeddings\.embeddings\.": r"decoder_frontend.speech_embed.",
+            # Output projections
+            r"^output\.": r"final_proj.",
+            # fmt: on
+        }
+
+    state_dict = convert_state_dict(checkpoint, key_map)
+
+    return state_dict
+
+
+# 'vision_projection.weight',
+# 'tok_embeddings.weight',
+# 'norm.weight',
+# 'output.weight',
+# 'layers.0.feed_forward.norm.weight', 'layers.0.attention.wo.weight', 'layers.0.feed_forward.global_gate_stats_3E', 'layers.0.feed_forward.expert_activation_DE', 'layers.0.feed_forward.running_gate_stats_3E', 'layers.0.feed_forward.router_DE', 'layers.0.feed_forward.w_in_shared_FD.weight', 'layers.0.feed_forward.w_swiglu_FD.weight', 'layers.0.feed_forward.w_out_shared_DF.weight', 'layers.0.feed_forward.experts.moe_w_in_eD_F', 'layers.0.feed_forward.experts.moe_w_swiglu_eD_F', 'layers.0.feed_forward.experts.moe_w_out_eF_D', 'layers.0.attention.wqkv.weight', 'layers.0.attention.wqkv.layer_norm_weight',
+# 'layers.1.feed_forward.norm.weight', 'layers.1.attention.wo.weight', 'layers.1.feed_forward.expert_activation_DE', 'layers.1.feed_forward.running_gate_stats_3E', 'layers.1.feed_forward.global_gate_stats_3E', 'layers.1.feed_forward.router_DE', 'layers.1.feed_forward.w_in_shared_FD.weight', 'layers.1.feed_forward.w_swiglu_FD.weight', 'layers.1.feed_forward.w_out_shared_DF.weight', 'layers.1.feed_forward.experts.moe_w_in_eD_F', 'layers.1.feed_forward.experts.moe_w_swiglu_eD_F', 'layers.1.feed_forward.experts.moe_w_out_eF_D', 'layers.1.attention.wqkv.weight', 'layers.1.attention.wqkv.layer_norm_weight',
+#
+# 'vision_embeddings.vision_adapter.mlp.c_fc.weight',
+# 'vision_embeddings.vision_adapter.mlp.c_proj.weight',
+# 'vision_embeddings.vision_encoder.class_embedding',
+# 'vision_embeddings.vision_encoder.conv1._linear.weight',
+# 'vision_embeddings.vision_encoder.ln_post.bias',
+# 'vision_embeddings.vision_encoder.ln_post.weight',
+# 'vision_embeddings.vision_encoder.ln_pre.bias',
+# 'vision_embeddings.vision_encoder.ln_pre.weight',
+# 'vision_embeddings.vision_encoder.positional_embedding_vlm',
+# 'vision_embeddings.vision_encoder.transformer.resblocks.0.attn.wk.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.0.attn.wk.weight', 'vision_embeddings.vision_encoder.transformer.resblocks.0.attn.wo.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.0.attn.wo.weight', 'vision_embeddings.vision_encoder.transformer.resblocks.0.attn.wq.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.0.attn.wq.weight', 'vision_embeddings.vision_encoder.transformer.resblocks.0.attn.wv.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.0.attn.wv.weight', 'vision_embeddings.vision_encoder.transformer.resblocks.0.ln_1.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.0.ln_1.weight', 'vision_embeddings.vision_encoder.transformer.resblocks.0.ln_2.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.0.ln_2.weight', 'vision_embeddings.vision_encoder.transformer.resblocks.0.mlp.c_fc.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.0.mlp.c_fc.weight', 'vision_embeddings.vision_encoder.transformer.resblocks.0.mlp.c_proj.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.0.mlp.c_proj.weight',
+# 'vision_embeddings.vision_encoder.transformer.resblocks.1.attn.wk.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.1.attn.wk.weight', 'vision_embeddings.vision_encoder.transformer.resblocks.1.attn.wo.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.1.attn.wo.weight', 'vision_embeddings.vision_encoder.transformer.resblocks.1.attn.wq.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.1.attn.wq.weight', 'vision_embeddings.vision_encoder.transformer.resblocks.1.attn.wv.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.1.attn.wv.weight', 'vision_embeddings.vision_encoder.transformer.resblocks.1.ln_1.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.1.ln_1.weight', 'vision_embeddings.vision_encoder.transformer.resblocks.1.ln_2.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.1.ln_2.weight', 'vision_embeddings.vision_encoder.transformer.resblocks.1.mlp.c_fc.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.1.mlp.c_fc.weight', 'vision_embeddings.vision_encoder.transformer.resblocks.1.mlp.c_proj.bias', 'vision_embeddings.vision_encoder.transformer.resblocks.1.mlp.c_proj.weight',
--- a/src/fairseq2/models/llama4/model/frontend.py
+++ b/src/fairseq2/models/llama4/model/frontend.py
@@ -1,0 +1,142 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+from __future__ import annotations
+
+import math
+from dataclasses import dataclass
+
+from torch import Tensor
+from torch.nn import Dropout
+from typing_extensions import override
+
+from fairseq2.models.llama4.model.vision.embedding import VisionEmbeddings
+from fairseq2.models.transformer import TransformerFrontend
+from fairseq2.nn import (
+    Embedding,
+    IncrementalStateBag,
+    LayerNorm,
+    PositionEncoder,
+    Projection,
+)
+from fairseq2.nn.padding import PaddingMask
+from fairseq2.nn.transformer import LayerNormFactory, create_standard_layer_norm
+
+
+@dataclass
+class MaskedEmbedding:
+    embedding: Tensor
+    mask: Tensor
+
+
+class LLaMA4DecoderFrontend(TransformerFrontend):
+    """Represents a Llama 4 front-end with different embeddings
+    for multiple modalities."""
+
+    embed: Embedding
+    vision_embed: VisionEmbeddings | None
+    vision_proj: Projection | None
+    scale: float
+    pos_encoder: PositionEncoder | None
+    layer_norm: LayerNorm | None
+    dropout: Dropout | None
+
+    def __init__(
+        self,
+        embed: Embedding,
+        vision_embed: VisionEmbeddings | None,
+        vision_proj: Projection | None,
+        pos_encoder: PositionEncoder | None,
+        *,
+        no_scale: bool = False,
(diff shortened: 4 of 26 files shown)


src/fairseq2/models/llama4/sharder.py
```diff
@@ -0,0 +1,123 @@
...
+        r".*\.ffn.output_proj$":       ShardSpec(dim=1, region_boundary=True),
+        # FFN sharding: MoE
+        r".*\.ffn$":                   ShardSpec(dim=-1),  # this goes through a custom sharder
+        # Final proj sharding
+        r"^final_proj$":               ShardSpec(dim=0),
+        # fmt: on
+    }
+
+
+class MoESharder(ModuleSharder):
```
MartinGleize, line 46: Is this the way of implementing some model-specific sharding logic now?

If we assume this MoE implementation is not meant to be re-used outside of Llama, this still means we have to register a niche MoEsharder in composition/lib.py, right?

src/fairseq2/models/llama4/model/vision/embedding.py
MartinGleize, line 1: The vision stuff is mostly pure Pytorch (no re-use of fs2 components) and not properly tested (checkpoint loading and sharding was tested before 0.5, but not now). I'm in favor of keeping it still so that we can iterate on it in another PR.

MartinGleize: Chose to remove it for now (only left config and stub class).

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+    )
+    torch.distributed.all_gather_into_tensor(
+        output_splits,
+        local_output_splits,
+        group=ep_gang.as_process_group(),
+    )
+    return output_splits.int().view(-1).tolist()
+
+
+class DP2EPTokenDispatcher(MoETokenDispatcher):
```
MartinGleize: DP2EP is a specific expert parallelism mode in Llama 4, whereby experts are distributed on DP ranks (so EP gangs piggy-back onto DP gangs).

Expert parallelism is not present in this PR but I've tested it before, so I'll leave this dispatcher here for later EP support (to come in another PR).

src/fairseq2/models/llama4/interop.py
```diff
@@ -0,0 +1,177 @@
...
+import re
+from typing import cast
+
+import torch
+
+from fairseq2.models.llama4.config import Llama4Config
+from fairseq2.models.utils.checkpoint import convert_state_dict
+
+
+def get_indices_to_split_wqkv(
```
cbalioglu: I think this should be a private function: `def _get_indices_to_split_wqkv`.

src/fairseq2/models/llama4/interop.py
```diff
@@ -0,0 +1,177 @@
...
+            n_local_heads * head_dim + 2 * n_local_kv_heads * head_dim,
+        ),
+    )
+
+
+def convert_llama4_state_dict(
+    state_dict: dict[str, object], config: Llama4Config
+) -> dict[str, object]:
+    try:
+        state_dict = cast(dict[str, object], state_dict["model"])  # legacy
```
cbalioglu: Nit: Since we haven't supported LLaMA 4 before, it is safe to remove this check.

src/fairseq2/models/llama4/interop.py
```diff
@@ -0,0 +1,177 @@
...
+    except KeyError:
+        pass
+
+    loaded_checkpoint = state_dict
+
+    # Check if we have a reference or Hugging Face checkpoint.
+    if "lm_head.weight" in loaded_checkpoint:  # HG
+        raise ValueError("Llama 4 Huggingface checkpoint is not supported yet.")
+    else:
```
cbalioglu: Nit: You can save 4 spaces if you remove `else` and unindent the code block.

src/fairseq2/models/llama4/interop.py
```diff
@@ -0,0 +1,177 @@
...
+        checkpoint: dict[str, object] = {}  # type: ignore
+
+        for k, v in loaded_checkpoint.items():
+            if ".moe_w_" in k:
+                assert isinstance(v, torch.Tensor)
```
cbalioglu: I know that `assert` feels more convenient, but I would prefer to raise `ModelCheckpointError` here so that it gets reported in a user-friendly way in the job logs:

```python
if not isinstance(v, Tensor):
  raise ModelCheckpointError(f"{k} parameter is expected to be of type `{Tensor}`, but is of type `{type(v)}` instead.")
```

That `assert` suggestion applies pretty much to all other assert statements in the PR. I always prefer to raise a regular exception instead of AssertionError for anything except invariant checks (which indicate a bug in the implementation)

src/fairseq2/models/llama4/interop.py
```diff
@@ -0,0 +1,177 @@
...
+            r"^output\.": r"final_proj.",
+            # fmt: on
+        }
+
+    state_dict = convert_state_dict(checkpoint, key_map)
+
+    return state_dict
+
+
+# 'vision_projection.weight',
```
cbalioglu: Should we remove these comments? Do they mean anything?

src/fairseq2/models/llama4/interop.py
```diff
@@ -0,0 +1,177 @@
...
+    # compute the expected unsharded combined wqkv dimension
+    n_kv_heads = config.num_key_value_heads or config.num_attn_heads
+    n_local_heads = config.num_attn_heads
+    head_dim = config.model_dim // config.num_attn_heads
+    n_local_kv_heads = n_kv_heads
+    expected_wqkv_dim_0 = n_local_heads * head_dim + 2 * n_local_kv_heads * head_dim
+    # deduce the tensor-parallelism size
+    tp_size = expected_wqkv_dim_0 // actual_wqkv_dim_0
+    # recompute local dimensions
+    n_kv_heads = config.num_key_value_heads or config.num_attn_heads
```
cbalioglu: Nit: If I follow this correctly, these three lines repeat lines 29-31?

src/fairseq2/models/llama4/tokenizer.py
```diff
@@ -0,0 +1,145 @@
...
+    "<|end_of_text|>",
+    "<|fim_prefix|>",
+    "<|fim_middle|>",
+    "<|fim_suffix|>",
+]
+
+
+@dataclass
+class Llama4TokenizerConfig:
+    impl: Literal["sp", "tiktoken", "hg"] = "tiktoken"
```
cbalioglu: I suggest removing this entirely since for now (and likely for the foreseeable future) we will only support tiktoken version. I don't think that there will be ever a sentencepiece version by the way. It was only available for LLaMA-1 and LLaMA-2

src/fairseq2/models/llama4/config.py
```diff
@@ -0,0 +1,160 @@
...
+    """Llama will use a MoE layer as FFN every ``interleave_moe_layer_step``-th layer.
+    If equal to 1, a MoE is used for every layer."""
+
+    eval_with_saved_stats: bool = False
+
+    expert_act_threshold: float = 0.0
+
+
+@dataclass(kw_only=True)
+class Llama4MetapConfig:
```
cbalioglu: Nit: I would name this `Llama4MetaPConfig` (note uppercase P). With lower-case, it feels like a typo :)

src/fairseq2/models/llama4/config.py
```diff
@@ -0,0 +1,160 @@
...
+    since the multiplier is 1.0 in all available checkpoints.
+    """
+
+
+@dataclass(kw_only=True)
+class Llama4Config(LLaMAConfig):
+    experts: Llama4ExpertsConfig = field(default_factory=lambda: Llama4ExpertsConfig())
+    """If not ``None``, specifies the configuration of Mixture-of-Experts."""
+
+    vision_config: Llama4VisionEncoderConfig | None = None
```
cbalioglu: Very nit: I would name this and `metap_config`, just `vision` and `meta_p`. They are part of a config dataclass, so `_config` suffix feels a bit redundant.

cbalioglu: Nit: I would prefer for vision and metap configs to be not `None` and instead having an `enabled` field in the corresponding config dataclass. That would make it easier to work with these configs in YAML config files.

src/fairseq2/models/llama4/model/frontend.py
```diff
@@ -0,0 +1,128 @@
...
+    pos_encoder: PositionEncoder | None
+    layer_norm: LayerNorm | None
+    dropout: Dropout | None
+
+    def __init__(
+        self,
+        embed: Embedding,
+        vision_embed: VisionEmbeddings | None,
+        vision_proj: Projection | None,
+        pos_encoder: PositionEncoder | None,
```
cbalioglu: Is there any LLaMA 4 architecture variant that accepts a positional encoder here. If not, we can remove this.

src/fairseq2/models/llama4/model/frontend.py
```diff
@@ -0,0 +1,128 @@
...
+        *,
+        no_scale: bool = False,
```
cbalioglu: Similarly, based on the factory, this seems to be always `True`. So we might remove it to simplify the code.

src/fairseq2/models/llama4/factory.py
```diff
@@ -0,0 +1,200 @@
...
+        dim: int,
+        *,
+        elementwise_affine: bool = True,
+    ) -> LayerNorm:
+        # This is llama-stack version, we can try switching
+        return RMSNorm(
+            dim,
+            bias=False,
+            elementwise_affine=elementwise_affine,
+            eps=1e-5,
```
cbalioglu: Nit: this is the default eps, so safe to remove.

src/fairseq2/models/llama4/factory.py
```diff
@@ -0,0 +1,200 @@
...
+        )
+
+    def create_layer_norm(self) -> LayerNorm:
+        return self.create_llama4_rms_norm(
```
cbalioglu: Nit: Considering that this function and `create_qk_norm` simply call `RMSNorm(..., elementwise_affine=[True/False]`, you might just want to call it directly instead of hopping through a relatively redundant `create_llama4_rms_norm()` function.

src/fairseq2/models/transformer/attention_bias.py
```diff
@@ -186,6 +219,22 @@ def _create_causal_bias_tensor(
     return bias
 
 
+def _create_chunked_bias_tensor(
+    q_len: int,
+    attention_chunk_size: int,
+    device: Device,
+) -> torch.Tensor:
+    block_pos = torch.abs(
```
cbalioglu: Looks like `torch.abs` is redudant? We only use block_pos on line 232 with `block_pos == 0` which does not get affected by `abs`?

src/fairseq2/models/transformer/attention_bias.py
```diff
@@ -186,6 +219,22 @@ def _create_causal_bias_tensor(
...
+        (torch.arange(q_len).unsqueeze(0) // attention_chunk_size)
+        - (torch.arange(q_len).unsqueeze(1) // attention_chunk_size)
+    )
+    token_pos = torch.arange(q_len).unsqueeze(0) - torch.arange(q_len).unsqueeze(1)
+    mask = (block_pos == 0) & (token_pos <= 0)
+
+    # (S, S)
+    return mask.to(device=device)
```
cbalioglu: Instead of moving from CPU to CUDA here which would cause a host-to-device copy, I suggest creating all intermediate tensors on the relevant device (i.e. `torch.arange(x, device=device)`

src/fairseq2/models/transformer/attention_bias.py
```diff
@@ -97,6 +97,39 @@ def __repr__(self) -> str:
...
+        return hash((self.attn_chunk_size,))
+
+    @override
+    def create_bias_tensor(
+        self, q_len: int, k_len: int, device: Device, dtype: DataType
+    ) -> Tensor:
+        if q_len != k_len:
+            raise ValueError(f"`q_len` and `k_len` must be equal: {q_len} != {k_len}")
+        # (S, S)
+        return _create_chunked_bias_tensor(q_len, self.attn_chunk_size, device)
```
cbalioglu: Nit: since this is the only place where `_create_chunked_bias_tensor` is called, you might consider putting its content right into this method (similar to ALiBi).

src/fairseq2/models/transformer/ffn.py
```diff
@@ -297,6 +308,13 @@ def forward(self, seqs: Tensor) -> Tensor:
         if self.inner_dropout is not None:
             seqs = self.inner_dropout(seqs)
 
+        return seqs
+
+    def forward_output(self, seqs: Tensor) -> Tensor:
```
cbalioglu: I don't think this method is even needed at all right? It is equivalent to simply calling `ffn.output_proj(seqs)`.

src/fairseq2/models/llama4/model/frontend.py
```diff
@@ -0,0 +1,128 @@
...
+            self.register_module("dropout", None)
+
+    @override
+    def forward(
+        self,
+        seqs: Tensor,  # [batch_size, seq_len] or [batch_size, seq_len, 3]
+        seqs_layout: BatchLayout,
+        *,
+        state_bag: IncrementalStateBag | None = None,
+        image_embedding: MaskedEmbedding | None = None,
```
cbalioglu: Ok, this looks problematic because strictly speaking this is not a `TransformerFrontend` anymore. We can remove this parameter till we introduce vision support and in the meantime think of a way to pass it in a "type-safe" way. Also wondering whether the `embedding` and `mask` of `MaskedEmbedding` can be represented as part of `seqs`.

src/fairseq2/models/llama4/model/moe/moe.py
```diff
@@ -0,0 +1,217 @@
...
+        """
+        Args:
+            seqs (torch.Tensor): Input tensor with shape ``(bs, slen, dim)``.
+
+        Returns:
+            out (torch.Tensor): Output tensor with shape ``(bs, slen, dim)``.
+        """
+        x = seqs
+        # Update running stats for experts threshold
+        self.running_gate_stats = self.running_gate_stats.float()  # type: ignore[has-type]
```
cbalioglu: Beyond not using these two buffers at all, these 4 statements feel also quite off. Why don't we just create these two buffers in `__init__` on the right device with `float32` instead of repeatedly calling these ops here? (assuming we want to keep them around for a reason that I am not aware of right now)

src/fairseq2/models/llama4/model/moe/router.py
```diff
@@ -0,0 +1,130 @@
...
+    Returns:
+        bool: True if the current computation is during AC recomputation, False otherwise.
+    """
+    return torch._C._current_graph_task_id() != -1
+
+
+class RouterOutput(NamedTuple):
+    """Router output for Token Choice and Expert Choice routing.
+
+    routed_input (torch.Tensor): tokens grouped together by experts indices with shape
```
cbalioglu: Have you verified the correctness of these docstrs? In particular for token routing my assumption was that the returned `routed_input` has a shape of `(E,S)` where `E` is the number of experts and `S` is batch size x sequence length containing the (sparse top-k) scores per expert.

src/fairseq2/models/llama4/model/moe/router.py
```diff
@@ -0,0 +1,130 @@
...
+    num_local_tokens_per_expert: Optional[torch.Tensor]
+
+
+class Router(nn.Module):
+    """This class implements experts choice routing. Each experts will select it's top K tokens based on
+        the router scores. Refer to more details in https://arxiv.org/abs/2202.09368
+
+    The router is never sharded.
+
+    Args:
```
cbalioglu: These args do not match the parameters of `__init__`.

src/fairseq2/models/llama4/model/moe/router.py
```diff
@@ -0,0 +1,130 @@
...
+    expert_act_threshold: float
+
+    def __init__(
+        self,
+        *,
+        model_dim: int,
+        num_experts: int,
+        top_k: int,
+        expert_act_threshold: float = 0.0,
+        is_tp_sharding_strategy: bool = True,
```
cbalioglu: Looks like not used?

src/fairseq2/models/llama4/model/moe/router.py
```diff
@@ -0,0 +1,130 @@
...
+        scores = (
+            torch.full_like(logits, float("-inf"))
+            .scatter_(1, token_indices, scores)
+            .transpose(0, 1)
+        )
+
+        token_indices = (
+            torch.arange(bszseqlen, device=x.device)
+            .view(1, -1)
+            .expand(scores.size(0), -1)
```
cbalioglu: If want to keep this around, it would be slightly more readable to use `num_experts` in place of `scores.size(0)`.

src/fairseq2/models/llama4/model/moe/router.py
```diff
@@ -0,0 +1,130 @@
...
+    token_indices (torch.Tensor): token indices for routed_input with shape
+        ``(bs*slen*experts_per_token,)`` for Token Choice, ``(num_experts*tokens_per_expert,)`` for Expert Choice.
+    num_local_tokens_per_expert (Optional[torch.Tensor]):
+        Number of tokens assigned to each expert with shape ``(num_experts,)`` if using Token Choice.
+        None if using Expert Choice since tokens_per_expert is fixed across experts.
+    """
+
+    routed_input: torch.Tensor
+    token_indices: torch.Tensor
+    num_local_tokens_per_expert: Optional[torch.Tensor]
```
cbalioglu: I believe this field was relevant for some advanced use cases that do not exist in this implementation where the number of experts could be dynamic due to some regularization/load balancing. If that is the case, we can remove it here.

src/fairseq2/models/llama4/model/moe/router.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+class Router(nn.Module):
+    """This class implements experts choice routing. Each experts will select it's top K tokens based on
+        the router scores. Refer to more details in https://arxiv.org/abs/2202.09368
```
cbalioglu: The pointed literature is the expert choice routing paper, but this implementation is a very basic token choice implementation.

src/fairseq2/nn/grouped_projection.py
```diff
@@ -0,0 +1,403 @@
...
+from fairseq2.device import META_DEVICE, Device
+from fairseq2.gang import Gang
+from fairseq2.nn.utils.module import to_empty
+from fairseq2.ops.tensor_parallel import gather, reduce, reduce_on_backward, scatter
+
+
+class GroupedProjection(Module, ABC):
+    """
+    Applies a grouped linear transformation to incoming data.
+    Alsp known as grouped GEMMs or batch matrix-matrix product (BMM).
```
cbalioglu: Typo: s/Alsp/Also.

Strictly speaking, I don't think we should refer to this as BMM. Mathematically it becomes only a BMM when all groups are used for a given input.

src/fairseq2/nn/grouped_projection.py
```diff
@@ -0,0 +1,403 @@
...
+
+    buffer = torch.empty(
+        int(cum_size[-1].item()),
+        weight.shape[-1],
+        dtype=x.dtype,
+        device=x.device,
+    )
+
+    start = 0
+    for expert_index, end in enumerate(cum_size):
```
cbalioglu: Do you mind adding a TODO here mentioning that this will be replaced with/used as a fallback to a `torch._grouped_mm` or a cuBLAS based implementation (if we want to support beyond bf16) in the near future.

src/fairseq2/nn/grouped_projection.py
```diff
@@ -0,0 +1,403 @@
...
+        start = end
+
+    return buffer
+
+
+@final
+class BatchLinear(GroupedProjection):
+    def __init__(
+        self,
+        extra_first_dim: int,
```
cbalioglu: Like above I would call this `group_dim`.

src/fairseq2/nn/grouped_projection.py
```diff
@@ -0,0 +1,403 @@
...
+            weight[expert_index],
+            out=buffer[start:end],
+        )
+        start = end
+
+    return buffer
+
+
+@final
+class BatchLinear(GroupedProjection):
```
cbalioglu: I would call this `GroupedLinear`. `BatchLinear` sounds a bit confusing tbh as it implies as if the regular `Linear` cannot be used with batched inputs (i.e. inputs with [batch_size, seq_len, model_dim]).

src/fairseq2/nn/grouped_projection.py
```diff
@@ -0,0 +1,403 @@
...
+        :param extra_first_dim:
+            The size of the extra first dimension (for example: num_experts).
+        :param input_dim:
+            The dimensionality of inputs.
+        :param output_dim:
+            The dimensionality of projected outputs.
+        """
+        super().__init__(extra_first_dim, input_dim, output_dim)
+
+        self.weight = Parameter(
```
cbalioglu: It does not necessarily have to be part of this PR since for now these modules are only used with pretrained LLaMA-4 models, but right now we leave the projection weight simply uninitialized once the module is constructed.

src/fairseq2/nn/grouped_projection.py
```diff
@@ -0,0 +1,403 @@
...
+            )
+        )
+
+    @override
+    def forward(self, x: Tensor, num_inputs_per_group: Tensor | None) -> Tensor:
+        return _maybe_bmm(x, self.weight, num_inputs_per_group)
+
+
+@final
+class BatchColumnShardedLinear(GroupedProjection):
```
cbalioglu: Same feedback as `BatchLinear`. I think it will be more understandable if we named this `GroupedColumnShardedLinear` and `group_dim` instead of `extra_first_dim`.

src/fairseq2/nn/grouped_projection.py
```diff
@@ -0,0 +1,403 @@
...
+            f"{s}, "
+            f"gather_output={self.gather_output}, "
+            f"rank={self.gang.rank}, "
+            f"world_size={self.gang.size}"
+        )
+
+        return s
+
+
+class BatchRowShardedLinear(GroupedProjection):
```
cbalioglu: Same feedback as `BatchLinear`. I think it will be more understandable if we named this `GroupedRowShardedLinear` and `group_dim` instead of `extra_first_dim`.

src/fairseq2/models/llama4/model/moe/experts.py
```diff
@@ -0,0 +1,86 @@
...
+        Args:
+            x (Tensor): tensor with shape ``(num_local_experts, tokens_per_expert, dim)``.
+            num_tokens_per_expert (Optional[Tensor]): tensor with shape ``(num_local_experts,)``.
+        Returns:
+            torch.Tensor: Tensor with shape (num_local_experts, tokens_per_expert, dim)
+        """
+        # x shape (num_local_experts, tokens_per_expert, dim)
+        h: Tensor = self.activation(self.gate_proj(x, num_tokens_per_expert))
+
+        if self.inner_proj is not None:
```
cbalioglu: `inner_proj` can never be `None` right?

src/fairseq2/models/llama4/model/moe/experts.py
```diff
@@ -0,0 +1,86 @@
...
+
+from __future__ import annotations
+
+from torch import Tensor, nn
+from torch.nn import Module, SiLU
+
+from fairseq2.nn.grouped_projection import BatchLinear, GroupedProjection
+
+
+class Experts(nn.Module):
```
cbalioglu: How about calling this `ExpertNetwork` to sound similar to regular `FeedForwardNetwork`?

src/fairseq2/models/llama4/model/moe/experts.py
```diff
@@ -0,0 +1,86 @@
...
+        self.activation: Module
+        if activation is None:
+            self.activation = SiLU()
+        else:
+            self.activation = activation
+
+    def create_layers(self) -> None:
+        """
+        TODO(mgleize): For finetuning
+        We intentionally fold the expert weights' ``num_local_experts`` dim-0
```
cbalioglu: I assume this is something not implemented so far? I think we need to address this before trying out this implementation in the fine-tuning recipe. Ideally it is also something that needs to be handled in `GroupedProjection` subclasses instead of here.

src/fairseq2/models/llama4/model/moe/experts.py
cbalioglu: Also does it make sense to move this to `fairseq2.transformer.ffn`? I believe this is pretty much standard for MoE architectures? If not, we can leave it under llama4.

src/fairseq2/models/llama4/model/moe/router.py
```diff
@@ -0,0 +1,130 @@
...
+
+import torch
+import torch.nn as nn
+
+from fairseq2.data_type import DataType
+from fairseq2.device import Device
+
+
+# TODO: update it with a formal AC check
+def check_ac_recompute() -> bool:
```
cbalioglu: This seems to be used nowhere. Safe to remove?

src/fairseq2/models/llama4/model/moe/moe.py
```diff
@@ -0,0 +1,217 @@
...
+        self.global_gate_stats = self.global_gate_stats.float()  # type: ignore[has-type]
+        self.global_gate_stats.data = self.global_gate_stats.data.to(x.device)
+
+        experts_threshold = None
+        if not self.training and self.eval_with_saved_stats:
+            # Assume std is 0, just take the mean of the scores as the threshold
+            experts_threshold = self.global_gate_stats[0]
+
+        bs, slen, dim = x.shape
+        # top_scores shape (num_experts, tokens_per_expert) for EC and (bs*slen*experts_per_token, ) for TC
```
cbalioglu: We can remove mentions of EC if LLaMA-4 does not use it at all.

src/fairseq2/models/llama4/model/moe/moe.py
```diff
@@ -0,0 +1,217 @@
...
+        # selected_indices shape (num_experts, tokens_per_expert) for EC training,
+        # (num_experts, bs*slen) for EC inference, and (bs*slen*experts_per_token,) for TC
```
cbalioglu: Is `(bs*slen*experts_per_token)` true for TC? At least the router implementation we have right now seems to be always returning `(num_experts, slen)` (where k elements of each sequence have a non-zero score).

src/fairseq2/models/llama4/model/moe/moe.py
```diff
@@ -0,0 +1,217 @@
...
+            )
+        else:
+            self.shared_expert = None
+
+        self.eval_with_saved_stats = eval_with_saved_stats
+
+        self.running_stats_ema = running_stats_ema
+
+        # running stats for each expert shape (mean, var*count, count)
+        self.register_buffer(
```
cbalioglu: Seems we are not using these two buffers anywhere. `global_gate_stats[0]` is passed as `experts_threshold` to the router, but it is not used there as well. If that is the case, we can remove these all together.

src/fairseq2/models/llama4/model/moe/moe.py
```diff
@@ -0,0 +1,217 @@
...
+
+        self.experts = Experts(
+            model_dim=model_dim,
+            inner_dim=inner_dim,
+            num_local_experts=num_experts,
+            activation=moe_activation,
+        )
+
+        # Use a TP token dispatcher for non-EP scenarios
+        self.token_dispatcher = TPTokenDispatcher(num_experts)
```
cbalioglu: My expectation would be for `MoE` to accept a `MoETokenDispatcher`. The llama-4 factory would construct a simple noop dispatcher (essentially an equivalent of `TokenTPDispatcher`) as if the entire model was going to be run on a single device. Later the model sharder would construct a distributed dispatcher (e.g. DP2EP, TP2EP) depending on the passed `gangs` parameter and update the `MoE` instance in-place.

src/fairseq2/models/llama4/model/moe/router.py
```diff
@@ -0,0 +1,130 @@
...
+    eval_with_expert_activation_model: bool
+    expert_act_threshold: float
+
+    def __init__(
+        self,
+        *,
+        model_dim: int,
+        num_experts: int,
+        top_k: int,
+        expert_act_threshold: float = 0.0,
```
cbalioglu: Looks like not used?

src/fairseq2/models/llama4/model/moe/router.py
```diff
@@ -0,0 +1,130 @@
...
+            .transpose(0, 1)
+        )
+
+        token_indices = (
+            torch.arange(bszseqlen, device=x.device)
+            .view(1, -1)
+            .expand(scores.size(0), -1)
+        )
+
+        scores = torch.sigmoid(scores)
```
cbalioglu: Should this be done in fp32?

src/fairseq2/models/llama4/model/moe/router.py
```diff
@@ -0,0 +1,130 @@
...
+
+        scores, token_indices = torch.topk(logits, self.top_k, dim=1)
+
+        scores = (
+            torch.full_like(logits, float("-inf"))
+            .scatter_(1, token_indices, scores)
+            .transpose(0, 1)
+        )
+
+        token_indices = (
```
cbalioglu: Unless we will have an alternative token choice implementation for LLaMA 4, I don't think this needs to be returned from router. It effectively can be represented by a simple `torch.repeat()` call in `MoE`.

src/fairseq2/models/llama4/model/moe/moe.py
```diff
@@ -0,0 +1,217 @@
...
+        self.running_gate_stats.data = self.running_gate_stats.data.to(x.device)
+        self.global_gate_stats = self.global_gate_stats.float()  # type: ignore[has-type]
+        self.global_gate_stats.data = self.global_gate_stats.data.to(x.device)
+
+        experts_threshold = None
+        if not self.training and self.eval_with_saved_stats:
+            # Assume std is 0, just take the mean of the scores as the threshold
+            experts_threshold = self.global_gate_stats[0]
+
+        bs, slen, dim = x.shape
```
cbalioglu: Note that input shape can be just `(slen, dim)` (no batch dim), in particular with packed batching.

src/fairseq2/models/llama4/model/moe/moe.py
```diff
@@ -0,0 +1,217 @@
...
+        # Gather the output from all experts (possibly with all-to-all)
+        routed_output = self.token_dispatcher.token_unpermutation(
+            routed_output,
+            permute_metadata=permute_metadata,  # type: ignore[arg-type]
+        )
+
+        # Finish shared expert computation
+        if self.shared_expert is not None:
+            assert shared_expert_out is not None
+            shared_expert_out = self.shared_expert.forward_output(shared_expert_out)
```
cbalioglu: What is the motivation for splitting the shared expert's computation in two parts? Why not do the whole forward pass of it at the same time with the dipatcher all-to-all?

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+                    The input splits used by all2all.
+                output_splits (torch.Tensor) [ep]:
+                    The output splits used by all2all.
+
+        Returns:
+            routed_in_egD (torch.Tensor) [num_local_experts, num_global_tokens_per_expert, hidden_dim]:
+                Permuted tokens for local experts.
+        """
+        # Prepare inputs_splits and output_splits for all2all
+        if tokens.dim() == 1:
```
cbalioglu: I don't think we have any use case where `tokens` can be 1-dimensional right? If not we can just remove this if condition.

src/fairseq2/models/llama4/config.py
```diff
@@ -0,0 +1,160 @@
...
+
+from fairseq2.models.llama import LLaMAConfig
+from fairseq2.runtime.config_registry import ConfigRegistrar
+from fairseq2.runtime.dependency import DependencyContainer
+
+LLAMA4_FAMILY: Final = "llama4"
+
+
+@dataclass(kw_only=True)
+class Llama4VisionEncoderConfig:
```
cbalioglu: Nit: Optionally we can remove this till we have the vision encoder ready.

src/fairseq2/models/llama4/config.py
```diff
@@ -0,0 +1,160 @@
...
+    auto_scale: float = True
+
+    top_k: int = 1
+    """MoE sends each token to the ``top_k`` top experts."""
+
+    interleave_moe_layer_step: int = 1
+    """Llama will use a MoE layer as FFN every ``interleave_moe_layer_step``-th layer.
+    If equal to 1, a MoE is used for every layer."""
+
+    eval_with_saved_stats: bool = False
```
cbalioglu: Looks like we don't use this anywhere in the current PR.

src/fairseq2/models/llama4/config.py
```diff
@@ -0,0 +1,160 @@
...
+
+    expert_act_threshold: float = 0.0
```
cbalioglu: Looks like we don't use this anywhere in the current PR.

src/fairseq2/models/llama4/config.py
```diff
@@ -0,0 +1,160 @@
...
+    num_experts: int = 16
+    """The number of MoE experts."""
+
+    use_shared_expert: bool = True
+    """If ``True``, an additional single expert is used for all tokens."""
+
+    capacity_factor: float = 1.0
+    """The capacity factor of experts."""
+
+    auto_scale: float = True
```
cbalioglu, line 32: Maybe a short docstr what this means? Something like "used to scale the inner dimensionality of the (shared) expert feed forward networks by `capacity_factor`."

src/fairseq2/models/llama4/factory.py
```diff
@@ -0,0 +1,200 @@
...
+    _config: Llama4Config
+
+    def __init__(self, config: Llama4Config) -> None:
+        self._config = config
+
+    @override
+    def create_decoder_frontend(self, embed: Embedding) -> TransformerFrontend:
+        config = self._config
+
+        if config.vision_config:
```
cbalioglu: Nit: As I mentioned in another feedback, I would completely remove this whole vision configuration for now and set them to just `None`.

src/fairseq2/models/llama4/factory.py
```diff
@@ -0,0 +1,200 @@
...
+            config.experts.num_experts,
+            config.experts.capacity_factor,
+            config.experts.top_k,
+            inner_dim_scale=config.ffn_inner_dim_scale,
+            inner_dim_to_multiple=config.ffn_inner_dim_multiple_of,
+            eval_with_saved_stats=config.experts.eval_with_saved_stats,
+            expert_act_threshold=config.experts.expert_act_threshold,
+        )
+
+        # TODO: re-introduce for Maverick
```
cbalioglu: We can remove this for now.

src/fairseq2/models/llama4/model/moe/moe.py
```diff
@@ -0,0 +1,217 @@
...
+        num_experts: int,
+        capacity_factor: float,
+        top_k: int,
+        *,
+        moe_activation: nn.Module | None = None,
+        inner_dim_scale: float = 2 / 3,
+        inner_dim_to_multiple: int = 1,
+        auto_scale: bool = True,
+        running_stats_ema: float = 0.99,
+        eval_with_saved_stats: bool = True,
```
cbalioglu: Like in the model configuration, I don't see any use of this in this PR.

src/fairseq2/models/llama4/model/moe/moe.py
```diff
@@ -0,0 +1,217 @@
...
+        expert_act_threshold: float = 0.0,
```
cbalioglu: Like in the model configuration, I don't see any use of this in this PR.

src/fairseq2/models/llama4/model/moe/moe.py
```diff
@@ -0,0 +1,217 @@
...
+        use_shared_expert: bool,
+        num_experts: int,
+        capacity_factor: float,
+        top_k: int,
+        *,
+        moe_activation: nn.Module | None = None,
+        inner_dim_scale: float = 2 / 3,
+        inner_dim_to_multiple: int = 1,
+        auto_scale: bool = True,
+        running_stats_ema: float = 0.99,
```
cbalioglu: Not used?

src/fairseq2/models/llama4/model/moe/moe.py
```diff
@@ -0,0 +1,217 @@
...
+        (
+            top_scores,
+            token_indices,
+            num_local_tokens_per_expert,
+        ) = self.router(x.reshape(bs * slen, dim), experts_threshold=experts_threshold)
+
+        tokens_per_expert = top_scores.shape[1]
+
+        # token_indices shape (num_experts*tokens_per_expert, dim) for training and (num_experts*bs*slen, dim) for inference
+        token_indices = token_indices.reshape(-1, 1).expand(-1, dim)
```
cbalioglu: Assuming we won't have futher changes in this MoE architecture, I think we can entirely get rid of `token_indices` and just call `routed_input = x.repeat(num_experts, dim=0)`. Maybe I am wrong though, haven't executed any of this code, so please take my feedback with grain of salt.

src/fairseq2/models/llama4/model/moe/moe.py
```diff
@@ -0,0 +1,217 @@
...
+        # routed_input shape (num_experts*tokens_per_expert, dim)
+        routed_input = torch.gather(x.view(-1, dim), dim=0, index=token_indices)
+        routed_input = routed_input * top_scores.reshape(-1, 1)
+
+        # Dispatch of tokens to experts (possibly with all-to-all)
+        # permute_metadata is only useful if sharding_strategy == "dp2ep-v1"
```
cbalioglu: What is `dp2ep-v1`, just `dp2ep`?

src/fairseq2/models/llama4/model/moe/moe.py
cbalioglu: Why do we split the async forward call of the shared expert into two parts (inner projection concurrent with permutation and outer projection concurrent with unpermutation)?

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+import torch
+from typing_extensions import override
+
+from fairseq2.gang import Gang
+
+
+class ShardingStrategy(Enum):
+    TP = auto()
+    # Not supported in fairseq2 for now
+    # TP2EP = auto()
```
cbalioglu: What is missing for us to support TP2EP?

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+        # do a reverse all2all along TP ranks to revert the layout back to [num_tokens/tp, dim],
+        # and an allgather among the TP ranks.
+        self.sharding_strategy = sharding_strategy
+        self.num_experts = num_experts
+
+    @abstractmethod
+    def token_permutation(
+        self,
+        tokens: torch.Tensor,  # [num_tokens, dim]
+        permute_metadata: Optional[Dict[str, torch.Tensor]] = None,
```
cbalioglu: We use the cleaner annotation syntax introduced in Python 3.10: `dict[str, Tensor] | None`. This applies to all annotations in this PR.

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+    ) -> torch.Tensor: ...
```
cbalioglu: I assume this has to be extended to `tuple[Tensor, Future]`? At least in DP2EP implementation we return a tuple here.

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+
+    def wait_token_permutation(
+        self,
+        tokens: torch.Tensor,
```
cbalioglu: Similar to `token_permutation()`, my understanding is that this has to take a tuple of `tuple[Tensor, Future]`.

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+        permute_metadata: Optional[Dict[str, torch.Tensor]] = None,
+    ) -> torch.Tensor:
+        return tokens
+
+    @abstractmethod
+    def token_unpermutation(
+        self,
+        tokens: torch.Tensor,
+        permute_metadata: Optional[Dict[str, torch.Tensor]] = None,
+    ) -> torch.Tensor: ...
```
cbalioglu: Same "tuple" feedback as the permutation counterpart.

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+        """
+        Initialize the MoE Token Dispatcher.
+        """
+        super().__init__(ShardingStrategy.TP, num_experts)
+
+    @override
+    def token_permutation(
+        self,
+        tokens: torch.Tensor,  # [num_tokens, dim]
+        permute_metadata: Optional[Dict[str, torch.Tensor]] = None,
```
cbalioglu: tbh, I don't like this sort of untyped dictionaries to carry additional parameters. It feels a leaky abstraction to me. At least for now, I don't see any other metadata than `tokens_per_expert` passed to both dispatcher types. We can simply define a `tokens_per_expert` parameter instead (or better yet, pass it in `__init__` since that information does not change).

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+    input_ = input_.contiguous()
+    if output_split_sizes is None:
+        # Equal split (all2all)
+        output = torch.empty_like(input_)
+    else:
+        # Unequal split (all2all-v)
+        output = input_.new_empty(
+            size=[sum(output_split_sizes)] + list(input_.size()[1:]),
+            dtype=input_.dtype,
+            device=torch.cuda.current_device(),
```
cbalioglu: `input_.device` is both cleaner and safer.

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+
+    local_output_splits = torch.tensor(
+        input_splits[0],
+        dtype=torch.int,
+        device=torch.cuda.current_device(),
+    )
+    output_splits = torch.zeros(
+        (ep_size),
+        dtype=torch.int,
+        device=torch.cuda.current_device(),
```
cbalioglu: Same as above.

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+    """Gather tensors and concatenate along the last dimension."""
+    ep_size = ep_gang.size
+    # Bypass the function if we are using only 1 GPU.
+    if ep_size == 1:
+        return input_splits
+
+    local_output_splits = torch.tensor(
+        input_splits[0],
+        dtype=torch.int,
+        device=torch.cuda.current_device(),
```
cbalioglu: Same as above.

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+    )
+    output_splits = torch.zeros(
+        (ep_size),
```
cbalioglu: Nit "bug". This should be either `ep_size` or `(ep_size,)`

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+    ) -> None:
+        """
+        Initialize the MoE Token Dispatcher.
+        """
+        super().__init__(ShardingStrategy.DP2EP, num_experts)
+        self.ep_gang = ep_gang
+        num_local_experts = num_experts // ep_gang.size
+        self.num_local_experts: int = num_local_experts
+
+        self.local_expert_indices: List[int] = list(
```
cbalioglu: Seems to be not used anywhere?

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+        if tokens.dim() == 1:
+            B_T = tokens.shape
+            D = 1
+        else:
+            B_T, D = tokens.shape  # type: ignore[assignment]
+
+        assert permute_metadata is not None
+
+        with torch.profiler.record_function("metadata preparation"), torch.no_grad():
+            if "input_splits" not in permute_metadata["nt"]:
```
cbalioglu: We never pass `input_splits` to this function. So safe to assume that we have `tokens_per_expert` here (ideally as a regular parameter or member variable)

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+    num_local_experts: int
+    local_expert_indices: List[int]
+    # This can be used when batch_size and seq_len are identical on DP ranks
+    are_output_input_splits_equal: bool
+
+    def __init__(
+        self,
+        num_experts: int,
+        ep_gang: Gang,
+        are_output_input_splits_equal: bool = True,
```
cbalioglu: Always `True` based on the rest of this PR?

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+                ):
+                    # if cudagraph (decode), we assume all ranks are having the same
+                    # batch size, otherwise cudagraph will be broken
+                    output_splits = input_splits
+                else:
+                    # for eager, we will first collect the tokens from other experts
+                    # to make sure alltoall will match. This will be slow due to the
+                    # extra allgather but will be safe (if we cannot guarantee the
+                    # prefill will have exactly the same token size)
+                    output_splits = _gather_output_splits(
```
cbalioglu: Do we even need `_gather_output_splits` considering in this PR input and output splits have always the same length?

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+        Initialize the MoE Token Dispatcher.
+        """
+        super().__init__(ShardingStrategy.TP, num_experts)
+
+    @override
+    def token_permutation(
+        self,
+        tokens: torch.Tensor,  # [num_tokens, dim]
+        permute_metadata: Optional[Dict[str, torch.Tensor]] = None,
+    ) -> torch.Tensor:
```
cbalioglu: Since we are returning a tuple in DP2EP implementation, we should revise the signature of this method (and the other related methods). We should either always return a tuple of tensor and future, or an opaque type for the user to pass to the wait methods.

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+        assert permute_metadata is not None
+        # Sort a2a outputs by local experts when num_local_experts > 1
+        if tokens_eg_D.dim() == 1:
+            D = 1
+        else:
+            _, D = tokens_eg_D.shape
+        with torch.profiler.record_function("token local unpermute"):
+            if self.num_local_experts > 1:
+                if torch.cuda.graphs.is_current_stream_capturing():
```
cbalioglu: Why this conditional? Why not always? Especially for the implementation in this PR?

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
```diff
@@ -0,0 +1,443 @@
...
+                    for i in output_splits:
+                        per_expert_output_splits.extend(
+                            [i // self.num_local_experts] * self.num_local_experts
+                        )
+                    tokens_eg_D_splits = torch.split(
+                        tokens_eg_D, per_expert_output_splits, dim=0
+                    )
+
+                tokens_eg_D_reorder = []
+                for eid in range(self.num_local_experts):
```
cbalioglu: Haven't executed the code so far, but this can be simplified to a simple indexing operation on `tokens_eg_D_splits`. Again, take this feedback with a grain of salt.

src/fairseq2/models/llama4/model/vision/embedding.py
```diff
@@ -0,0 +1,37 @@
...
+from __future__ import annotations
+
+from typing import List
+
+import torch
+
+from fairseq2.models.llama4.config import Llama4VisionEncoderConfig
+
+
+class VisionEmbeddings(torch.nn.Module):
```
cbalioglu: I suggest removing this entirely for now.

src/fairseq2/models/llama4/sharder.py
cbalioglu: Under this API design, I feel like this `MoESharder` seems more like a hack to workaround the limitation of the sharder API. Let's keep this as it is, but I will think of a way to make this "cleaner".

src/fairseq2/models/transformer/attention_bias.py
```diff
@@ -97,6 +97,39 @@ def __repr__(self) -> str:
         return f"CausalAttentionBias(attn_window_len={self.attn_window_len})"
 
 
+@final
+class ChunkedAttentionBias(AttentionBias):
```
cbalioglu, line 101: Have you tried out how this would play out with flash2/flash3 attention? As far as I know neither version has support for it. I also just found out this: https://x.com/winglian/status/1910430245854773523

That would mean a significant slow-down during training, especially for doc-causal masking. Looks like flex attention is the only potential short-term solution (which has another PR) that we should prioritize if we want to use LLaMA 4 for any real-world use case.

Not sure how your experience has been so far without an optimized attention support for chunked bias.

src/fairseq2/nn/grouped_projection.py
```diff
@@ -0,0 +1,403 @@
...
+    """Represents a batched 3D tensor multiplication that is sharded across its input dimension."""
+
+    @staticmethod
+    def from_batch_linear(
+        linear: BatchLinear,
+        gang: Gang,
+        scatter_input: bool = False,
```
cbalioglu: I would make the default `True` to be consistent with `ColumnShardedLinear`.

src/fairseq2/sharder.py
```diff
@@ -29,7 +32,20 @@
 @dataclass
 class ShardSpec:
     dim: int
+    """The sharded dimension."""
+
     region_boundary: bool = False
+    """Whether the sharded dimension is at the boundary between
+    two model parallel regions. Allows to avoid unnecessary communication."""
+
+    disable_end_reduce: bool = False
```
cbalioglu: I plan to revise this API (in a BC way) so we can have more descriptive name for this. For the time being it can stay as is.

src/fairseq2/sharder.py
```diff
@@ -29,7 +32,20 @@
...
+    """Whether to disable a reduce planned at the end
+    of the forward of a sharded layer.
+    If ``True``, it is the responsibility of the layer's user to perform the reduce."""
+
+    shard_children: bool = False
```
cbalioglu: We likely won't need this. Let me think more about the sharding APIs. In the meantime it can stay here as it is.

cbalioglu (commented):
Sorry for the delay. It is a huge PR so took me a while to go over it. I left quite a few comments, but most of them are not really major issues and more about simplifying the code. It is clear that the core pieces of the PR were taken from the internal code base and modified for OSS, so tried to address the issues caused from that. One additional feedback is to verify that all docstrings are correct. I think overall there are inconsistencies between what is implemented and what is documented (especially in MoE, Router, and TokenDispatcher).

Thanks again for the work, looking forward to next revision!

src/fairseq2/sharder.py
MartinGleize: Reverted my changes to this and moved the custom logic to the MoESharder.

src/fairseq2/sharder.py
MartinGleize: Reverted my changes to this and moved the custom logic to the MoESharder.

src/fairseq2/models/llama4/interop.py
MartinGleize: There were definitely redundant lines there!

src/fairseq2/models/llama4/config.py
MartinGleize: Removed.

src/fairseq2/models/llama4/config.py
MartinGleize: Both removed.

src/fairseq2/models/llama4/model/frontend.py
MartinGleize: Removed the frontend since we don't need vision for now.

src/fairseq2/models/transformer/attention_bias.py
MartinGleize: Good catch!

src/fairseq2/models/llama4/model/moe/router.py
MartinGleize: Router is removed. It used to contain like 3 different implementations in earlier versions, but the few lines we have right now don't justify it.

src/fairseq2/nn/grouped_projection.py
MartinGleize: This is a great point, I didn't think about the distinction when naming these classes. Anyway I got rid of grouped_projection.py altogether since I only needed to use 3 BMMs for now (unlike reference code which has a lot of different variants).

Decided to shift effort and logic to experts.py (which contains the TP logic now).

src/fairseq2/models/llama4/model/moe/experts.py
MartinGleize: Good idea. Ended up putting experts.py in fairseq2.models.transformer.
* ExpertNetwork is the ABC
* GroupedExpertNetwork is the basic unsharded implem with 3D weights and BMMs.
* TPShardedExpertNetwork is the TP-sharded version.
* We can add others when EP comes.

MartinGleize: Moved experts to fairseq2.models.transformer . It was enriched, documented and made slightly more general so this felt like a good place to put it, but happy to keep it under llama4 too.

src/fairseq2/models/llama4/model/moe/moe.py
MartinGleize: Hmm I simplified so much that I didn't realize I removed this, but I'll have another go soon at documenting the MoE forward with comments, shapes, etc to give a better idea of what's happening.

src/fairseq2/models/llama4/model/moe/moe.py
MartinGleize: Removed the dispatching, but we'll probably need to re-introduce it soon.
EP was not present in my PR (for example, it doesn't set up gangs, and it doesn't shard the expert weights, although I got both of those in our internal code) so I'll get rid of the dispatching temporarily as well.


Will keep your advice in mind! I think this is pretty much what I was doing in the internal code.

src/fairseq2/models/llama4/model/moe/router.py
MartinGleize: Good question, reference code has a flag to turn this on, and llama-stack has this version. I think in the end for most models that matter, this should be done in float32.

src/fairseq2/models/llama4/model/moe/moe.py
MartinGleize: Very important, thank you.

src/fairseq2/models/llama4/model/moe/moe.py
MartinGleize: What I think it did:
1. Compute first part of the shared expert forward    |   Dispatch expert inputs
2. Compute (non-shared) expert forward
3. Compute last part of the shared expert forward    |   Dispatch expert outputs

In any case I removed all this since we have a no-op dispatching.

src/fairseq2/models/llama4/config.py
MartinGleize: Removed.

src/fairseq2/models/llama4/factory.py
MartinGleize: Added the implementation.

src/fairseq2/models/llama4/model/moe/moe.py
MartinGleize: There is a dp2ep-v2 :P. Anyway what we had was the v1.

src/fairseq2/models/llama4/model/moe/moe.py
MartinGleize: Yes that's it (see answer above). Honestly I'm not sure this was the product of the precise benchmarking this type of finegrained optim requires (imo). I can look into it when we introduce EP.

src/fairseq2/models/llama4/model/moe/token_dispatcher.py
MartinGleize: Noted.

MartinGleize:
* Removed most of the currently-dead code.

* Refactored experts into a single generic experts.py including the TP-sharded impl. (in src/fairseq2/models/transformer/experts.py)

* Gathered MoE sharding logic to a single place instead of modify the core lib.

src/fairseq2/models/transformer/experts.py
MartinGleize, line 1: Would appreciate another look at this. Should it be there or stay in llama4? @cbalioglu

src/fairseq2/models/transformer/experts.py
```diff
@@ -0,0 +1,328 @@
...
+    if TYPE_CHECKING:
+        __call__ = forward
+
+
+class _GLUExpertNetwork:
+    """A mixin for GLU-based experts.
+    Their layers are implemented as 3-D tensors used in BMM.
+    """
+
+    def _create_layers(
```
cbalioglu: Nit: this can be a standalone function like `create_expert_glu_layers` instead of a mixin.

src/fairseq2/models/transformer/experts.py
cbalioglu: Thanks a lot @MartinGleize! This looks great. Just left a very nit comment. Also note that I plan to land a change today/tomorrow about how we do model sharding that will potentially deprecate ModelSharder. I have already applied that change to other models. I can quickly do that for LLaMA-4 as well one we land this.

cbalioglu (approved):
Thanks a lot for the revision! It looks much cleaner now. I only left a very nit comment. Otherwise, looks good to me!

src/fairseq2/models/transformer/attention_bias.py
MartinGleize: This is definitely an issue yes. I haven't experimented with inputs long enough for the chunk size to matter yet. But I added this issue to our internal L4 tracking doc as a priority next.

As far as I see, the reference training code heavily referenced xformers' memory-efficient attention as a backend.

src/fairseq2/models/llama4/model/moe/experts.py
MartinGleize: This is the very next thing I'll implement. This is probably not such a huge problem for low-enough DP but critical to enable scaling.

Issue #1303, opened 2025-09-12
[doc] llama tokenizer


zyaoj:
It's not well documented about the use of llama tokenizer e.g. w/ {"impl": "hg"} by default.

For example, we should include examples to make it work by defining tokenizer section:

```py
from fairseq2.models.llama.tokenizer import LLaMATokenizerConfig
from fairseq2.recipe.config import TokenizerSection

@dataclass(kw_only=True)
class YourConfig:
    ...

    tokenizer: TokenizerSection = field(
        default_factory=lambda: TokenizerSection(
            name="llama3",
            config_overrides=LLaMATokenizerConfig(impl="hg", use_eot=True),
        )
    )
```

We should also provide examples to make it work with asset cards:

```
$ python -m fairseq2.assets show llama3_2_1b_instruct
llama3_2_1b_instruct
  source          : 'package:omt.cards'
  base            : 'llama3_instruct'
  model_arch      : 'llama3_2_1b'
  checkpoint      : '/path/to/models/Llama-3.2-1B-Instruct/original/consolidated.00.pth'

llama3_instruct
  source          : 'package:fairseq2.assets.cards'
  base            : 'llama3'
  tokenizer_config_override: {'use_eot': True}

llama3
  source          : 'package:fairseq2_ext.cards'
  model_family    : 'llama'
  checkpoint      : 'https://ai.meta.com/llama/;gated=true'
  tokenizer       : '/path/to/models/Llama-3.1-8B/'
  tokenizer_family: 'llama'
  tokenizer_config_override: {'impl': 'hg'}
  use_v2_tokenizer: True
```

PR #1306, opened 2025-09-15
[doc] enhance v0.5.* documentation (sprint 1 - round 1)

### LINKED ISSUE #1300: [doc] Hub (dataset, tokenizer, model)
Document the use of hubs in fairseq2, similar to [what we have for model hub](https://facebookresearch.github.io/fairseq2/stable/reference/api/models/hub.html#quick-start).

For example for [TokenizerHub](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/data/tokenizers/hub.py#L27-L92), we expose additional helper functions now (i.e.` load_tokenizer`, `load_custom_tokenizer(path)`, `get_tokenizer_config()`, and `iter_cards()`).

zyaoj:

- Add doc around hub (dataset, tokenizer, model)
- Add doc around llama (and it's tokenizer)
- Add doc around recipe asset registry (extra path)

Fixes #1300 #1303

N/A

**Check list:**


doc/source/basics/assets.rst
```diff
@@ -16,7 +16,13 @@ Understanding the Asset System
 The fairseq2 asset system consists of three main components:
 
 1. **Asset Cards**: YAML files that describe assets and their metadata
-2. **Asset Stores**: Collections of asset cards from various sources
+2. **Asset Stores**: Collections of asset cards from various sources:
+
+   - Built-in cards: ``fairseq2/assets/cards/``
+   - System-wide cards: ``/etc/fairseq2/assets/``
```
cbalioglu: maybe a brief mention in parentheses that this is overridable via FAIRSEQ2_ASSET_DIR/FAIRSEQ2_USER_ASSET_DIR?

doc/source/reference/api/data/tokenizers/hub.rst
```diff
@@ -0,0 +1,133 @@
...
+
+Functions
+---------
+
+load_tokenizer
+~~~~~~~~~~~~~~
+
+.. autofunction:: load_tokenizer
+
+    The primary function for loading tokenizers. This is a high-level function that handles
```
cbalioglu: I wouldn't call this "primary". It is a global, family-agnostic convenience function to load tokenizers. If the user knows the family of tokenizer beforehand, this is almost equivalent to `get_xyz_tokenizer_hub().load_tokenizer(...)`.

Similarly, the same complexities are also handled by the hub methods.

doc/source/reference/api/data/tokenizers/index.rst
```diff
@@ -68,7 +68,70 @@ Loading a Tokenizer
...
+Loading a Specific Model's Tokenizer
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: python
+
+    from fairseq2.models.qwen import get_qwen_tokenizer_hub
+
+    hub = get_qwen_tokenizer_hub()
+    
+    for card in hub.iter_cards():
```
cbalioglu: I think this loop is not really needed to load a tokenizer. It would make more sense to have a separate short code snippet for demonstrating how the user can get the list of all available tokenizers of a specific family.

doc/source/reference/api/data/tokenizers/index.rst
```diff
@@ -68,7 +68,70 @@ Loading a Tokenizer
...
+
+
+This loads the tokenizer and its associated vocabulary for the specified model.
+
+
+Using TokenizerHub
+~~~~~~~~~~~~~~~~~~
+
+:class:`TokenizerHub` provides more advanced/customized operations for working with tokenizers.
+This is helpful if you want to create your own tokenizer, and configuration.
```
cbalioglu: I think "create" is not the right term. You still "load" a tokenizer, you just don't need an asset card and instead provide a path and a configuration.

doc/source/reference/api/data/tokenizers/index.rst
```diff
@@ -68,7 +68,70 @@ Loading a Tokenizer
...
+Here's how to use it with Qwen tokenizers:
+
+.. code-block:: python
+
+    from fairseq2.data.tokenizers.hub import TokenizerHubAccessor
+    from fairseq2.models.qwen import QwenTokenizer, QwenTokenizerConfig
+    from pathlib import Path
+
+    # this is the same as get_qwen_tokenizer_hub()
+    hub = TokenizerHubAccessor(
```
cbalioglu: I would suggest just using `get_qwen_tokenizer_hub()` here. Ideally `TokenizerHubAccessor` is a type that should only be used by the implementor of a tokenizer. For regular users, the way to access an accessor should be via standalone `get_xyz_tokenizer_hub()` functions.

doc/source/reference/api/data/tokenizers/index.rst
```diff
@@ -68,7 +68,70 @@ Loading a Tokenizer
...
+        "qwen",  # tokenizer family name
+        QwenTokenizer,  # concrete tokenizer class
+        QwenTokenizerConfig,  # concrete tokenizer config class
+    )()
+
+    for card in hub.iter_cards():
```
cbalioglu: Again, I believe this loop as well as the direct `load_tokenizer` call right below it cause noise. Both are demonstrated in the same document above. It would be nicer if we could just show the relevant part.

doc/source/reference/api/datasets/hub.rst
```diff
@@ -0,0 +1,161 @@
...
+    .. code-block:: python
+
+         from fairseq2.composition import register_dataset_family
+
+        register_dataset_family(
+            container,             # DependencyContainer instance
+            "custom_dataset",      # family name
+            CustomDataset,         # dataset class
+            CustomDatasetConfig,   # config class
+            opener=custom_opener   # optional opener function
```
cbalioglu: This is not really optional. The user must provider either `opener` or `advanced_opener`. The difference between `opener` and `advanced_opener` is that advanced variant accepts an extra `resolver: DependencyResolver` as first parameter. This would be helpful if the user needs to access other services/objects in fairseq2. For most basic use cases, or just quick hacky tests, `opener` would be slightly easier. Having said that, I would mention `advanced_opener` later in the doc instead of mentioning it here to make it easier for the user to understand the concepts.

doc/source/reference/api/datasets/hub.rst
```diff
@@ -0,0 +1,161 @@
...
+
+DatasetHubAccessor
+~~~~~~~~~~~~~~~~~~
+
+.. autoclass:: DatasetHubAccessor
+    :members:
+    :show-inheritance:
+
+    Factory class for creating :class:`DatasetHub` instances for specific dataset families.
+    Used internally by model-specific hub accessors.
```
cbalioglu: Maybe instead of saying "internally", we should say that the direct use of `DatasetHubAccessor` (similarly model and tokenizer hub accessors) is mostly relevant for dataset/tokenizer/model implementors. It is by no means internal and has to be used in fairseq2 code base only.

doc/source/reference/api/datasets/index.rst
```diff
@@ -0,0 +1,81 @@
...
+    # 3. Register in Recipe
+    class YourRecipe(TrainRecipe):  # or EvalRecipe/GenerationRecipe
+        @override
+        def register(self, container: DependencyContainer) -> None:
+            register_dataset_family(
+                container,
+                "custom_dataset",           # family name
+                CustomDataset,              # dataset class
+                CustomDatasetConfig,        # config class
+                opener=open_custom_dataset  # optional opener function
```
cbalioglu: Again, this is not really optional.

doc/source/reference/api/datasets/index.rst
```diff
@@ -0,0 +1,81 @@
...
+    @dataclass
+    class CustomDatasetConfig:
+        path: Path
+        # Other config options
+
+    # 3. Register in Recipe
+    class YourRecipe(TrainRecipe):  # or EvalRecipe/GenerationRecipe
+        @override
+        def register(self, container: DependencyContainer) -> None:
+            register_dataset_family(
```
cbalioglu, line 108: Maybe we can have a similar example like this where we register the dataset directly as part of the recipe for tokenizers and models as well.

doc/source/reference/api/models/llama.rst
```diff
@@ -0,0 +1,238 @@
...
+    .. code-block:: yaml
+
+        name: llama3
+        tokenizer: "/path/to/Llama-3.1-8B"  # HuggingFace tokenizer directory
+        tokenizer_family: llama
+
+    Directory structure should contain e.g. ``config.json``, ``tokenizer.json``, ``tokenizer_config.json``, ``special_tokens_map.json``.
+
+2. **Tiktoken Implementation**:
+    Legacy implementation using Tiktoken.
```
cbalioglu: I would not say "legacy" since this still has the extra `split_regex` feature that is not available in the HG version.

doc/source/reference/api/models/llama.rst
```diff
@@ -0,0 +1,238 @@
...
+
+        name: tiktoken_llama_instruct
+        tokenizer_config_override:
+            impl: tiktoken
+            use_eot: True  # For instruction models
+        tokenizer_family: llama
+        tokenizer: "/path/to/tokenizer.model"  # Tiktoken model file
+
+3. **SentencePiece Implementation**:
+    Legacy implementation using SentencePiece.
```
cbalioglu: This is not legacy, but is only available for LLaMA-1 and LLaMA-2.

cbalioglu (commented):
Overall looks great. My comments are mostly minor and would be great if you could address them before we land this.

doc/source/reference/api/data/tokenizers/hub.rst
```diff
@@ -0,0 +1,133 @@
...
+
+    Example:
+
+    .. code-block:: python
+
+        from fairseq2.models.qwen import get_qwen_tokenizer_hub
+
+        hub = get_qwen_tokenizer_hub()
+
+        for card in hub.iter_cards():
```
cbalioglu, line 44: Maybe add a short inline comment on top of this loop similar saying what it does similar to other statements below?

doc/source/reference/api/data/tokenizers/hub.rst
```diff
@@ -0,0 +1,133 @@
...
+
+TokenizerHubAccessor
+~~~~~~~~~~~~~~~~~~~~
+
+.. autoclass:: TokenizerHubAccessor
+    :members:
+    :show-inheritance:
+
+    Factory class for creating :class:`TokenizerHub` instances for specific tokenizer families.
+    Used internally by model-specific hub accessors like :func:`fairseq2.models.qwen.get_qwen_tokenizer_hub`.
```
cbalioglu: I think I mentioned it for datasets in my previous review; I think we can remove "internally" since this is a public API for model/tokenizer/dataset authors.

doc/source/reference/api/data/tokenizers/index.rst
```diff
@@ -0,0 +1,160 @@
...
+
+.. currentmodule:: fairseq2.data.tokenizers
+
+The tokenizer has multiple concrete implementations for different tokenization algorithms.
+The main :class:`Tokenizer` interface defines the contract for creating encoders and decoders, while concrete implementations
+handle specific tokenization methods like SentencePiece and tiktoken.
+
+.. mermaid::
+
+   classDiagram
```
cbalioglu: I think it is fine if we just dump the API of Tokenizer and just refer to the concrete implementations. We do not really convey any real information here besides their names.

doc/source/reference/api/data/tokenizers/index.rst
```diff
@@ -0,0 +1,160 @@
...
+~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: python
+
+    from fairseq2.data.tokenizers import load_tokenizer
+
+    tokenizer = load_tokenizer("qwen3_0.6b")
+
+
+Listing Available Tokenizers
```
cbalioglu: I think this should go a bit down, right above "Listing Available Tokenizers (CLI)"

doc/source/reference/api/datasets/hub.rst
```diff
@@ -0,0 +1,161 @@
...
+.. currentmodule:: fairseq2.datasets.hub
+
+The dataset hub provides a centralized way to manage and access datasets in fairseq2.
+It offers functionality for discovering available datasets, loading datasets, and working
+with custom dataset configurations.
+
+Dataset Registration and Asset Cards
+------------------------------------
+
+Datasets in fairseq2 can be registered in two ways:
```
cbalioglu: I don't think this is right. A dataset implementation can only be registered via `register_dataset_family()`. That function can potentially be called in `Recipe.register()` or in a fairseq2 extension function.

Assets cards though describe the actual datasets. The corresponding dataset family has to be already registered while loading fairseq2.

cbalioglu (approved):
Ok, left a few nit comments to address, but already stamping. Feel free to merge it once ready.

PR #1314, opened 2025-09-15
Improve metric value handling in recorders

cbalioglu:
This PR includes two improvements in existing `MetricRecorder` classes:

(1) Beyond scalars, provided metric values can also be sequences of scalars (recursively). This is in particular needed for use cases like having multiple learning rates under a single metric.
(2) Instead of failing with a recorder specific (typically non-user friendly) error, all recorders check that the provided metric values are either scalar (i.e. int, float, Tensor, str) or a sequence of them before attempting to write them.

All four recorders (log, jsonl, tb, wandb) have been tested with various metric types to verify correctness. In the future, we should ideally cover these cases with some proper integration tests.

### DIFF
--- a/src/fairseq2/metrics/formatters.py
+++ b/src/fairseq2/metrics/formatters.py
@@ -7,6 +7,7 @@
 from __future__ import annotations
 
 import math
+from collections.abc import Sequence
 from functools import partial
 from typing import Final
 
@@ -22,6 +23,8 @@ def format_as_int(value: object, *, postfix: str | None = None) -> str:
             i = int(value)
         except ValueError:
             return f"{value}"
+    elif isinstance(value, Sequence):
+        return ", ".join(format_as_int(e) for e in value)
     else:
         return f"{value}"
 
@@ -46,6 +49,8 @@ def format_as_float(value: object, *, postfix: str | None = None) -> str:
             f = float(value)
         except ValueError:
             return f"{value}"
+    elif isinstance(value, Sequence):
+        return ", ".join(format_as_float(e) for e in value)
     else:
         return f"{value}"
 
@@ -66,6 +71,8 @@ def format_as_percentage(value: object) -> str:
             f = float(value)
         except ValueError:
             return f"{value}"
+    elif isinstance(value, Sequence):
+        return ", ".join(format_as_percentage(e) for e in value)
     else:
         return f"{value}"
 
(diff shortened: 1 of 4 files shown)


src/fairseq2/metrics/recorders/wandb.py
```diff
@@ -39,19 +40,41 @@ def __init__(
...
+        if isinstance(value, (int, float, Tensor, str)):
+            output[name] = value
+
+            return
+
+        if isinstance(value, Sequence):
+            for idx, elem in enumerate(value):
+                self._add_value(f"{name} ({idx})", elem, output)
+
+            return
```
cirquit, line 73: Does wandb support multiple scalars for the same timestep naively? I am not sure what I would expect to see in the default view for a LR plot with multiple points per timestep.

zyaoj: I think here it's going to plot several lines w/ different line keys? Don't know if idx would be clear enough for the multiple LR use case to help distinguish to which idx/value the LR correspond to though.

src/fairseq2/metrics/formatters.py
```diff
@@ -22,6 +23,8 @@ def format_as_int(value: object, *, postfix: str | None = None) -> str:
             i = int(value)
         except ValueError:
             return f"{value}"
+    elif isinstance(value, Sequence):
+        return ", ".join(format_as_int(e) for e in value)
```
zyaoj, line 27: [optional] do we want to add a max constraint to avoid logging a super long sequence just in case? But normally I would assume that this is not the case for our metrics.

zyaoj (approved):
LGTM! 2 nit comments

src/fairseq2/metrics/recorders/wandb.py
cbalioglu: Yeah, as Julien mentioned, the extra `idx` appended to the name distinguishes between different LRs. I believe we can resolve the idx to parameter group mapping simply by documentation. Index 0 means default/global parameter group which is not associated with any "key selection pattern" (will be available in the third PR), other indices correspond to the order of selection patterns defined in the recipe configuration. In virtually almost all cases that I can think of, there are at most two LRs. So I don't think figuring out which one is which will be too much of a work for the user.

Just in case, I also include a screenshot how wandb dashboard looks like with three LRs.
Screenshot From 2025-09-15 16-32-22

src/fairseq2/metrics/formatters.py
cbalioglu: This API is almost always called by recipe authors. So I think it would be better if we let them decide what is considered "long" for their use case.

src/fairseq2/metrics/recorders/wandb.py
cirquit: Ah I get it, I think the screenshot shows this nicely!

PR #1320, opened 2025-09-17
Support multiple parameter groups in LR scheduler factories

### LINKED ISSUE #1316: Support multiple parameter groups in recipe LR factories

cbalioglu:
This PR (1) adds support for multiple parameter groups in LR scheduler factories, (2) starts adding/improving docstrs for existing and new code constructs, and (3) adds unit tests for all LR factories with near 100% coverage. Resolves #1316

### DIFF
--- a/doc/source/reference/api/index.rst
+++ b/doc/source/reference/api/index.rst
@@ -58,3 +58,4 @@ Utilities
    device
    logging
    metrics
+   fairseq2.recipe.optim
--- a/src/fairseq2/optim/lr_schedulers/cosine_annealing.py
+++ b/src/fairseq2/optim/lr_schedulers/cosine_annealing.py
@@ -21,7 +21,7 @@ from fairseq2.optim.lr_schedulers.lr_scheduler import (
 
 @final
 class CosineAnnealingLR(AbstractLRScheduler):
-    """Represents the learning rate schedule described in
+    """Represents the cosine annealing learning rate schedule described in
     :cite:t:`https://doi.org/10.48550/arxiv.1608.03983`.
 
     **During warmup:**
@@ -43,13 +43,14 @@ class CosineAnnealingLR(AbstractLRScheduler):
     of starting with a large learning rate that is relatively rapidly decreased
     to a minimum value before being increased rapidly again.
 
-    Please refer to the paper to learn more about the details.
+    Refer to the paper to learn more about the details.
 
     In addition to the original schedule, this implementation also supports a
     warmup phase where the learning rate is linearly increased for the first
     :math:`T_{warmup}` training steps to the base learning rate.
 
     .. note::
+
         This scheduler is not chainable.
     """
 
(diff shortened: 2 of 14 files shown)


doc/source/reference/api/index.rst
```diff
@@ -58,3 +58,4 @@ Utilities
    device
    logging
    metrics
```
cbalioglu, line 60: I think we should move to a package-based file naming convention for the rest of the reference docs.

src/fairseq2/optim/lr_schedulers/cosine_annealing.py
```diff
@@ -121,31 +116,31 @@ def _compute_lrs(self) -> list[float]:
...
-        lr_mul = self._lr_mul**cycle_nr
+        lr_mul = self.lr_mul**cycle_nr
 
         c = math.cos(math.pi * cycle_pos / cycle_len)
 
-        min_lrs, max_lrs = self._final_lrs, base_lrs
+        min_lrs, max_lrs = self.final_lrs, base_lrs
 
-        return [self._cycle_lr(mn, mx, lr_mul, c) for mn, mx in zip(min_lrs, max_lrs)]
+        def cycle_lr(min_lr: float, max_lr: float) -> float:
```
zyaoj, line 140: Curious about the motivation of moving `_cycle_lr` from a private method to a local function.

src/fairseq2/recipe/config.py
```diff
@@ -908,16 +940,25 @@ def validate(self) -> ValidationResult:
 
 @dataclass(kw_only=True)
 class TriStageLRConfig(Validatable):
+    """Represents the configuration of :class:`TriStageLRConfig`."""
```
zyaoj: Did you want to put:
```suggestion
    """Represents the configuration of :class:`TriStageLR`."""
```

src/fairseq2/recipe/trainer.py
```diff
@@ -881,7 +883,10 @@ def _publish_metrics(self) -> None:
 
             self._unit.process_metric_values(values)
 
-            values["lr"] = self._last_lr
+            if len(self._last_lrs) == 1:
```
zyaoj, line 888: [nit] Feels like this could be potentially confusing for people who are not aware of our code and the goal of supporting multiple lr. Maybe add a comment here or abstract these for lines into a helper function?

src/fairseq2/recipe/optim.py
```diff
@@ -0,0 +1,57 @@
...
+            if isinstance(start_lr, float):
+                return [start_lr] * num_param_groups
+
+            maybe_raise_param_group_length_error("start_lr", start_lr, num_param_groups)
+
+            return start_lr
+    """
+    if len(value) != num_param_groups:
+        raise ValidationError(
+            f"The length of `{field}` must match the number of optimizer parameter groups ({num_param_groups}), but is {len(value)} instead.", field="lr_scheduler.config"  # fmt: skip
```
zyaoj, line 56: [nit] Make error message consistent across the board (or merge into the same function?) -- e.g. in [`optim.lr_schedulers.lr_scheduler`](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/optim/lr_schedulers/lr_scheduler.py#L57):

```py
    if len(value) != num_param_groups:
        raise ValueError(
            f"The length of `{name}` must be equal to the number of parameter groups ({num_param_groups}), but is {len(value)} instead."
        )
```

doc/source/reference/api/index.rst
zyaoj: Yeah many of those files were placeholders. Do we want to use a simple flat structure then instead of the hierarchical layout?

zyaoj (approved):
LGTM! Just some nit comment and questions. Thank you for boosting the pytest coverage!

```
src/fairseq2/optim/__init__.py                                   3      0   100%
src/fairseq2/optim/fp16_loss_scaler.py                         130     62    52%
src/fairseq2/optim/lr_schedulers/__init__.py                    10      0   100%
src/fairseq2/optim/lr_schedulers/cosine_annealing.py            43      0   100%
src/fairseq2/optim/lr_schedulers/lr_scheduler.py                34      4    88%
src/fairseq2/optim/lr_schedulers/myle.py                        22      1    95%
src/fairseq2/optim/lr_schedulers/noam.py                        18      0   100%
src/fairseq2/optim/lr_schedulers/polynomial_decay.py            31      1    97%
src/fairseq2/optim/lr_schedulers/tri_stage.py                   42      1    98%
src/fairseq2/recipe/internal/lr_schedulers.py                  122      9    93%

doc/source/reference/api/index.rst
cbalioglu: Yes, I think that will be easier to maintain as well as to quickly spot which file maps to which module.

src/fairseq2/recipe/config.py
cbalioglu: Oh good catch, thanks!

src/fairseq2/optim/lr_schedulers/cosine_annealing.py
cbalioglu: `_cycle_len` was only used in `_compute_lrs` as part of the last list comprehension statemet and required capturing `lr_mul` and `c` local variables. I thought that it would be cleaner and also reduce code if it became a local function (two less parameters to pass).

doc/source/reference/api/index.rst
zyaoj: Sounds good. Will create a follow-up PR to restructure. Also a quick note that there are many examples squeezed into the reference api pages for now. We should probably also surface them to basics or tutorials.

PR #1326, opened 2025-09-23
Move cluster discovery to clusterscope

luccabb:
 
moving cluster discovery and operations to clusterscope

## test plan

```shell
$ conda activate fairseq2
$ pip install flash-attn
$ srun --gpus-per-node=1 -A admin_testing --pty python -m recipes.lm.train --config model.config_overrides="{num_layers: 2}" dataset.config_overrides.path=/checkpoint/seamless_fs2/shared/datasets/shuffled/dclm_baseline_1.0 --config model.compile=False -- ~
```

training succeeds: P1971178271

### DIFF
--- a/src/fairseq2/cluster.py
+++ b/src/fairseq2/cluster.py
@@ -97,7 +97,7 @@ class ClusterNotDetectedError(Exception):
 @final
 class SlurmHandler(ClusterHandler):
     @cached_property
-    def _job(self) -> clusterscope.JobInfo:
+    def _job(self) -> clusterscope.job_info.JobInfo:
         try:
             return clusterscope.get_job()
         except RuntimeError as ex:


src/fairseq2/recipe/config.py
```diff
@@ -525,7 +526,7 @@ class CommonSection:
 
     debug: bool = False
 
-    cluster: str = "auto"
+    cluster: str = clusterscope.cluster()
```
cbalioglu: This should ideally stay as "auto" for two reasons (1) "auto" is a static text and means that fairseq2 should infer the cluster which might be slurm, ray, kubernetes, or simply torchrun (2) changing this breaks BC where we have several teams relying on this to be "auto"

recipes/chatbot/cli.py
```diff
@@ -68,14 +64,6 @@ def _main() -> None:
         get_console().print()
 
         raise
-    except ClusterNotDetectedError as ex:
```
cbalioglu, line 71: I would much prefer for us to keep these strongly-typed exceptions for error reporting purposes.

src/fairseq2/recipe/cli.py
```diff
@@ -558,23 +555,6 @@ def _handle_checkpoint_not_found_error(ex: CheckpointNotFoundError) -> int:
     return 2
 
 
-def _handle_cluster_not_detected_error(ex: ClusterNotDetectedError) -> int:
```
cbalioglu, line 588: Same as above, these errors indicate how the user can fix their configuration in case they have a cluster issue. I would prefer them to stay in the code.

src/fairseq2/recipe/internal/cluster.py
```diff
@@ -1,33 +0,0 @@
...
-    ) -> None:
-        self._section = section
-        self._cluster_resolver = cluster_resolver
-
-    def prepare(self) -> None:
-        handler = self._cluster_resolver.resolve(self._section.cluster)
-
-        handler.set_torch_distributed_env_variables()
-
-        if self._section.cluster == "auto":
```
cbalioglu, line 29: `common.cluster` recipe configuration field is widely used and we should keep supporting it.

src/fairseq2/cluster.py
```diff
@@ -1,195 +0,0 @@
...
-
-class ClusterNotDetectedError(Exception):
-    def __init__(self, cluster: str) -> None:
-        super().__init__(f"Process is not running on a {cluster} cluster.")
-
-        self.cluster = cluster
-
-
-@final
-class SlurmHandler(ClusterHandler):
```
cbalioglu, line 98: I think an integration with clusterscope makes sense mainly within this class. In particular we can use it inside `SlurmHandler.set_torch_distrbuted_env_variables()`.

src/fairseq2/recipe/cli.py
skalyan: We will look into the exceptions you are using and what we need to fill in terms of gaps. If it is purely for communicating a message to user it is a bit different from giving an opportunity to clients/app-code to handle it and take alternate routes.

How do you see your design right now? Sounds like only error-msg/guidance right?

src/fairseq2/recipe/config.py
skalyan: Seems like this is a special annotation, we should leave it as "auto" @luccabb

recipes/chatbot/cli.py
luccabb: In what scenario would we expect cluster to not be detected?

in clusterscope, we're preferring to name it as `local-node` in case none of cluster detection works: https://github.com/facebookresearch/clusterscope/blob/main/clusterscope/cluster_info.py#L200

cbalioglu: For instance, if a recipe configuration has `common.cluster` set to `slurm`, but a user accidentally tries to run the recipe on a submit node or locally, then we raise `ClusterNotDetectedError`.

src/fairseq2/cluster.py
```diff
@@ -97,81 +95,15 @@ def __init__(self, cluster: str) -> None:
 @final
 class SlurmHandler(ClusterHandler):
     def __init__(self, env: Environment) -> None:
-        self._job_id: int | None = None
-        self._env = env
+        self._job = clusterscope.get_job()
```
cbalioglu: Unfortunately this is problematic. First of all `get_job()` is a heavy call when `is_slurm_job()` is true including running scontrol as a subprocess which we should not really call in a constructor. This constructor is also executed way early in the process lifetime while we build the dependency graph. So if anything fails in `get_job()`, the user will have no clue what is going on even if they don't use our cluster API at all.

src/fairseq2/cluster.py
```diff
@@ -97,81 +95,15 @@ def __init__(self, cluster: str) -> None:
...
-            "Hostname or IP address of the Slurm node corresponding to rank 0 cannot be retrieved."  # fmt: skip
-        )
-
-    def _get_master_port(self, job_id: int) -> str:
-        port = self._env.maybe_get("MASTER_PORT")
-        if port is not None:
-            return port
-
-        return str(Random(job_id).randint(20_000, 60_000))
+        self._job.set_torch_distributed_env_from_slurm()
```
cbalioglu: We need to have a `ClusterNotDetectedError` check here as well as proper handling of `RuntimeError` (wrapping it as `OperationalError`).

src/fairseq2/cluster.py
```diff
@@ -97,81 +95,15 @@ def __init__(self, cluster: str) -> None:
...
 
     @override
     def supports_current_cluster(self) -> bool:
-        return self._env.has("SLURM_PROCID")
+        return self._job.is_slurm_srun()
```
cbalioglu, line 118: Similar to the constructor issue, we have to pay a big price in this function (parsing all env variables, determining master address and port) just to query if we are in an srun/sbatch process. IMO the constructor of `JobInfo` should be much more lightweight. Instead of eagerly initializing all job information, it should be done lazily (in fact I see `lru_cache` annotations in JobInfo, but not sure why they are called in __init__ which seems redundant to me)

luccabb: yeah, good point, lets remove these: https://github.com/facebookresearch/clusterscope/pull/149

src/fairseq2/cluster.py
```diff
@@ -97,81 +96,21 @@ def __init__(self, cluster: str) -> None:
 @final
 class SlurmHandler(ClusterHandler):
     def __init__(self, env: Environment) -> None:
-        self._job_id: int | None = None
-        self._env = env
+        self._job = clusterscope.get_job()
```
cbalioglu: Ok, two final nits. Just to be on the safe side in case `get_job()` for some reason becomes expensive again in the future, it would be nice to lazily initialize it:

```python
from functools import cached_property

...

class SlurmHandler:
  ...
  @cached_property
  def _job(self) -> JobInfo:
    try:
      return clusterscope.get_job()
    except RuntimeError as ex:
      raise OperationalError("`clusterscope.get_job()` has failed.") from ex
```

Can you also remove `env` parameter from `__init__` please since we do not use it anymore.

luccabb: done, removed `__init__ ` as we're moving `env` and `_job` out of it

zyaoj (approved):
LGTM!

PR #1328, opened 2025-09-23
Extend ObjectValidator to support composite objects

cbalioglu:
This PR (1) extends `ObjectValidator` to support composite sub-objects. Till now, it only supported dataclasses and their sub-dataclasses reachable from their fields. With this PR, we now support `list`, `Mapping`, `Set`, and `tuple` as well, (2) adds both docstrings and some tutorial documentation, (3) introduces unit tests for `StandardObjectValidator` and `ValidationError`.

### DIFF
--- a/src/fairseq2/utils/validation.py
+++ b/src/fairseq2/utils/validation.py
@@ -12,7 +12,7 @@ that all options are set correctly.
 A class -*typically a configuration dataclass*- that wants to support validation
 should expose a ``validate(self) -> ValidationResult`` method. Optionally,
 the class can derive from the runtime-checkable :class:`Validatable` protocol
-to make its intend more clear.
+to make its intent more clear.
 
 A typical implementation of a ``validate()`` method looks like the following:
 
@@ -199,17 +199,13 @@ class ValidationResult:
 
     def has_error(self) -> bool:
         """
-        Returns ``True`` if the object or any of its sub-ojects have a
+        Returns ``True`` if the object or any of its sub-objects have a
         validation error.
         """
         if self._errors:
             return True
 
-        for sub_result in self._sub_results.values():
-            if sub_result.has_error():
-                return True
-
-        return False
+        return any(r.has_error() for r in self._sub_results.values())
 
     @property
     def errors(self) -> Sequence[str]:
@@ -227,31 +223,31 @@ class ValidationResult:
     def __str__(self) -> str:
         output: list[str] = []
 
-        self._create_error_string(output, path=[])
+        self._create_error_string(output, field_path=[])
 
         return " ".join(output)
 
-    def _create_error_string(self, output: list[str], path: list[str]) -> None:
+    def _create_error_string(self, output: list[str], field_path: list[str]) -> None:
         s = " ".join(self._errors)
         if s:
-            if path:
-                pathname = self._build_pathname(path)
+            if field_path:
+                pathname = self._build_pathname(field_path)
 
                 output.append(f"`{pathname}` is not valid: {s}")
             else:
                 output.append(s)
 
         for field, result in self._sub_results.items():
-            path.append(field)
+            field_path.append(field)
 
-            result._create_error_string(output, path)
+            result._create_error_string(output, field_path)
 
-            path.pop()
+            field_path.pop()
 
-    def _build_pathname(self, path: list[str]) -> str:
-        segments = [path[0]]
+    def _build_pathname(self, field_path: list[str]) -> str:
+        segments = [field_path[0]]
 
-        for p in path[1:]:
+        for p in field_path[1:]:
             if not p.startswith("[") or not p.endswith("]"):
                 segments.append(".")
 


src/fairseq2/utils/validation.py
```diff
@@ -4,10 +4,95 @@
...
 
+"""
+This module provides a set of helper types for object state validation. These
+types are primarly used by recipe and asset configuration dataclasses to ensure
+that all options are set correctly.
+
+A class -*typically a configuration dataclass*- that wants to support validation
+should expose a ``validate(self) -> ValidationResult`` method. Optionally,
+the class can derive from the runtime-checkable :class:`Validatable` protocol
+to make its intend more clear.
```
cirquit:
```suggestion
to make its intent more clear.
```

src/fairseq2/utils/validation.py
```diff
@@ -44,28 +145,63 @@ def _do_validate(self, obj: object) -> ValidationResult:
...
+            name of the field, for a ``Mapping`` it is the name of the key
+            formatted as ``f"[{repr(key)}]"``, for ``list``, ``Set``, and ``tuple``
+            it is the index of the value formatted as ``f"[{index}]"``.
+        :param result: The validation result of the sub-object.
+        """
         self._sub_results[field] = result
 
     def has_error(self) -> bool:
+        """
+        Returns ``True`` if the object or any of its sub-ojects have a
```
cirquit:
```suggestion
        Returns ``True`` if the object or any of its sub-objects have a
```

src/fairseq2/utils/validation.py
```diff
@@ -17,12 +102,28 @@
 
 
 class ObjectValidator(ABC):
+    """
+    Validates an object along with its sub-objects if it is a composite object
+    (i.e. a ``dataclass``, ``list``, ``Mapping``, ``Set``, or ``tuple``) and
```
cirquit, line 107: Why are we not targeting `collections.Iterable` instead of this subset?

src/fairseq2/utils/validation.py
cirquit, line 1: Nit: Can't comment on this line (https://github.com/facebookresearch/fairseq2/blob/067a03b1ac8f06090f85521313c5af2c699db6fc/src/fairseq2/utils/validation.py#L208C9-L210C28), but this could be refactored into a more functional form. Feel free to ignore if you prefer the clarity of the current code. 

```python
return any([r.has_error() for r in self._sub_results.values()])
```

tests/unit/utils/test_validation.py
```diff
@@ -0,0 +1,141 @@
...
+        )
+
+        validator = StandardObjectValidator()
+
+        with pytest.raises(
+            ValidationError, match=r"^root is not valid\. `field2\['foo2_2'\]` is not valid: value2_2 error\. `field3\[0\]` is not valid: value3_1 error\. `field4.field4_2\[0\]` is not valid: value4_2 error\.$"  # fmt: skip
+        ):
+            validator.validate(config)
+
+    def test_validate_works_when_root_has_no_error(self) -> None:
```
cirquit, line 81: This test fails for me with:
```bash
E       Failed: DID NOT RAISE <class 'fairseq2.utils.validation.ValidationError'>
```

tests/unit/utils/test_validation.py
```diff
@@ -0,0 +1,141 @@
...
+            },
+            field3=(FooSubConfig("value3_1 error."), FooSubConfig()),
+            field4=FooComplexSubConfig(
+                field4_1=3, field4_2=[FooSubConfig("value4_2 error.")], field4_3="xyz"
+            ),
+        )
+
+        validator = StandardObjectValidator()
+
+        with pytest.raises(
```
cirquit, line 76: This test fails for me with:
```bash
E       AssertionError: Regex pattern did not match.
E        Regex: "^root is not valid\\. `field2\\['foo2_2'\\]` is not valid: value2_2 error\\. `field3\\[0\\]` is not valid: value3_1 error\\. `field4.field4_2\\[0\\]` is not valid: value4_2 error\\.$"
E        Input: 'root is not valid.'
```

Looks like the cascaded accumulation of errors is not working.

src/fairseq2/utils/validation.py
```diff
@@ -77,41 +213,68 @@ def has_error(self) -> bool:
...
+        output: list[str] = []
 
-        self._create_error_str(parts, path=[])
+        self._create_error_string(output, path=[])
 
-        return " ".join(parts)
+        return " ".join(output)
 
-    def _create_error_str(self, parts: list[str], path: list[str]) -> None:
+    def _create_error_string(self, output: list[str], path: list[str]) -> None:
```
cirquit: I don't understand what path means here specifically (also don't understand it holistically, is this to show where the config lives?) . Its only called in [L230](https://github.com/facebookresearch/fairseq2/pull/1328/files#diff-f9d24c0482675e7918e0e2f94a692b68414907050e9f1a72f70ad4e215071e49R230) with an empty list and is nowhere used in the tests.

tests/unit/utils/test_validation.py
cbalioglu: That is very surprising. Are you sure you have not edited any code in your local branch?

src/fairseq2/utils/validation.py
cbalioglu: `path` is the path to the field in the object structure. This is tested in `pytest.raises` statements where you have `field2['foo2_2']`,  `field4.field4_2[0]`.

src/fairseq2/utils/validation.py
cbalioglu: That is an annoying limitation in Python since too many things are `Iterable`. For instance `str`, `bytes`, `memoryview` are all iterable types and we certainly do not want to iterate over them. There has been several posts about how to "fix" this in the official Python discussion group but I don't think anyone has written a PEP to address this. One example here: https://discuss.python.org/t/collections-abc-nonstringsequence/53274

src/fairseq2/utils/validation.py
cbalioglu: yeah, let me use a comprehension here. I think it is still readable enough. Thanks for the suggestion.

src/fairseq2/utils/validation.py
cirquit: Ah, this makes sense!

tests/unit/utils/test_validation.py
cirquit: Argh, sorry for this false positive, I was in an environment without an editable installation of fs2, so it likely resolved some functioncalls of 0.5.2 while using the 0.6 test code. Passes without issues now!

tests/unit/utils/test_validation.py
cirquit: My error, works if tested in the correct environment 🙈

src/fairseq2/utils/validation.py
cirquit: Ah, now I get the naming logic and why there are not explicit tests. 👍

PR #1332, opened 2025-09-24
Introduce support for optimizer parameter groups

cbalioglu:
This PR introduces support for optimizer parameter groups in recipes. Both first-party optimizers AdamW and Adafactor are updated to use this new feature. Users can now set optimizer options at parameter level whenever needed (e.g. GAN training). For instance to use a higher learning rate for attention layers in an encoder-decoder model, one can specify the following YAML:

```yaml
optimizer:
  name: adamw
  config:
    lr: 0.0003  # default, top-level LR
    betas:
    - 0.9
    - 0.99
    groups:
      # override the learning rate for encoder attention layers
    - params: # either a string or an array of strings containing regexes to match parameter names.
      - "encoder.layers.[0-9]+.self_attn.*"
      - "encoder.layers.[0-9]+.encoder_decoder_attn.*"
      lr: 0.0008
      # override the learning rate for decoder attention layers
    - params: "decoder.layers.[0-9]+.self_attn.*"
      lr: 0.0006
      betas: [0.9, 0.97]  # for decoder, override betas as well
 ```

Note that each parameter will be assigned to the first group in the list that matches its name. Parameters that do not match any group will use the top-level configuration. For instance to set a custom LR for the decoder stack of a model and yet another LR for decoder FFNs, one can have the following YAML:

```yaml
optimizer:
  name: adamw
  config:
    lr: 0.0003  # default, top-level LR
    betas:
    - 0.9
    - 0.99
    groups:
    - params: "decoder.layers.ffn.*"  # note that order matters. This should come before "decoder.*"
      lr: 0.0006
    - params: "decoder.*
      lr: 0.0004
 ```

During recipe startup, if any parameter groups are specified, all groups along with their parameters will be logged. This is helpful for users to verify that parameters are assigned to the right groups.

Also note that LR schedulers already support multiple parameter groups as of #1320. When combined together, they can be used as follows:

```yaml
optimizer:
  name: adamw
  config:
    groups:
    - params: ".*ffn.*"
      lr: 0.002
    - params: ".*"
      lr: 0.003
lr_scheduler:
  name: cosine-annealing
  config:
    start_lr: [0.0001, 0.0002]
    final_lr_scale: 0.2  # if scalar, shared by all parameter groups
```

### DIFF
--- a/src/fairseq2/recipe/config.py
+++ b/src/fairseq2/recipe/config.py
@@ -59,8 +59,36 @@ class SupportsStructure(ABC):
 Default: TypeAlias = Literal["default"]
 """
 Indicates that the default value of a particular configuration field, as defined
-in a higher-level configuration, should be used. See :class:`AdamWGroupConfig`
+in a higher-level configuration, can be used. See :class:`AdamWGroupConfig`
 as an example.
+"""
+
+default: Final = "default"
+"""
+The singleton sentinel value assignable to a field of type ``Default``
+indicating that the default value must be used.
+
+.. code:: python
+
+    from dataclasses import dataclass, field
+
+    from fairseq2.recipe.config import Default, default
+
+    @dataclass(kw_only=True)
+    class MyOptimizerConfig:
+        lr: float = 0.01
+        \"\"\"The top-level, default learning rate.\"\"\"
+
+        groups: list[MyOptimizerGroupConfig] = field(default_factory=list)
+        \"\"\"The parameter groups.\"\"\"
+
+    @dataclass(kw_only=True)
+    class MyOptimizerGroupConfig:
+        lr: float | Default = default
+        \"\"\"
+        The learning rate of the parameter group. If ``default``, uses 0.01 as
+        specified in ``MyOptimizerConfig``.
+        \"\"\"
 """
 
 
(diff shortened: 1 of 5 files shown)


src/fairseq2/recipe/optim.py
```diff
@@ -5,17 +5,242 @@
...
+
+    @property
+    def param_names(self) -> Iterable[str]:
+        """Gets the names of the parameters belonging to this group."""
+        return self._param_names
+
+    @property
+    def empty(self) -> bool:
+        """Gets whether this group has no parameters."""
+        return bool(self._params)
```
zyaoj: This seems to be a logical error to me? Should be:
```suggestion
        return not bool(self._params)
```

src/fairseq2/recipe/optim.py
```diff
@@ -5,17 +5,242 @@
...
+                break
+
+    output: list[dict[str, object]] = []
+
+    for idx, group in enumerate(groups):
+        if not group.empty:
+            if group.is_fallback:
+                continue
+
+            log.warning("Optimizer parameter group {} is empty.", idx)
```
zyaoj, line 172: not sure if this was intended, but I see contradictory logic here -- `if not group.empty` then warning `group is empty`?

src/fairseq2/recipe/config.py
```diff
@@ -773,11 +820,51 @@ class AdafactorConfig:
...
+    The regular expression(s) used to select the parameters that belong to this
+    group.
+    """
+
+    lr: float | Default = "default"
+    """The learning rate."""
+
+    beta2_decay: float | Default = "default"
+    """
+    The decay rate of beta2. beta2 standardly refers to the coefficient used for
```
zyaoj: Similar to another comment for `beta2_decay`

```suggestion
    The decay rate of beta2. beta2 refers to the coefficient used for computing
    the running average of the gradient squared.
```

zyaoj (approved):
Overall looks good to me! A couple of nits and should be good to go! Thanks!

PR #1339, opened 2025-09-30
more verbosity in structured errors

uralik:

when config can not be structured, its useful to see what objects it get stuck on, added more verbosity in the messages in 3 places which were enough in my specific case


cbalioglu (approved):
LGTM with the latest improvements.

PR #1340, opened 2025-09-30
Add family to ModelArchitectureNotKnownError

cbalioglu:
This nit PR introduces a new `family` parameter to `ModelArchitectureNotKnownError` to improve error messaging. We also use the deprecation notice feature for the first time.

### DIFF
--- a/src/fairseq2/models/hub.py
+++ b/src/fairseq2/models/hub.py
@@ -6,6 +6,7 @@
 
 from __future__ import annotations
 
+import warnings
 from collections.abc import Iterator
 from pathlib import Path
 from typing import Generic, TypeVar, cast, final
@@ -43,7 +44,7 @@ class ModelHub(Generic[ModelT, ModelConfigT]):
     def get_arch_config(self, arch: str) -> ModelConfigT:
         config = self.maybe_get_arch_config(arch)
         if config is None:
-            raise ModelArchitectureNotKnownError(arch)
+            raise ModelArchitectureNotKnownError(arch, self._family.name)
 
         return config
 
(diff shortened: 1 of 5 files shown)


src/fairseq2/recipe/run.py
```diff
@@ -160,6 +167,10 @@ def create_output_dir(resolver: DependencyResolver) -> Path:
     container.register(Path, create_output_dir)
 
 
+def _setup_warnings() -> None:
+    warnings.filterwarnings("once", category=DeprecationWarning, module="fairseq2.*")
```
cirquit: Is this going to work only in recipes and not library-wide? I'm thinking about the test suites that do not test recipes and are likely the first way of how people find out about deprecation warnings.

src/fairseq2/models/hub.py
```diff
@@ -214,10 +215,22 @@ def __init__(self, name: str) -> None:
...
-    def __init__(self, arch: str) -> None:
-        super().__init__(f"{arch} is not a known model architecture.")
+    def __init__(self, arch: str, family: str | None = None) -> None:
+        """
+        ``family`` defaults to ``None`` due to backwards-compatibility. New code
+        must specify a model family when raising this error.
+        """
+        if family is None:
+            warnings.warn(
+                "`ModelArchitectureNotKnownError` will require a `family` argument starting fairseq2 v0.12.", DeprecationWarning  # fmt: skip
```
cirquit: It would be nice to have a consolidated way to call `warnings.warn` with either a `warning_type=DeprecationWarning` or a default name `warnings.warn_deprecation`. 

This minimized imports and helps with downstream changes to this warning name. If we expose it visibly to users it will need to be taken care of during upgrading, which I'd prefer not to do.

src/fairseq2/recipe/run.py
cbalioglu: Yes, that is one thing that I also find annoying, but apparently Python does not expose an easy way to turn on `DeprecationWarning`s by default (as of Python 3.2 it is off by default). One workaround would be to use `UserWarning` instead, but that warning is conventionally meant to be used to warn about user errors instead of deprecation notices.

Official Python documentation advises "developers of test runners" to always enable all warnings according to the second paragraph of [this section](https://docs.python.org/3/library/warnings.html#overriding-the-default-filter). The following paragraph in that same section is what I followed as "the best practice" they recommend.

I would actually prefer the opposite behavior from Python folks. It should be easy to turn off when needed, but ideally all warnings should be enabled by default. This makes it tricky because we can never be 100% sure if our users see our deprecation warnings.

src/fairseq2/models/hub.py
cbalioglu: Sure, I can wrap `warnings.warn(category=DeprecationMessage)` into a convenience function.

cbalioglu:
@cirquit tried to address your feedback, would appreciate a quick second look before I merge it.

src/fairseq2/recipe/run.py
cirquit: I think given the situation this is likely the best solution for now. Lets not forget to add this warning-enabling code as part of our own test-suite to lead with a good example.

PR #1349, opened 2025-10-03
Revise shard dim handling in model checkpoint loaders

cbalioglu:
This PR includes several closely related changes:

(1) Introduces a new and simpler `shard_dims: dict[str, int]` parameter in `ModelCheckpointLoader` that works at parameter granularity to overcome the limitation of `shard_specs`' module-level handling. We maintain full backwards compatibility. The existing `shard_specs` parameter will be removed in fairseq2 v0.12 (in about 6 months). Till then, both parameters will function as documented, but we will print a warning/deprecation message if `shard_specs` is specified instead of `shard_dims`.
(2) Introduces a new `fairseq2.nn.Sharded` interface and a standalone helper `get_shard_dims()` function. As its name suggests, `Sharded` interface marks a class as sharded and is implemented by `ColumnShardedLinear`, `RowShardedLinear`, `VocabShardedEmbedding` and `ShardedEmbedding`. It exposes only a single method called `get_shard_dims()` that returns a list of `(parameter, shard_dim)` tuples indicating which parameters/buffers of the module are sharded along which dimension. The standalone `get_shard_dims()` accepts a PyTorch module, traverses its submodules, and for each `Sharded` module it finds, adds its sharded parameters' fully-qualified name along with the dimension information to a dictionary that it returns at the end of the call. The returned dictionary can be passed to `ModelCheckpointLoader` to specify the sharding information. The advantage of this approach is two-fold: (1) it operates at the parameter level (2) it is dynamically extracted at runtime and does not require the model author to manually create a "shard spec".
(3) Updates `StandardModelFamily` to support the new `shard_dims` feature in a backwards compatible way. If a model family has an explicit `shard_spec`, things work as in the past. If no `shard_spec` is defined, the implementation attempts to dynamically extract the sharding information via `get_shard_dims()`. With v0.12, we will remove the `shard_spec` implementation.
(4) Revises and/or significantly extends the documentation of projection modules, embedding modules, as well as model checkpoint APIs.
(5) Squeezes a small extra feature in `RowShardedLinear` and introduces a `reduce_output` option which is required for MoE models. This feature was taken from Martin's LLaMA-4 Scout PR.

This PR will be accompanied by a quick follow-up PR that also revises/simplifies sharded model loading that will entirely deprecate the existing `ModelSharder` API.

### DIFF
--- a/src/fairseq2/model_checkpoint/basic.py
+++ b/src/fairseq2/model_checkpoint/basic.py
@@ -28,6 +28,8 @@ from fairseq2.sharder import ShardSpec
 
 @final
 class BasicModelCheckpointLoader(ModelCheckpointLoader):
+    """Loads single-file PyTorch checkpoints (.pt, .pth, .bin)."""
+
     def __init__(self, file_system: FileSystem, tensor_loader: TensorLoader) -> None:
         self._file_system = file_system
         self._tensor_loader = tensor_loader
@@ -42,6 +44,7 @@ class BasicModelCheckpointLoader(ModelCheckpointLoader):
         restrict: bool = True,
         state_dict_converter: StateDictConverter | None = None,
         shard_specs: Mapping[str, ShardSpec] | None = None,
+        shard_dims: Mapping[str, int] | None = None,
     ) -> Iterator[tuple[str, Tensor]]:
         try:
             checkpoint = self._tensor_loader.load(
@@ -82,6 +85,7 @@ class BasicModelCheckpointLoader(ModelCheckpointLoader):
                 target_shard_sizes,
                 target_shard_ranks,
                 shard_specs,
+                shard_dims,
             )
 
             yield key, tensor
--- a/src/fairseq2/model_checkpoint/common.py
+++ b/src/fairseq2/model_checkpoint/common.py
@@ -53,7 +53,7 @@ def reshard_tensor(
 
     ``source_shard_sizes`` specifies the distributed source configuration in the
     form of ``(tp_size, dp_size)``. Similarly, ``target_shard_sizes`` specifies
-    the target configuration as ``(tp_size, dp_size)``.
+    the target configuration.
 
     ``target_shard_ranks`` specifies the ranks of the current process in the
     target configuration in the form of ``(tp_rank, dp_rank)``.
@@ -141,23 +141,31 @@ def reshard_tensor(
         tp_splits.append(tp_split)
 
     # Reshard the tensor over the target parallel dimension.
+
+    # Source shard dimensions
     source_tp_dim_size = tp_splits[0].size(tp_dim)
 
+    # Total unsharded dimension size
     tp_dim_size = source_tp_dim_size * source_tp_size
 
+    # Target shard dimension
     target_tp_dim_size = tp_dim_size // target_tp_size
 
-    f_target_idx = target_tp_rank * target_tp_dim_size
-    l_target_idx = target_tp_rank * target_tp_dim_size + target_tp_dim_size - 1
+    # Slice boundaries for this target rank.
+    first_target_idx = target_tp_rank * target_tp_dim_size
+    last_target_idx = target_tp_rank * target_tp_dim_size + target_tp_dim_size - 1
 
-    f_source_tp_shard_idx = f_target_idx // source_tp_dim_size
-    l_source_tp_shard_idx = l_target_idx // source_tp_dim_size
+    # Determine which source ranks contain the target slice.
+    first_source_tp_shard_idx = first_target_idx // source_tp_dim_size
+    last_source_tp_shard_idx = last_target_idx // source_tp_dim_size
 
-    f_source_idx = f_source_tp_shard_idx * source_tp_dim_size
+    # Starting index of the first relevant source shard
+    first_source_idx = first_source_tp_shard_idx * source_tp_dim_size
 
+    # Collect sub-slices from relevant source shards.
     tp_sub_splits = []
 
-    for idx in range(f_source_tp_shard_idx, l_source_tp_shard_idx + 1):
+    for idx in range(first_source_tp_shard_idx, last_source_tp_shard_idx + 1):
         tp_sub_splits.append(tp_splits[idx])
 
     del tp_splits
@@ -166,8 +174,9 @@ def reshard_tensor(
 
     del tp_sub_splits
 
+    # Extract exact slice needed for this target rank.
     return tensor.narrow(
-        dim=tp_dim, start=f_target_idx - f_source_idx, length=target_tp_dim_size
+        dim=tp_dim, start=first_target_idx - first_source_idx, length=target_tp_dim_size
     )
 
 
--- a/src/fairseq2/model_checkpoint/common.py
+++ b/src/fairseq2/model_checkpoint/common.py
@@ -29,11 +29,14 @@ def reshard_tensor(
     Reshards a parameter tensor from a distributed source configuration to a
     target configuration.
 
-    This function handles the complex task of resharding tensors when loading
-    checkpoints from one distributed configuration (e.g., 4-way tensor parallelism)
-    to a different target configuration (e.g., 8-way tensor parallelism). It
-    efficiently concatenates and slices tensors to produce the correct shards
-    for the target rank.
+    This function is meant for authors of new :class:`ModelCheckpointLoader`
+    implementations. It handles the complex task of resharding tensors when
+    loading checkpoints from one distributed configuration (e.g. 4-way tensor
+    parallelism) to a different target configuration (e.g. 8-way tensor
+    parallelism). It efficiently concatenates and slices tensors to produce the
+    correct shards for the target rank. The existing implementations such as
+    :class:`NativeModelCheckpointLoader` may be inspected to see how ``reshard_tensor``
+    is used in practice.
 
     The resharding process involves:
 
--- a/src/fairseq2/model_checkpoint/common.py
+++ b/src/fairseq2/model_checkpoint/common.py
@@ -26,8 +26,8 @@ def reshard_tensor(
     shard_dims: Mapping[str, int] | None = None,
 ) -> Tensor:
     """
-    Reshards a parameter from a distributed source configuration to a target
-    configuration.
+    Reshards a parameter tensor from a distributed source configuration to a
+    target configuration.
 
     This function handles the complex task of resharding tensors when loading
     checkpoints from one distributed configuration (e.g., 4-way tensor parallelism)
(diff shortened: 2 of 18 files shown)


src/fairseq2/model_checkpoint/basic.py
cirquit, line 74: When does this happen? I thought that the checkpoint items would be unique and are not sharded yet here, so there should be no duplication?

src/fairseq2/model_checkpoint/common.py
```diff
@@ -22,7 +23,86 @@ def reshard_tensor(
...
+
+    ``target_shard_ranks`` specifies the ranks of the current process in the
+    target configuration in the form of ``(tp_rank, dp_rank)``.
+
+    ``shard_dims`` is the mapping from parameter names to dimensions along which
+    parameters should be sharded for tensor parallelism. Omitted for replicated
+    tensors. See :func:`~fairseq2.nn.get_sharding_dims` for more information.
+
+    ``shard_specs`` is deprecated and will be removed in a future release;
+    please use ``shard_dims`` instead.
```
cirquit, line 69: Is there some value in providing some help on how to migrate to `shard_dims` from `shard_spec` here? Or should this be its own post later?

src/fairseq2/model_checkpoint/common.py
cirquit: Nit: Is this common to use `f` and `l` as an abbreviation when working with indices? I understood that you mean "first" and "last" after grokking the code, but I would've used "start" and "end". When I initially glanced over it I did not get that.

src/fairseq2/model_checkpoint/common.py
```diff
@@ -91,7 +171,16 @@ def reshard_tensor(
     )
```
cirquit, line 183: While I love how terse this code is for what it does, I think we can improve readability a bit without needing to parse the parameter names. 

```suggestion
# Reshard the tensor over the target parallel dimension.

# Source shard dimensions
source_tp_dim_size = tp_splits[0].size(tp_dim)

# Total unsharded dimension size
tp_dim_size = source_tp_dim_size * source_tp_size

# Target shard dimensions 
target_tp_dim_size = tp_dim_size // target_tp_size

# Slice boundaries for this target rank
f_target_idx = target_tp_rank * target_tp_dim_size
l_target_idx = target_tp_rank * target_tp_dim_size + target_tp_dim_size - 1

# Which source shards contain our target slice
f_source_tp_shard_idx = f_target_idx // source_tp_dim_size
l_source_tp_shard_idx = l_target_idx // source_tp_dim_size

# Starting index of the first relevant source shard
f_source_idx = f_source_tp_shard_idx * source_tp_dim_size

# Collect sub-slices from relevant source shards
tp_sub_splits = []

for idx in range(f_source_tp_shard_idx, l_source_tp_shard_idx + 1):
    tp_sub_splits.append(tp_splits[idx])

del tp_splits

tensor = torch.cat(tp_sub_splits, dim=tp_dim)

del tp_sub_splits

# Extract exact slice needed for this target rank
return tensor.narrow(
dim=tp_dim, start=f_target_idx - f_source_idx, length=target_tp_dim_size
)
```

src/fairseq2/model_checkpoint/safetensors.py
```diff
@@ -29,6 +29,13 @@
 
 @final
 class SafetensorsCheckpointLoader(ModelCheckpointLoader):
+    """
+    Loads checkpoints saved in Safetensors format.
+
+    This loader supports both single-file and directory-based Safetensors
```
cirquit: Nit:  Can we give a quick example here just to be very clear that this is only a single inner-directory walk? E.g. ./{dir-name}/shard_{idx}.pt

Just saw that we can trivially use `_file_system.rglob` here as well, is there any issue in doing this for recursive traversing? I can imagine this to be an edge case, but it would be nice if this would work out of the box.

src/fairseq2/models/family.py
```diff
@@ -452,12 +458,27 @@ def iter_checkpoint(
...
 
-        return self._do_iter_checkpoint(path, config, gangs, mmap, restrict)
+        if not self._supports_meta:
+            _warn_deprecated(
+                "In fairseq2 v0.12, `ModelFamily.iter_checkpoint` will stop supporting models that cannot be meta-initialized."
+            )
+
+        if self._shard_specs is None:
+            model = self._do_create_model(
+                config, gangs, torch.float16, self._supports_meta
```
cirquit, line 471: Where does the default `torch.float16` come from? Does this correctly handle the case of normalization layers that should be kept at fp32?

cirquit (commented):
Mostly nits!

I'm thinking about how I'd like an optimal sharded loading solution would look like for the future and if this implementation allows for this upgrade.

IMO, the best (simple) case scenario for sharded model loading is that an entire DP rank loads the complete model (while respecting local TP shards) and then distributes it to the other DP ranks to make use of the faster interconnect vs. reading from hard drives.

I think this kind of hierarchical distribution is not possible with the current setup because we only work from a single GPU perspective and only care about our own shard, regardless of who else might need it, right? Not sure if you already thought about this and found it to be a worse solution to what we currently have.

src/fairseq2/model_checkpoint/basic.py
cbalioglu: This is about *shared* tensors rather than *sharded* tensors. PyTorch tensor files natively support deserializing a tensor multiple times under different parameter names while preserving the same underlying storage (safetensors does not support this feature). This is in particular used for tied weights (e.g. embedding and final projection using the same weights in vanilla transformer). It is a nice feature that unfortunately safetensors lacks. They even have a guide on how to accomplish the same thing in safetensors [here](https://huggingface.co/docs/safetensors/en/torch_shared_tensors), but it is not really very satisfactory. Here we make sure that we yield such tensors only once since assigning them to one of the parameters implicitly assigns them to others as well.

src/fairseq2/model_checkpoint/common.py
cbalioglu: Yeah, that is certainly something I consider doing. It is just a cyclic dependency. We first have to merge this to have a post/announcement. Once we have it, I plan to update these deprecation notices with the URL of it.

src/fairseq2/model_checkpoint/common.py
cbalioglu: Not really, it is just mean being too "spartan" here. I can certainly use "start" and "end" as prefixes

src/fairseq2/model_checkpoint/common.py
cbalioglu: Neat! Thanks a lot! Let me apply this.

src/fairseq2/model_checkpoint/safetensors.py
cbalioglu: Ok, I can mention that this is one-level walk only. The reason is because safetensors checkpoints are structured in two forms. Either a single file, or a one-level directory with the convention "model-X-of-N.safetensors". For instance see [here](https://huggingface.co/Qwen/Qwen3-Next-80B-A3B-Instruct/tree/main).

src/fairseq2/models/family.py
cbalioglu: This is purely a "dummy" model initialization. Ideally if the model supports meta-initialization, it is almost instantaneous since there is no memory allocation happening. We use it solely to extract the parameter-shard mappings, so we don't really care about what the dtype we use here.

Having said that, in our module implementations, if we know that a parameter/buffer has to have a specific data type, we always explicitly specify its data type and do not allow it to be overriden. For instance, most sinusoidal positional encoders require fp32 to avoid numerical issues, therefore, we initialize them directly in fp32 like [here](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/nn/position_encoder.py#L97-L99).

cbalioglu:
> Mostly nits!
> 
> I'm thinking about how I'd like an optimal sharded loading solution would look like for the future and if this implementation allows for this upgrade.
> 
> IMO, the best (simple) case scenario for sharded model loading is that an entire DP rank loads the complete model (while respecting local TP shards) and then distributes it to the other DP ranks to make use of the faster interconnect vs. reading from hard drives.
> 
> I think this kind of hierarchical distribution is not possible with the current setup because we only work from a single GPU perspective and only care about our own shard, regardless of who else might need it, right? Not sure if you already thought about this and found it to be a worse solution to what we currently have.

This is actually what we do right now. In tp/dp gang topology, you can think of every model shard to be in its own data parallel gang. Within a data parallel gang, we actually initialize (or load) a model only on the first rank, while we meta-initialize the model on all other ranks. During FSDP initialization, we broadcast parameters and persistent buffers from rank0 to other ranks while doing on-the-fly initialization.

You can check out the recipe model loading logic [here](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/recipe/internal/model.py#L152-L186). In the same file, we also have the logic for model initialization a bit below. This combined with the DDP/FSDP broadcasting we have [here](https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/nn/fsdp/fsdp2.py#L142-L146) for FSDP2 (We have similar logic for FSDP1 and DDP) gives the behavior that you described I believe.

src/fairseq2/model_checkpoint/safetensors.py
cirquit: Yea, this is also what I typically saw! Just wanted to minimize misunderstandings of the recent issue with the nested checkpoint loader.

src/fairseq2/models/family.py
cirquit: Makes sense, but this looks a bit random when glancing over this code! Hopefully somebody will find this discussion when the need for refactoring arises (or just understands that this is not an important parameter).

cbalioglu: Ok added some inline comment to make it a bit more clear why we initialize a model here.

src/fairseq2/model_checkpoint/common.py
cbalioglu: Addressed in latest revision.

src/fairseq2/model_checkpoint/safetensors.py
cbalioglu: Addressed in latest revision.

src/fairseq2/model_checkpoint/common.py
cbalioglu: Applied

src/fairseq2/model_checkpoint/safetensors.py
cirquit: Looks super clear now!

zyaoj (approved):
LGTM! Thanks for the clean PR for a very solid goal! I had similar questions to Alexander's e.g. regarding the cost of initializing model to get the model `shard_dims` specs at parameter level. But the added comments look good.

PR #1351, opened 2025-10-06
Wav2Vec2 SSL/ASR recipes

cirquit:

Fixes #{159}


recipes/wav2vec2/asr/data/dataset.py
```diff
@@ -0,0 +1,435 @@
...
+    """The minimum audio sequence length."""
+
+    max_audio_len: int = 800_000
+    """The maximum audio sequence length."""
+
+    dtype: DataType = torch.float16
+    """Numerical precision for audio decoding. Overridden to `torch.float32` when ``config.normalize_audio = True``."""
+
+    # Batching parameters
+    batching_strategy: BatchingStrategy = BatchingStrategy.LENGTH
```
cbalioglu: In the past, when datasets were centralized and meant to be generically usable, having an option to set batch strategy made sense. Having said that, the "vanilla" wav2vec2 ASR finetuning has always used length-based batching. Unless you have more context from your collaboration with the internal teams to justify a static batching, I would suggest just to support length-based to reduce the implementation complexity.

recipes/wav2vec2/asr/data/dataset.py
```diff
@@ -0,0 +1,435 @@
...
+    """If `batching_strategy = BatchingStrategy.LENGTH, ignores `batch_size` and each batch will have
+    `<= max_num_elements` elements with `<= num_seqs_multipe_of` or `num_seqs_multiple_of * N` sequences.
+    This is primarily for hardware optimization, but will only work when sufficient enough sequences are available for bucketing.
+    """
+
+    # Audio processing
+    normalize_audio: bool = False
+    """If ``True``, normalizes audio to have zero mean and unit variance and uses torch.float32 during audio decoding (overriding ``config.dtype``)."""
+
+    no_padding: bool = True
```
cbalioglu: Similar to my feedback above, the vanilla ASR finetuning (and, as far as I can think of, any sort of ASR training), requires padding since you want to feed non-truncated sequences to the training.

recipes/wav2vec2/asr/data/dataset.py
```diff
@@ -0,0 +1,435 @@
...
+    """The data synchronization mode among processes."""
+
+    # Misc
+    max_num_batches: int | None = None
+    """The maximum number of batches to return."""
+
+    num_prefetch: int = 4
+    """The number of batches to prefetch in background."""
+
+    seed: int = 2
```
cbalioglu: We already have `common.seed` that we should use instead of this field.

recipes/wav2vec2/asr/data/dataset.py
```diff
@@ -0,0 +1,435 @@
...
+            with manifest_file.open(encoding="utf-8") as fp:
+                header = fp.readline().rstrip()
+        except OSError as ex:
+            raise DataReadError(
+                f"The {manifest_file} manifest file cannot be read. See the nested exception for details.",
+            ) from ex
+
+        try:
+            audio_dir = Path(header)
+            if audio_dir.exists():
```
cbalioglu: This if/else condition looks a bit peculiar to me. I think it would be cleaner this way:

```python
try:
  audio_dir = Path(header)
except ValueError:
  raise DatasetError(...) from None

if not audio_dir.is_dir():
  raise DataReadError(f"{audio_dir} pointed by the manifest file is not a directory.")

return audio_dir

File: recipes/wav2vec2/asr/data/dataset.py [hunk 5]
```diff
@@ -0,0 +1,435 @@
...
+        (...)
+        ```
+        """
+        tsv_file = self._manifest_dir.joinpath(f"{split}.tsv")
+
+        builder = read_text(
+            tsv_file,
+            rtrim=True,
+            memory_map=True,
+            block_size=10 * 1024 * 1024,
```
cbalioglu: Just curious, how did you come up with this block size? :)

recipes/wav2vec2/asr/data/dataset.py
```diff
@@ -0,0 +1,435 @@
...
+        )
+
+        if audio_dir is not None:
+            builder.skip(1)  # Path to the data directory
+
+        field_splitter = StrSplitter(names=["audio", "audio_size"])
+        builder.map(field_splitter)
+
+        # Convert audio_size to int and clamp to max_audio_len
+        builder.map(
```
cbalioglu: I think this map and the filter below it can be made slightly faster by:

```python
# fairseq2 DataPipeline has native support for primitive type
# conversions directly in C++
builder.map(int, selector="audio_size")

builder.filter(lambda e: e["audio_size"] >= min_audio_len and e["audio_size"] <= max_audio_len)
```

recipes/wav2vec2/asr/data/dataset.py
```diff
@@ -0,0 +1,435 @@
...
+    """The size of the sliding window for shuffling batches."""
+
+    # Batching behavior
+    drop_remainder: bool = False
+    """If ``True``, drops the last set of batches if they have in total fewer examples than requested."""
+
+    sync_batches: bool = True
+    """If ``True``, ensures that each process reads the same number of batches."""
+
+    sync_mode: SyncMode = SyncMode.UNTIL_FIRST
```
cbalioglu: I don't think it makes much sense to expose this as a configurable setting. Training should always be `UNTIL_FIRST` and eval/validation should always be `UNTIL_LAST`. Setting it otherwise would be a bug in both cases.

recipes/wav2vec2/asr/data/dataset.py
```diff
@@ -0,0 +1,435 @@
...
+    """The size of the sliding window for shuffling examples (pre-batch)."""
+
+    batch_shuffle_window: int = 1000
+    """The size of the sliding window for shuffling batches."""
+
+    # Batching behavior
+    drop_remainder: bool = False
+    """If ``True``, drops the last set of batches if they have in total fewer examples than requested."""
+
+    sync_batches: bool = True
```
cbalioglu: Maybe a warning message here indicating, that if this is set to `False`, it might cause hangs during training. It may be set to `False` only if the user is certain that the training will finish before reaching the end of the dataset. Or we can play safer and not expose this as an option at all.

recipes/wav2vec2/asr/data/dataset.py
```diff
@@ -0,0 +1,435 @@
...
+                "Collating without padding is currently not supported, defaulting to padding."
+            )
+
+        # Collating -> { audio.path: list(str), audio.data.sample_rate: list(int), audio.data.format: list(int),
+        #                audio.data.waveform.is_ragged: bool, audio.data.waveform.seqs: [tensor],
+        #                audio.data.waveform.seq_lens: list(int), audio_size: int,
+        #                text.is_ragged: bool, text.seqs: [tensor], text.seq_lens: list(int) }
+        builder = text_pipeline.collate_with_pad_ix(
+            builder, no_padding=False
+        )  # no_padding)
```
cbalioglu: Can be removed?

recipes/wav2vec2/asr/data/dataset.py
```diff
@@ -0,0 +1,435 @@
...
+    in the YAML asset definition:
+
+    ```yaml
+    name: mydataset
+    dataset_config:
+      data: (all keys here must have a companion parameter in this config)
+    ```
+    """
+
+    data: Path = field(default_factory=Path)
```
cbalioglu: Nit: maybe calling this `manifest_dir` makes it more readable now. In the past, `data` was the name of the asset field used by all datasets, but now we can use a more descriptive name rather than just "data".

recipes/wav2vec2/asr/data/preprocessing.py
```diff
@@ -0,0 +1,270 @@
...
+    tmask_len = random.randint(0, time_mask_param)
+    tmask_i = random.randint(0, (n_t - tmask_len - 1))
+
+    masked_spec = spec.clone()
+    masked_spec[..., tmask_i : (tmask_i + tmask_len)] = 0.0
+    return masked_spec
+
+
+@torch.no_grad()
+def apply_spec_augment(
```
cbalioglu: Is there any case where we apply a spectogram during wav2vec2 ASR finetuning? We already use masking as a regularization within the model itself, but I haven't seen the use of spectogram. At least in the vanilla/reference recipe in fairseq.

recipes/wav2vec2/asr/data/preprocessing.py
```diff
@@ -0,0 +1,270 @@
...
+        waveform = apply_spec_augment(
+            waveform,
+            freq_mask_param=spec_aug_freq_mask_param,
+            time_mask_param=spec_aug_time_mask_param,
+        )
+
+    return waveform.to(dtype)
+
+
+class AudioCropper:
```
cbalioglu: Same as above. Not sure if there is any cropping involved in ASR (and if it makes sense at all considering it is a seq2seq learning task)

recipes/wav2vec2/asr/data/preprocessing.py
```diff
@@ -0,0 +1,270 @@
...
+                item[self.audio_feature] = audio[start : start + crop_size]
+
+        return batch
+
+
+class AudioProcessingPipeline:
+    """Composable audio processing pipeline builder."""
+
+    @staticmethod
+    def add_path_resolution(
```
cbalioglu: Nit: this method name seems a bit too opaque to me. Would naming it `add_file_mapping` or something along the line make more sense?

recipes/wav2vec2/asr/eval/recipe.py
```diff
@@ -0,0 +1,142 @@
...
+            raise RecipeError(
+                "`dataset.valid_split` must be defined for evaluation but is `None`."
+            )
+
+        seed = config.common.seed
+
+        # Initialize validation units
+        eval_units = []
+        eval_data_readers = []
+        eval_splits = config.dataset.valid_split.split(",")
```
cbalioglu: Nit: We can either simplify this and just support one validation split (librispeech and librilight have one split for validation anyways), or if we want to support this more advanced variant, I would prefer `eval_splits` to be a `str | list[str]` instead of doing some string manipulations.

recipes/wav2vec2/asr/config.py
```diff
@@ -0,0 +1,119 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
cbalioglu, line 1: Why not just name this file `config.py` like other recipes? It implies as if there is another config dataclass that is not the default one :)

recipes/wav2vec2/asr/recipe.py
```diff
@@ -0,0 +1,305 @@
...
+        asr_module: Wav2Vec2AsrModel = cast(Wav2Vec2AsrModel, model.module)
+
+        if model.newly_initialized:
+            self.prepare_ctc_training_from_encoder(
+                context=context, asr_module=asr_module
+            )
+
+        # Make sure that the final projection layer is instantiated along with
+        # the pretrained parameters if it was on the meta device.
+        if context.gangs.dp.rank == 0:
```
cbalioglu: This if statement is not needed anymore in the new recipe architecture. You can safely remove this.

recipes/wav2vec2/asr/recipe.py
```diff
@@ -0,0 +1,305 @@
...
+            no_padding=config.dataset.no_padding,
+            npc=config.dataset.npc,
+            # Shuffling and performance parameters
+            example_shuffle_window=config.dataset.example_shuffle_window,
+            batch_shuffle_window=config.dataset.batch_shuffle_window,
+            num_accumulate=config.trainer.grad_accumulation.num_batches,
+            num_prefetch=config.dataset.num_prefetch,
+            drop_remainder=config.dataset.drop_remainder,
+            sync_batches=config.dataset.sync_batches,
+            sync_mode=config.dataset.sync_mode,
```
cbalioglu: As I mentioned above, this is dangerous and should not be really settable by the user.

recipes/wav2vec2/ssl/data/dataset.py
```diff
@@ -0,0 +1,415 @@
...
+
+    # Upsampling
+    beta_corpus: float | None = None
+    """Corpus sampling temperature; between [0,1]."""
+
+    beta_language: float | None = None
+    """Language sampling temperature; between [0,1]."""
+
+    # SpecAugment
+    spec_aug_p: float | None = None
```
cbalioglu: Same feedback as above for ASR. At least for the vanilla wav2vec2 pretraining, masking acts a similar role as a spectogram. Not sure in which setting this is used.

recipes/wav2vec2/ssl/data/dataset.py
```diff
@@ -0,0 +1,415 @@
...
+    no_padding: bool = True
+    """If ``True``, all elements in the batch will be truncated to by batch minimal length.
+    Therefore, no padding will be applied to the batch.
+    """
+
+    npc: int = 10
+    """The number of parallel calls to use in the pipeline."""
+
+    # Upsampling
+    beta_corpus: float | None = None
```
cbalioglu: What exactly are these two parameters? I assume these are more relevant for the internal teams. There is no multi-linguality or corpus in vanilla wav2vec2 pretraining.

recipes/wav2vec2/ssl/data/dataset.py
```diff
@@ -0,0 +1,415 @@
...
+    max_num_elements: int = 1_500_000
+    """If `batching_strategy = BatchingStrategy.LENGTH, ignores `batch_size` and each batch will have
+    `<= max_num_elements` elements with `<= num_seqs_multipe_of` or `num_seqs_multiple_of * N` sequences.
+    This is primarily for hardware optimization, but will only work when sufficient enough sequences are available for bucketing.
+    """
+
+    normalize_audio: bool = False
+    """If ``True``, normalizes audio to have zero mean and unit variance and uses torch.float32 during audio decoding (overriding ``config.dtype``)."""
+
+    use_fbank: bool = False
```
cbalioglu: This is likely relevant for the internal team. Original wav2vec2 supports only waveform as input.

recipes/wav2vec2/ssl/data/dataset.py
```diff
@@ -0,0 +1,415 @@
...
+    """If ``True``, use fbank features instead of waveform."""
+
+    no_padding: bool = True
```
cbalioglu: As far as I remember, during pretraining we were always cropping each batch to the shortest sequence in the batch, so there was never any padding (unlike ASR). Is this needed for internal teams?

recipes/wav2vec2/ssl/data/dataset.py
```diff
@@ -0,0 +1,415 @@
...
+
+    cached_fd_count: int = 1000
+    """Enables an LRU cache on the last ``cached_fd_count`` files read.
+    ``FileMapper`` will memory map all the cached file, so this is especially
+    useful for reading several slices of the same file.
+    """
+
+
+@final
+class Wav2Vec2SslDataset:
```
cbalioglu, line 109: Looks like we have some extra fields in the dataset configuration that were not really used in the original pretraining recipe. I am fine with keeping them if they are really used by some of our teams, but otherwise, my suggestion would be to simplify this dataset implementation to match the (much more basic) needs of the vanilla pretraining. Just single-channel plain waveform and always cropped.

recipes/wav2vec2/ssl/data/dataset.py
```diff
@@ -0,0 +1,415 @@
...
+    in the YAML asset definition:
+
+    ```yaml
+    name: mydataset
+    dataset_config:
+      data: (all keys here must have a companion parameter in this config)
+    ```
+    """
+
+    data: Path = field(default_factory=Path)
```
cbalioglu: Same as ASR dataset, naming this `manifest_dir` would be more descriptive.

recipes/wav2vec2/asr/recipe.py
```diff
@@ -0,0 +1,305 @@
...
+        return self._criterion(batch, metric_bag)
+
+    @property
+    @override
+    def model(self) -> RecipeModel:
+        return self._criterion.model
+
+
+@final
+class Wav2Vec2AsrEvalUnit(EvalUnit[Seq2SeqBatch]):
```
cbalioglu, line 258: Why don't we just use the identical implementation already residing in eval/recipe.py?

recipes/wav2vec2/ssl/eval/config.py
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
```
cbalioglu, line 1: Same as ASR config. I think it would make more sense to name this file just "config.py".

recipes/wav2vec2/ssl/recipe.py
```diff
@@ -0,0 +1,199 @@
...
+            max_num_batches=config.dataset.max_num_batches,
+            cached_fd_count=config.dataset.cached_fd_count,
+        )
+
+        valid_units = []
+        valid_data_readers = []
+
+        if config.dataset.valid_split is not None:
+            # Support multiple validation splits
+            valid_splits = config.dataset.valid_split.split(",")
```
cbalioglu: Same feedback as in ASR.

recipes/wav2vec2/ssl/recipe.py
```diff
@@ -0,0 +1,199 @@
...
+        return self._criterion(batch, metric_bag)
+
+    @property
+    @override
+    def model(self) -> RecipeModel:
+        return self._criterion.model
+
+
+@final
+class Wav2Vec2SslEvalUnit(EvalUnit[SequenceBatch]):
```
cbalioglu, line 157: Simiar to ASR. Why don't we just use the eval unit under eval/recipe.py?

src/fairseq2/models/wav2vec2/asr/interop.py
```diff
@@ -18,11 +18,21 @@
...
     state_dict: dict[str, object], config: Wav2Vec2AsrConfig
 ) -> dict[str, object]:
+
     try:
         state_dict = cast(dict[str, object], state_dict["model"])
     except KeyError:
         pass
 
+    # TODO: cirquit - remove after this was merged into v0.5.0a2
+    if "module.encoder_frontend.feature_extractor.layers.0.conv.weight" in state_dict:
```
cbalioglu: This is all "fixed". You can remove this if block.

recipes/wav2vec2/asr/data/dataset.py
cirquit: Happy to remove, this is easily to add if need be and I never saw it being used in practice.

recipes/wav2vec2/asr/data/dataset.py
cirquit: Good question, I think this was already part of the original code-base and I copy-pasted it :D Let me investigate this and add a comment to not have random constants in the code.

recipes/wav2vec2/asr/data/dataset.py
cirquit: Isn't your proposed code slightly different?

The original code keeps `max_audio_len>=` samples and truncates them to that size, while your code filters them out resulting in less total samples.

recipes/wav2vec2/asr/data/dataset.py
cirquit: Agreed, was an artifact of the old code and I kept it without thinking too much about it.

recipes/wav2vec2/asr/data/dataset.py
cirquit: 👍 to not exposing this at all.

recipes/wav2vec2/asr/data/dataset.py
cirquit: I specifically left this there to showcase that this is where the packing functionality should be added first, but this just looks a bit unfinished. Will remove.

recipes/wav2vec2/asr/data/preprocessing.py
cirquit: Ah, true, can be removed!

recipes/wav2vec2/asr/config.py
cirquit: It kind of made more sense for me to think of this being the default recipe configuration because we have the yaml configs which overwrite it, coupled with the fact that we have many other different configuration types (models, datasets) so giving this its unique name helped me grok the structure of fairseq2's recipes. Just a personal opinion, happy to align with the idiomatic naming.

recipes/wav2vec2/ssl/data/dataset.py
cirquit: Happy to do that!

recipes/wav2vec2/asr/recipe.py
cirquit: I deliberately added a bit of code duplication for users to not be tied to the validation not needing to be identical with the evaluation (as a showcase to make something of their own).

If we re-use this implementation it creates a hard dependency of the training recipe to the eval recipe, which can make thinking about it a bit harder (before they only shared the dataset implementation). Basically a trade-off between code duplication and a slight UX improvement.

recipes/wav2vec2/asr/data/dataset.py
cirquit: I wasn't able to find the source of this so I removed it.

recipes/wav2vec2/asr/data/dataset.py
cirquit: The call to `Path(header)` actually does not raise any errors in any case, so this could be simplified a lot.

recipes/wav2vec2/asr/data/preprocessing.py
cirquit: Actually, I found a few more functions that were unused, I removed all of them. I had the idea that this could be something like a playground to create your own recipe but this just adds unnecessary noise.

recipes/wav2vec2/asr/data/preprocessing.py
cirquit: Renamed to `add_audio_file_loading`.

recipes/wav2vec2/asr/eval/recipe.py
cirquit: Lets keep it simple for this recipe!

recipes/wav2vec2/asr/config.py
cirquit: Renamed to config.py

recipes/wav2vec2/asr/recipe.py
cirquit: I used the implementation in the training recipe and referenced it in the eval recipe, this made a bit more sense to me from an abstraction point of view (import path goes training recipe -> eval recipe and not the other way around).

recipes/wav2vec2/asr/data/dataset.py
cirquit: I also just noticed that this is a bug. We only cropped audio metadata before, not the actual audio which let `>=max_audio_len` samples through to the length-based batching. We actually need to filter by min and by max audio length for this to function without additional cropping (exactly what you've proposed).

recipes/wav2vec2/ssl/data/dataset.py
cirquit: Not needed, removed!

cirquit:
I just noticed that we might need to move the dataset assets into the recipe definition itself, as it otherwise pollutes the the asset namespace but can't resolve the datasets if they are not registered during runtime (which only happens while running the wav2vec2 recipes). Unsure how this would interface with fairseq2-ext though where we populate the actual asset paths @cbalioglu wdyt?

PR #1352, opened 2025-10-07
Revise model sharding

cbalioglu:
Although this PR appears large, most of the changes are related to documentation rather than code. The PR includes the following updates:

- **New API Documentation:** Added documentation for ``fairseq2.device`` and ``fairseq2.models.hub``.
- **Documentation Fixes and Improvements:** Enhanced the documentation for ``fairseq2.gangs``, ``fairseq2.model_checkpoint``, and ``fairseq2.models.transformers.ffn``.
- **New Helper Function:** Introduced a new, single-line ``get_current_device()`` helper function in ``fairseq2.device``.
- **Context Manager Support for ``Gangs``:** ``Gangs`` can now be used as a context manager, along with a new ``maybe_get_current_gangs()`` helper function. This feature was introduced for two reasons:
    * To better align gangs with the device APIs of PyTorch and make it easier to access the "ambient" gangs in procedural code.
    * To avoid changing the signature of model factory functions (i.e. maintain backwards compatibility) that require access to gangs.
- **Tensor-Parallelism Support:** All subclasses of ``FeedForwardNetwork`` and ``StandardMultiheadAttention`` now accept an optional ``Gangs`` parameter. When provided, these modules apply tensor-parallelism to their parameters during initialization.
- **Factory Function Updates:** Updated the factory functions for Llama and Qwen. They now access the current gangs via ``maybe_get_current_gangs()`` and shard the constructed model on the fly during initialization.
- **``StandardModelFamily`` Updates:** Updated ``StandardModelFamily`` to call model factories within a ``with gangs`` context. The use of ``shard_spec`` and ``sharder`` parameters is now marked as deprecated, as the responsibility for sharding a model has shifted to the model factories.
- **Deprecation Notice:** Marked the entire ``fairseq2.sharder`` module as deprecated.
- **Behavior Change in Recipes:** Introduced a minor (but backwards-compatible) behavior change in recipes: if a model does not support a certain parallelism strategy, an error is no longer raised. This is because it is often difficult to determine how a model uses a parallelism strategy (e.g. due to many different variants of MoE) in a generic way. Since a recipe can still run even if the model omits the use of some gangs, it is now the user's responsibility to ensure that the targeted model can effectively use the requested parallelism modes.
- **StandardMultiheadAttention:** Marked `q_proj`, `k_proj`, `v_proj`, and `out_proj` parameters as deprecated. They were originally meant for handling parallelism in a centralized location with the goal of keeping modules themselves agnostic to parallelism modes, but the truth is that most parallelism techniques require intimate knowledge of module internals, so we changed our approach to handling parallelism directly within modules (e.g. `StandardMultiheadAttention` can shard itself for tensor-parallelism when a `Gangs` instance is provided)

## Testing
Tested both LLaMa and Qwen family models in old and new sharding implementations and verified exact bitwise parity.

## Follow-Up Tasks Before v0.6
- Update the recently merged LLaMA 4 model factory to use the new sharding approach
- Write a migration post that describes how to move from `ModelSharder` to factory-based sharding
- Update the deprecation notices in this PR with the URI of the migration post
- Update CHANGELOG


src/fairseq2/models/qwen/factory.py
```diff
@@ -81,10 +87,17 @@ def init_embed(embed: StandardEmbedding) -> None:
...
             _init_truncated_normal(embed.weight, bias=None, std=std)
 
-        return StandardEmbedding(
+        embed = StandardEmbedding(
             config.vocab_size, config.model_dim, init_fn=init_embed
         )
 
+        gangs = self._gangs
+
+        if gangs is not None and gangs.tp.size > 1:
```
cirquit, line 96: I like the change of putting the TP logic into the factory, it seems a home here. 

Nit: The general look of implementing TP in a model looks hard to follow me though, because we're not delegating the gangs argument in the respective functions but using it as a private member. I roughly get the picture of "sometimes we use gangs for TP, its up to you to figure out how to do it correctly yourself, just think", but instead I would have liked to get the vibe of "if you see gangs, we need to do TP, have you handled it correctly for this particular layer if you have the argument available?"

To be fully clear, I would have expected something like this:

```python
def create_decoder(self, gangs) -> TransformerLMDecoder:
                         ^^^^^
(...)
    for idx in range(config.num_layers):
        layer = self.create_decoder_layer(idx, pos_encoder, gangs=gangs)
                                                            ^^^^^^^^^^^
```

Its trading verbosity and code duplication for slightly a more traceable TP implementation. Just some food for thought, I don't have a strong opinion, just a slight preference here.

src/fairseq2/models/qwen/factory.py
```diff
@@ -232,10 +247,17 @@ def init_projection(proj: Linear) -> None:
...
 
-        return Linear(
+        final_proj = Linear(
             config.model_dim, config.vocab_size, bias=False, init_fn=init_projection
         )
 
+        gangs = self._gangs
+
+        if gangs is not None and gangs.tp.size > 1:
+            return ColumnShardedLinear.from_linear(final_proj, gangs.tp)
```
cirquit, line 257: Quick sanity check as I saw the gangs.tp, how does this interplay with pipeline parallelism and the fact that we moved the responsibility of a factory governing TP implementation?

My current understanding is that we define the rules on how to shard a model in the TP dimension in the factory, but define the precise sharding strategy in `shard_dims`. DP is done on a trainer level, similar to FSDP. Is the factory ever going to need knowing about the PP dimension in any way? (I don't think so, but would be happy to be corrected)

src/fairseq2/models/transformer/ffn.py
```diff
@@ -153,6 +188,17 @@ def __init__(
             dtype=dtype,
         )
 
+        self.inner_proj: Projection
+
+        if gangs is None or gangs.tp.size == 1:
+            self.inner_proj = inner_proj
+        else:
+            self.inner_proj = ColumnShardedLinear.from_linear(
+                inner_proj, gangs.tp, gather_output=False
```
cirquit, line 197: The `gather_output=False` looks a bit hard-coded to me. I thought this was defined in the shard_dims, but when I checked I noticed that only define the row/col dimension there (right?)

This was part of the shard_spec with `region_boundary`, how are we handling this now?

cirquit: Or are you thinking about this implementation in a way where the factory is not simply the "recipe" on how to create the model, but also takes over some of the parallelization logic? The followup from that would be that this should be handled on a model-configuration level, similar to how the embeddings are col/row-sharded optionally with the wav2vec2 models.

I'm ok with that, it just makes reasoning about where to parallelization strategy holistically a bit hard:
- We define (FS)DP in trainer
- We define **how to run generic TP** in factory with configuration options in the model config
- However, we also define `shard_dims` to actually specify when to use TP, but the specific way of doing TP must be done in the model_config which parametrizes the factory.

Am I misunderstanding something?

The previous shard_spec had this coupled together, and this was very easy to reason about.

src/fairseq2/models/family.py
```diff
@@ -527,14 +532,13 @@ def _do_create_model(
                 )
 
             device = META_DEVICE
-        elif gangs.root.size != gangs.dp.size:
```
cirquit, line 530: This is a funny thing to debug, I hope nobody was depending on this.

src/fairseq2/models/hub.py
```diff
@@ -130,6 +356,32 @@ def load_model(
...
+        config: ModelConfigT,
+        *,
+        device: Device | None = None,
+        dtype: DataType | None = None,
+        mmap: bool = False,
+        restrict: bool | None = None,
+        progress: bool = True,
+    ) -> ModelT: ...
+
     def load_custom_model(
```
cirquit, line 385: qq - does this support loading models checkpoints that we created during training, e.g. for a quick manual evaluation run with EvalRecipe without consolidating and creating a card for it?

If yes, this would be the better example to showcase/or at least mention, this is a common friction point I heard from MMS.

src/fairseq2/gang.py
```diff
@@ -690,15 +693,63 @@ def __post_init__(self) -> None:
...
+
+    Within fairseq2, this function is used by model factories to retrieve the
+    current gangs and shard the constructed models accordingly. The current gangs
+    are set internally by fairseq2 before calling the factories.
+
+    Note that the return value of this function is thread specific. Individual
+    threads may have their own set of current gangs.
+    """
+    if _thread_local.current_gangs:
+        return cast(Gangs, _thread_local.current_gangs[-1])
```
cirquit, line 743: Cool trick! Any specific reason we might expect to have nested context gang calls for this to be index -1 instead of 0?

cirquit (commented):
Neat PR, just have a few general architecture and things I maybe did not understand fully yet.

src/fairseq2/models/qwen/factory.py
cbalioglu: Well, doing that would be nice, but it would also break BC if you don't do specify your parameter as `gangs: Gangs | None = None`. Having said that, the same can be said for the `config` variable as well. The model factory code is more tuned to researchers than engineers and for them brevity is always preferred. I am 100% sure that they would complain about having to pass that "annoying" gang parameter all over their code, why do they need that useless thing.

In fact the "advanced" version of the model factory also passes a dependency container, but we deliberately kept this basic version so that things stay simple enough for researchers doing just some architectural research.

In summary, nothing wrong with your suggestion, and I ideally agree that passing in fact both config and gangs would be more "pedantic" and right, but this is one of those places where we interface with researchers.

src/fairseq2/models/qwen/factory.py
cbalioglu: I think the factory needs to be aware of PP as well. It is very hard for fairseq2 to infer at what boundaries the model should be split. So something like the following might be used:

```
def create_model(self) -> Model:
   if gangs.pp.rank == 0:
      return self.create_frontend_and_first_stack_split()

  if gangs.pp.rank == gangs.pp.size - 1:
     return self.create_output_proj_and_last_stack_split()

  return self.create_intermediate_stack_split()
```

We can certainly expose a helper function though that would work for majority of transformer-based model architectures that the model author can call within their "create_model" method.

```
def create_model(self) -> Model:
   model = self._do_create_model()
   
   # Assumes a vanilla decoder stack.
   return split_for_pp(model, gangs.pp)
```

This is still something that is handled within the model factory though.

src/fairseq2/models/transformer/ffn.py
cbalioglu: Well `gather_output=False` is an "optimization" that FFN and MHA do. It is not hard-coded, but is a very obvious trick to perform to avoid an extra scatter/gather between input and output projections.

> However, we also define shard_dims to actually specify when to use TP, but the specific way of doing TP must be done in the model_config which parametrizes the factory.

We actually don't specify `shard_dims`. `shard_dims` is auto-generated by `FFN` and `MHA` (and other potential shard-aware modules) and used solely for checkpoint loading. Model authors never define or have to interact with `shard_dims`. It is purely inferred from the model produced by the model factory.

src/fairseq2/models/hub.py
cbalioglu: For model checkpoints, the ideal way to load them is with `register_checkpoint_models(<path>)` and then do a `load_model("last_checkpoint")`, but yes, this function does not require any consolidation. It uses the same model checkpoint loader as `load_model()`. It is exactly meant to be used for quickly loading a model bypassing creating an asset definition.

src/fairseq2/gang.py
cbalioglu: The `__enter__` and `__exit__` methods of `Gangs` use `current_gangs` as a stack and push/pop themselves to it. So the "most current" gangs are always the last one in the stack.

src/fairseq2/models/qwen/factory.py
cirquit: Makes sense, this is useful context!

src/fairseq2/models/qwen/factory.py
cirquit: This makes sense, so kind of a step-by-step process - first create the model, then split for pp. I guess that in practice one would need to specialize the model configuration definition depending on its size, the worldsize + hardware specs to create the optimal split points for PP. But this looks to be entirely possible with our current model configuration setup where we can create whichever parameters we need to make this useful in practice. Thanks for the explanation!

src/fairseq2/gang.py
cirquit: Ah, makes sense!

src/fairseq2/models/transformer/ffn.py
cirquit: Argh, I had the wrong mental model for this even after reading the previous PR 🤦 ! Does this mean that if we want to create a region boundary for TP we must now do this in factory as well?

This looks to be harder to trace than the previous implementation which made this super obvious by exposing the regex + shard dim + region_boundary line after line per module.

I'm wondering whether there is a way to at least retroactively visualize the sharding strategy to help users verify that it works like they wanted to if its not easily visible from factory definition.

src/fairseq2/models/hub.py
cirquit: Oh wow, I was not aware of the "register_checkpoint_models" functionality, this is super cool. We need to add  documentation for this use-case!

src/fairseq2/models/transformer/ffn.py
cbalioglu: You don't have to strictly do it in the factory, since such region boundaries are implemented within reusable modules like `FFN`, or `MHA`. As a model author as long as you pass a `gangs` to those modules, they correctly handle boundary conditions in their initialization logic. In fact I would say that this approach is less error prone to shard spec because as model author you don't really care about details like region boundaries, gathers/scatters. You reuse modules that have already builtin support for parallelism strategies and just imply whether you want them to be parallelized.

I think the easier way to verify if your model is constructed correctly under a certain parallelism strategy is to print it (as we do at the beginning of the training). It clearly gives you which of your modules are sharded and what sharding settings they use.

cirquit: Ah I forgot about the initial visualization, I think I'm a bit log-fatigued when running fs2 recipes haha. Yea, this should be the easiest way to see and debug this!

src/fairseq2/models/qwen/factory.py
cbalioglu: You are welcome!

src/fairseq2/models/hub.py
cbalioglu: I plan to write much more detailed documentation on assets. There are so many features that many people are not aware of. That is one of my goals for next week (along with model authoring)

PR #1369, opened 2025-10-10
Fold expert dim-0 in MoE weights for FSDPv2 compatibility

MartinGleize:

We fold the expert dim and the following dimension into one much larger dim. This dim is unfolded at forward time.

Why: FSDP v2 shards parameters in their dimension 0. In the current MoE weights used in Llama 4, dim 0 is the number of experts, which is typically quite small, compared to how high DP size can scale.

**Test plan:**

Verified parity for inference and checked compatibility with the SFT recipe.

No

**Check list:**

### DIFF
--- a/src/fairseq2/models/transformer/experts.py
+++ b/src/fairseq2/models/transformer/experts.py
@@ -87,7 +87,7 @@ def create_expert_glu_layers(
         module.activation = activation
 
 
-def forward_glu_bmm_with_folded_experts(
+def _forward_glu_bmm_with_folded_experts(
     x: Tensor,
     gate_proj: Tensor,
     inner_proj: Tensor,
@@ -203,7 +203,7 @@ class GroupedExpertNetwork(ExpertNetwork):
         :param x: Input tensor of shape (num_local_experts, tokens_per_expert, dim).
         :returns: Output tensor of shape (num_local_experts, tokens_per_expert, dim).
         """
-        return forward_glu_bmm_with_folded_experts(
+        return _forward_glu_bmm_with_folded_experts(
             x,
             self.gate_proj,
             self.inner_proj,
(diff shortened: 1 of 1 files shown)


src/fairseq2/models/transformer/experts.py
```diff
@@ -100,6 +87,33 @@ def create_expert_glu_layers(
         module.activation = activation
 
 
+def forward_glu_bmm_with_folded_experts(
```
cbalioglu: Nit: My suggestion would be to make this a "private" function until we finalize all the LLaMA MoE work. Otherwise, we will have to maintain it for BC if it becomes part of a rolling release.

src/fairseq2/models/transformer/experts.py
```diff
@@ -60,11 +48,12 @@ def create_expert_glu_layers(
     :param dtype:
         The data type of the layers.
     """
+    # The expert dimension is folded with the first of model/inner dim
+    # to optimize for FSDPv2 (which shards params on dim 0).
```
cbalioglu, line 52: Nit: Not just FSDP2, but FSDP as well.

MartinGleize: I think I misunderstood then. I thought FSDP v1 was flattening params in a 1D tensor and so this concern about the first dim didn't exist.

cbalioglu: True if you use “vanilla” FSDP, but if you use “use_orig_parameters” argument (which we do like most other modern tooling) then FSDP maintains the original shape of the parameters similar to FSDP2

cbalioglu: Having a second thought you might be right though since this is about FSDP’s internal machinery which always keeps params in 1D. Nevermind looks good in all cases :)

PR #1373, opened 2025-10-14
Introduce register_recipe_assets function

cbalioglu:
This PR introduces a new `register_recipe_assets` helper function. The docstr of the function below can be checked out to learn more about its purpose.

The function is successfully tested along with some of the internal speech SSL recipes. Later PRs will migrate the existing fairseq2 recipes to start using it that it will be covered by the relevant integration tests.


zyaoj (approved):
LGTM!

PR #1374, opened 2025-10-15
Fix fp16 scaling when used with TP

cbalioglu:
This PR fixes a bug in `StandardFloat16LossScaler` that fails to sync grad scales correctly when used with tensor parallelism. Since nowadays fp16 is rarely used, we haven't catched this before.


cirquit (approved):
LGTM!

PR #1377, opened 2025-10-16
Make RecipeModel callable

cbalioglu:
This PR makes `RecipeModel `callable, resulting in a more natural API for users. In recipes instead of:

```python
def process_batch(...):
   loss = self.model.module(input)
```

one can now use:

```python
def process_batch(...):
  loss = self.model(input)
```


cbalioglu:
Yep, I agree. As we discussed offline, I will have a draft PR today to get your feedback.

PR #1379, opened 2025-10-16
Handle OS and gang errors as operational in recipes

cbalioglu:
Make life easier for recipe authors and handle `OSError` and `GangError` gracefully by converting them to `OperationalError` so they are reported in a nicely formatted way for users.

### DIFF
--- a/src/fairseq2/recipe/internal/task.py
+++ b/src/fairseq2/recipe/internal/task.py
@@ -10,7 +10,8 @@ from signal import SIGUSR1, signal
 from types import FrameType
 from typing import final
 
-from fairseq2.gang import Gangs
+from fairseq2.error import raise_operational_system_error
+from fairseq2.gang import GangError, Gangs, raise_operational_gang_error
 from fairseq2.logging import log
 from fairseq2.recipe.task import Task, TaskStopException
 from fairseq2.utils.stopwatch import Stopwatch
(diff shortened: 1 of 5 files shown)


src/fairseq2/recipe/internal/task.py
```diff
@@ -10,22 +10,20 @@
...
-    def __init__(self, task: Task, gangs: Gangs, wall_watch: Stopwatch) -> None:
-        self._task = task
+    def __init__(self, gangs: Gangs, wall_watch: Stopwatch) -> None:
         self._gangs = gangs
         self._wall_watch = wall_watch
 
-    def run(self) -> None:
-        task = self._task
-
+    def run(self, task: Task) -> None:
```
zyaoj, line 26: This change makes sense! Then at the same time, I have the impression that we need to better document our concept of `Task` along with our `Trainer(Task)` -- this is somewhat not explicit in our documentation yet. Something I'll also poke around and see what would be the best way forward.

zyaoj (approved):
LGTM!

src/fairseq2/recipe/internal/task.py
cbalioglu: yeah, I agree (again). Overall the whole recipe architecture and how things are composed together has literally zero documentation right now. We should heavily invest into documenting it.

PR #1383, opened 2025-10-17
Richard/flat ref api

rsyue:

- Renamed ‘index.rst’ file within dirs to corresponding name
- flattened the models, data, and datasets directories
- Renamed and capitalized headings
- Replaced :toctree: with * :doc: path/file.rst references
- Added descriptions where applicable
- Some static links were fixed and underscore notation was modified to reflect the module name (e.g. fairseq2.nn.normalization for the LayerNorm class)

Fixes #{170}

Nothing to report

**Check list:**


doc/source/reference/nn/batch_layout.rst
```diff
@@ -1,7 +1,7 @@
 .. _batch_layout:
 
-Batch Layout
```
zyaoj, line 3: Let's keep the format consistent and make the title the same as the module name, in this case: `fairseq2.nn.batch_layout`. Same applies for embedding, incremental state, position encoder,  projection.

doc/source/reference/models/hub.rst
```diff
@@ -241,6 +241,6 @@ See Also
 
 * :doc:`/guides/add_model` - Tutorial on adding new models
 * :doc:`/basics/assets` - Understanding the asset system
-* :doc:`/reference/models/index` - Models API overview
```
zyaoj, line 244: A comment for the future us -- I think lots of the content here should be actually moved to basics/guides. The reference part should be kept as API references with minimal examples (preferably as inline docstring as well).

For this model hub api specifically, the API is actually already covered in `referece/fairseq2.models.rst`.

doc/source/reference/nn/batch_layout.rst
zyaoj: Just a follow up to this discussion, Richard made a valid point about keeping the title the same as the class name, and especially for `BatchLayout` which is probably an important concept that we want to cover independently.

As we are moving to a "flat" structure for all our `.rst` files, we might also want to discuss how to better organize the pages so that we keep the indexing well maintained and easy to access. One example is to keep `fairseq2.nn.batch_layout.rst` with its own doc file but insert it as a subpage of `fairseq2.nn` with:

```sphinx
.. toctree::
   :maxdepth: 1
   
   fairseq2.nn.batch_layout
```

so that we can get the following structure -- `BatchLayout` under `fairseq2.nn`:

Screenshot 2025-10-17 at 19 38 24

Open to discussion about UX!

doc/source/reference/fairseq2.nn.Embedding.rst
```diff
@@ -1,7 +1,7 @@
 .. _embedding:
 
-Embeddings
-==========
+fairseq2.nn.Embedding
```
cbalioglu: I think this should be in lower-case, since this is the name of the module. `Embedding` (the interface) is on line 8 in this file.

doc/source/reference/fairseq2.nn.IncrementalState.rst
```diff
@@ -1,7 +1,7 @@
 .. _incremental_state:
 
-Incremental State
------------------
+fairseq2.nn.IncrementalState
```
cbalioglu: Same as above. If we name this `IncrementalState` instead of module's name, the user would have no visual clue that this page also includes the description of `IncrementalStateBag` (which is actually the "more important" class in this module).

doc/source/reference/nn/batch_layout.rst
cbalioglu: I think `BatchLayout` is an exception since that particular module contains only the `BatchLayout` class, but we have many Python files where we have multiple classes and interfaces. For instance, in `fairseq2.nn.position_encoder` we not only have `PositionEncoder` interface and all its implementation, but also `InterpolatedPositionEncoder` interface which is not a derivative of `PositionEncoder`.

I think it is mentally easier to have a 1-to-1 mapping between rst and Python files, instead of requiring users to guess which one is the "right" one.

doc/source/reference/models/hub.rst
cbalioglu: yeah, we should make a clear cut between what goes into a concept document vs. API. I agree that the API documentation should be mostly about how to use the API and maybe with some brief mention of the relevant concepts. The core learning material should be in concepts and/or how-to guides.

doc/source/reference/nn/batch_layout.rst
rsyue: Noted on my end. I will look at the module file name when titling.

doc/source/reference/models/hub.rst
rsyue: Good point. These were artifacts from the nested portion where someone could click on the root of fairseq.nn > and see its children. They can be moved to basics/guides.

N.B. If the number of classes within modules grows by a lot, we can reintegrate them as roots in the tree if we go back to that style.

zyaoj (approved):
LGTM!

PR #1386, opened 2025-10-20
Introduce FAIRSEQ2_NO_PROGRESS env variable

cbalioglu:
This PR introduces a new `FAIRSEQ2_NO_PROGRESS` environment variable and a new `no_progress` parameter in `init_fairseq2` function. As their names suggest, they are meant to switch off all progress bars used for various purposes such as asset downloading and model checkpoint loading. With that, the current `progress` parameters spread out through the code base are all deprecated and will be removed in v0.13. Note though due to the limitations of huggingface_hub, we cannot intervene its progress bars and users should explicitly disable it via `HF_HUB_DISABLE_PROGRESS_BARS` environment variable as described [here](https://huggingface.co/docs/huggingface_hub/v1.0.0.rc6/en/package_reference/file_download#huggingface_hub.snapshot_download).

### DIFF
--- a/src/fairseq2/assets/download_manager.py
+++ b/src/fairseq2/assets/download_manager.py
@@ -24,12 +24,12 @@ from zipfile import BadZipFile, ZipFile
 
 from huggingface_hub import snapshot_download
 from huggingface_hub.errors import HfHubHTTPError
-from tqdm import tqdm  # type: ignore[import]
 from typing_extensions import override
 
 from fairseq2.error import NotSupportedError
 from fairseq2.logging import log
 from fairseq2.runtime.dependency import get_dependency_resolver
+from fairseq2.utils.progress import ProgressReporter
 from fairseq2.utils.uri import Uri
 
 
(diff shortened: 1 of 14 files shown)


src/fairseq2/assets/download_manager.py
```diff
@@ -72,13 +73,15 @@ def download_tokenizer(
...
-        :param tokenizer_name: The name of the tokenizer.
-        :param force: If ``True``, downloads the tokenizer even if it is already
-            in cache.
-        :param progress: If ``True``, displays a progress bar to stderr.
+        Returns the path to the downloaded tokenizer.
+
+        If ``force`` is ``True``, the tokenizer will be downloaded even if it is
+        already in cache.
+
+        ``progress`` is deprecated and will be removed in v0.13.
```
cirquit: This looks as if we're deprecating and not providing an alternative. Can we add a reference to `FAIRSEQ2_NO_PROGRESS`? Not sure what the most top-level reference in the documentation is that we should link to though.

src/fairseq2/utils/warn.py
```diff
@@ -23,3 +23,11 @@ def enable_deprecation_warnings() -> None:
 
 def _warn_deprecated(msg: str) -> None:
     warnings.warn(msg, DeprecationWarning, stacklevel=2)
+
+
+# TODO: Remove in v0.13
```
cirquit, line 28: Can we create a test or something in the CI to track what needs to be removed at a particular dev version before we release (not part of this PR ofc)? I imagine this will become an issue very soon to keep track of all of the deprecations.

cbalioglu: How about extending `_warn_deprecated` with a mandatory `version` parameter? In `_warn_deprecated` we can then throw an `InternalError` if the current version of fairseq2 matches the specified `version`. Something like:

```python
from packaging.Version

def _warn_deprecated(msg: str, eol_version: Version) -> None:
  if fairseq2.__version__ >= eol_version:
    raise InternalError("Deprecated feature not removed at or after end-of-life version")
```

src/fairseq2/assets/download_manager.py
cbalioglu: Makes sense. Let me add it.

src/fairseq2/utils/warn.py
cirquit: This is a neat solution, the CI will catch that while building for the target version at worst!

cbalioglu: https://github.com/orgs/fairinternal/projects/173/views/1?pane=issue&itemId=134623353&issue=facebookresearch%7Cfairseq2%7C1387

PR #1395, opened 2025-10-23
Register no-op as the CPU stat tracker to avoid recipes crashing on CPU

MartinGleize:
Register the no-op tracker as the recipe default CPU stat tracker to avoid recipes crashing on CPU.

A subsequent PR might add an actual CPU stat tracker, but for now since CPU-only is not a major use-case, no-op is good enough, while allowing users to debug and unit test their recipes on CPU.

No

**Check list:**

### DIFF
--- a/src/fairseq2/recipe/composition/device_stat.py
+++ b/src/fairseq2/recipe/composition/device_stat.py
@@ -9,9 +9,9 @@ from __future__ import annotations
 from fairseq2.recipe.internal.device_stat import _RecipeDeviceStatTrackerProvider
 from fairseq2.runtime.dependency import DependencyContainer, DependencyResolver
 from fairseq2.utils.device_stat import (
+    NOOP_DEVICE_STAT_TRACKER,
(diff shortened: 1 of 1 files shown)


src/fairseq2/recipe/composition/device_stat.py
```diff
@@ -24,3 +28,7 @@ def get_device_stat_tracker(resolver: DependencyResolver) -> DeviceStatTracker:
 
     # CUDA
     container.register_type(DeviceStatTracker, CudaDeviceStatTracker, key="cuda")
+
+    # CPU
+    # No-op implementation is enough for now, since CPU-only is not a primary use case
+    container.register_type(DeviceStatTracker, _NoopDeviceStatTracker, key="cpu")
```
cbalioglu: A slightly cleaner way would be:

```python
container.register_instance(DeviceStatTracker, NOOP_DEVICE_STAT_TRACKER, key="cpu")
```

PR #1396, opened 2025-10-28
set VLLM_ALLOW_INSECURE_SERIALIZATION=1 for newer vllm versions and add a comment for calling newer vllm generate function

lydiadli:

- I updated the code so that they are compatible to newer vllm versions
1. after updating to newer vllm version, we observed model sync issue with error """TypeError: Object of type <class 'torch.dtype'> is  not serializable. Set VLLM_ALLOW_INSECURE_SERIALIZATION=1 to allow fallback to pickle-based serialization. """
Adding this flag would relax it and enable jobs to run. 
2.  generate function's signature has been updated in newer version. and I updated the prompt inputs to cater to the change.

Fixes #{issue number}

**Check list:**
- [ ] Did you write any **new necessary tests**?  No
- [X] Did you verify new and **existing tests pass** locally with your changes? Yes
    1. i applied a unit test by running vllm generation function itself with both text and token inputs and validated their equality using llama 1b model
    2. further i run the training pipeline with qwen model to test the token input and validated that the training steps are progressing with losses logged.
- [ ] Did you **update the [CHANGELOG](https://github.com/facebookresearch/fairseq2/blob/main/CHANGELOG.md)**? (no need for typos, documentation, or minor internal changes)? not applicable

### DIFF
--- a/src/fairseq2/recipes/lm/_online_finetune/_remote_model.py
+++ b/src/fairseq2/recipes/lm/_online_finetune/_remote_model.py
@@ -12,7 +12,6 @@ from abc import ABC, abstractmethod
 from collections import Counter
 from dataclasses import dataclass, field
 from typing import Any, Dict, Union
-
 import ray
 import torch
 from ray.util.placement_group import placement_group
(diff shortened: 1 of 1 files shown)


uralik (approved):
thanks!!

PR #1399, opened 2025-10-30
Introduce DataParallelFacade

cbalioglu:
This PR is the first in a series of PRs that are going to deprecate `RecipeModel` and remove the complexities it brings to the recipe APIs. This first PR introduces a new `DataParallelFacade` interface along with accompanying `set_data_parallel_facade()` and `get_data_parallel_facade()` functions to offer an API-agnostic way to access data parallelism implementations.

```python
model = FSDP(...)

# state_dict is access the same way regardless if the data parallel module is of type FSDP or DDP.
dp_facade = get_data_parallel_facade(model)
dp_facade.state_dict()

model = DDP(...)

dp_facade = get_data_parallel_facade(model)
dp_facade.state_dict()
```

### DIFF
--- a/src/fairseq2/nn/data_parallel.py
+++ b/src/fairseq2/nn/data_parallel.py
@@ -192,15 +192,16 @@ def get_data_parallel_facade(module: Module) -> DataParallelFacade:
     """
     facade = getattr(module, _FACADE_KEY, None)
     if facade is None:
-        if isinstance(module, DDPModule):
-            return _DDPFacade(module)
+        match module:
+            case DDPModule():
+                facade = _DDPFacade(module)
+            case FSDP1Module():
+                facade = _FSDP1Facade(module)
+            case FSDP2Module():
+                facade = _FSDP2Facade(module)
+            case _:
+                facade = _NoopDataParallelFacade(module)
 
-        if isinstance(module, FSDP1Module):
-            return _FSDP1Facade(module)
-
-        if isinstance(module, FSDP2Module):
-            return _FSDP2Facade(module)
-
-        facade = _NoopDataParallelFacade(module)
+        setattr(module, _FACADE_KEY, facade)
 
     return facade


src/fairseq2/nn/data_parallel.py
```diff
@@ -0,0 +1,206 @@
...
+        if isinstance(module, DDPModule):
+            return _DDPFacade(module)
+
+        if isinstance(module, FSDP1Module):
+            return _FSDP1Facade(module)
+
+        if isinstance(module, FSDP2Module):
+            return _FSDP2Facade(module)
+
+        facade = _NoopDataParallelFacade(module)
```
zyaoj: [really nit] do we want to cache this instance (or mapping like using a global dict (or `weakref.WeakKeyDictionary[Module, DataParallelFacade]`)?

cbalioglu: Thanks for the tip! I think it makes sense. I just improved the code a bit to cache the returned facade instances.

cbalioglu:
> Is this to overwrite parallelism strategies' configurations inside the recipe instead of doing it as part of the yaml configs (and do this abstracted over parallelism types)?

This is (so far) not related to recipes at all. It is under `fairseq2.nn` and is just a fancier/cleaner way of accessing common operation on data parallel models. For instance, you can think of `DataParallelFacade.state_dict()` effectively equivalent of:

```python
from torch.nn.parallel import DistributedDataParallel as DDP
from torch.distributed.fsdp import FSDPModule

def state_dict(module: Module) -> dict[str, object]:
  if isinstance(module, DDP):
    return module.state_dict()
    
  if isinstance(module, FSDPModule):
    return fsdp2_local_state_dict(module)
 ```

Since DDP, FSDP, and FSDP2 all have varying APIs for state handling, gradient clipping, parameter summoning, we hide them behind a "facade", so that the rest of the code base does not need to have API specific logic.

PR #1404, opened 2025-11-03
Improve the implementation of Uri

cbalioglu:
This PR improves the implementation of `Uri` type with some new features. This is mainly a preliminary work for some of the updates coming to `AssetDownloadManager`.

### DIFF
--- a/src/fairseq2/utils/uri.py
+++ b/src/fairseq2/utils/uri.py
@@ -134,7 +134,7 @@ class Uri:
 
         s = str(self)
 
-        return Path(s[5:])
+        return Path(s[7:])
 
     def __str__(self) -> str:
         return urlunparse(self._result)


src/fairseq2/utils/uri.py
```diff
@@ -37,8 +88,9 @@ def from_path(path: Path) -> Uri:
 
         return uri
 
-    def __init__(self, _result: ParseResult) -> None:
```
zyaoj, line 40: This seems to be a minor breaking change in case people might be using `Uri` independently as a public class. Shall we default `_params` to `{}` or some other values?

src/fairseq2/utils/uri.py
```diff
@@ -64,5 +120,25 @@ def query(self) -> str:
...
+
+    def to_path(self) -> Path:
+        if self.scheme != "file":
+            raise NotSupportedError(
+                "`to_path()` is only supported for URIs with file scheme."
+            )
+
+        s = str(self)
+
+        return Path(s[5:])
```
zyaoj: This seems a bit fragile at the first glance. The least we can do is to add a comment if we don't want to harden this with regex and handle other os. This might not be an issue if the URI is instantiated from `parse` but we might not be able to avoid this public class being abused externally.

zyaoj (approved):
LGTM! Just 2 nit comments.

src/fairseq2/utils/uri.py
cbalioglu: This `__init__` method was "marked" as private from the beginning using the underscore parameter convention. We expect people to always use `parse()`, `maybe_parse()`, and `from_path()` methods to construct a `Uri`.

src/fairseq2/utils/uri.py
cbalioglu: I think this is safe since we check `self.schema` and know for sure that the string version of the path will always start with a "file://" prefix.

PR #1415, opened 2025-11-07
ppl/logp reward

lydiadli:
add ppl/logp reward for grpo

Fixes #{issue number}

**Check list:**
-- I did a test run locally before submitting and had a longer run with 1k+ steps training.

### DIFF
--- a/src/fairseq2/recipes/lm/_online_finetune/_common.py
+++ b/src/fairseq2/recipes/lm/_online_finetune/_common.py
@@ -116,7 +116,6 @@ def generate_rollouts(
     dp_gang,
     vllm_model,
     sampling_params=None,
-    string_input=False,
 ):
     prompts_to_generate = [None] * dp_gang.size
     if dp_gang.rank == 0:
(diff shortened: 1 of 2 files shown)


src/fairseq2/recipes/lm/_online_finetune/_common.py
```diff
@@ -129,7 +130,9 @@ def generate_rollouts(
             flat_request_list.extend(rank_prompts)
 
         rollouts = vllm_model.rollout_from_model(
-            flat_request_list, sampling_params=sampling_params
+            prompt_list=flat_request_list,
+            sampling_params=sampling_params,
+            string_input=string_input,
```
jacklanchantin: this will screw up other calls to `generate_rollouts`, no? should we create another function for just this?

lydiadli: this should be backward compatible because the vllm_model.rollout_from_model being called has:
"""def rollout_from_model(self, prompt_list, sampling_params=None, string_input=False)""" the default behavior is having string_input as false. we keep the same default here but only allowing controling this flag :)

jacklanchantin: we're never using `string_input=True`, right? We can probably remove that

PR #1445, opened 2025-11-14
Hg model training with torch DataLoader

annasun28:
* Update fairseq2 HgTokenizer with some methods required from the base transformers Tokenizer. TBD: whether we should actually subclass and override methods so that we automatically fall back to the base transformers Tokenizer methods?
* Enable FSDP for Hg model (matching the way FSDP wrapping is handled in `transformers` / `accelerate`). Note: to override transformer_cls_names_to_wrap, one needs to register a new model family (see the corresponding seamless_next PR for example)
* Fix this issue:
  * fs2 overrides model.state_dict in _FSDP1Model to return fsdp1_local_state_dict (e.g. when checkpointing). If the tensor is sharded unevenly (e.g. if you have a nn.Embedding(2, dim) and sdp.size=4), the key is skipped (https://github.com/facebookresearch/fairseq2/blob/main/src/fairseq2/nn/fsdp/fsdp1.py#L192-L193) which causes a hang when reloading (IIUC due to an allgather in the FSDP pre state dict hook)

Tested with https://github.com/fairinternal/seamless_next/pull/750

N/A

**Check list:**

### DIFF
--- a/tests/unit/models/hg/test_hg_factory.py
+++ b/tests/unit/models/hg/test_hg_factory.py
@@ -524,12 +524,19 @@ class TestImportClassFromTransformers:
         assert "TestClass" in str(exc_info.value)
         assert "not found" in str(exc_info.value)
 
-if torch.cuda.is_available() and torch.cuda.device_count() > 1:
-    from fairseq2.assets import get_asset_store
-    from fairseq2.device import get_default_device
-    from fairseq2.gang import Gang, Gangs, ProcessGroupGang, create_parallel_gangs, maybe_get_current_gangs
-    from fairseq2.models.hg import get_hg_model_hub
-    import torch.distributed as dist
+if __name__ == "__main__":
+    """
+    Hardware dependent test:
+    If more than one GPU is present, model sharding with tp
+    can be tested.
+    """
     
-    mg_factory = TestHgFactory()
-    mg_factory.test_create_model_special_model_with_gangs()
+    if torch.cuda.is_available() and torch.cuda.device_count() > 1:
+        from fairseq2.assets import get_asset_store
+        from fairseq2.device import get_default_device
+        from fairseq2.gang import Gang, Gangs, ProcessGroupGang, create_parallel_gangs, maybe_get_current_gangs
+        from fairseq2.models.hg import get_hg_model_hub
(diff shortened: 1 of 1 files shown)


src/fairseq2/models/hg/factory.py
rsyue, line 1: This looks good. Just wondering if the process granularity is handled by the RecipeModel class during loading. I put barrier() in there because occasionally model info would be on another process and it would not recognize from interleaving.

src/fairseq2/nn/fsdp/fsdp1.py
```diff
@@ -189,10 +189,12 @@ def fsdp1_local_state_dict(module: FSDP1Module) -> dict[str, object]:
         for name, item in module.state_dict().items():
             if isinstance(item, ShardedTensor):
                 local_shards = item.local_shards()
-                if not local_shards:
-                    continue  # means the tensor is sharded unevenly.
+                if local_shards:
+                    item = item.local_tensor().detach()
+                else:  # means the tensor is sharded unevenly.
+                    # still keep the empty tensor in the state_dict
+                    item = Tensor(item.local_shards())
```
zyaoj: Thanks for this fix! Looking forward to an upstreamed PR to main for this.

src/fairseq2/models/hg/factory.py
annasun28: Hm, in the seamless_next recipe, I called `model = model_hub.load_model(card, config=hg_model_config)` before setting up gangs, which I guess is why this was causing an error. Let me try initializing the gangs first and see. But just curious, why are barrier() needed for this HF model, but not needed for others?

rsyue: From what I understand, the _load_special_model in factory sets a specific config as opposed to using AutoConfig. Qwen2.5-Omni does not accept AutoConfig because it is sort of special purpose. I noticed with torchrun that sometimes the main process would correctly set the config and others would not, leading to an error: "Invalid model config. Select from BertConfig, LlamaConfig, ..." on the other handles. I assumed this was because they were not properly synced, so thought that letting them all reach that point would resolve it. But it seems like you grab the config directly first in your recipe. I loaded from the card and let it find the config itself.

PR #1468, opened 2025-11-28
Change log.warn to log.warning

LittlePea13:
Quite self explanatory, there's a typo on log.warn instead of log.warning


cbalioglu (approved):
Thanks for the fix! LGTM!

PR #1471, opened 2025-12-01
Fix sampling error with many pipelines when allow_repeats=False

caciolai:

Fixes https://github.com/facebookresearch/fairseq2/issues/1472

**Check list:**

### DIFF
--- a/doc/source/reference/fairseq2.data_type.rst
+++ b/doc/source/reference/fairseq2.data_type.rst
@@ -13,6 +13,5 @@ ABCs
 Functions
 =========
 
-.. autofunction:: current_dtype
 .. autofunction:: get_current_dtype
 .. autofunction:: set_dtype


native/src/fairseq2n/data/sample_data_source.cc
```diff
@@ -212,9 +237,10 @@ sample_data_source::block(std::size_t idx)
         weight_cumsums_[i] -= weight;
     }
 
-    float32 sum = weight_cumsums_.back();
+    // Find the actual maximum cumulative weight (total remaining probability mass)
+    float32 sum = *std::max_element(weight_cumsums_.begin(), weight_cumsums_.end());
```
caciolai: this might not be needed

native/src/fairseq2n/data/sample_data_source.cc
```diff
@@ -212,9 +237,10 @@ sample_data_source::block(std::size_t idx)
...
 
-    if (!are_close(sum, 1.0F)) {
+    if (!are_close(sum, 1.0F) && sum > 0.0F) {
```
caciolai: this might not be needed

native/src/fairseq2n/data/sample_data_source.cc
```diff
@@ -170,9 +175,26 @@ sample_data_source::random_pipeline_index()
...
 
+    // The binary search returns the index where cumsum[idx] >= sample and cumsum[idx-1] < sample
+    // If this pipeline is blocked, it means cumsum[idx] == cumsum[idx-1], which is a
+    // degenerate interval. We need to search forward to find the pipeline that actually
+    // owns this probability mass (the next unblocked one with cumsum[idx] > cumsum[idx-1])
+    while (lptr < weight_cumsums_.size() && is_blocked_[lptr]) {
+        lptr++;
+    }
+
+    // If we went past the end, wrap around and search from the beginning
```
caciolai: this should not be needed

native/src/fairseq2n/data/sample_data_source.cc
```diff
@@ -20,14 +20,11 @@
...
 
-sample_data_source::sample_data_source(
-    std::vector<data_pipeline> &&pipelines,
-    std::vector<float32> &&weights,
-    std::optional<std::uint64_t> maybe_seed,
-    bool allow_repeats)
-  : pipelines_(std::move(pipelines)), 
-    is_epoch_done_(pipelines_.size()),
-    allow_repeats_{allow_repeats}
+sample_data_source::sample_data_source(std::vector<data_pipeline> &&pipelines,
```
cbalioglu: Do you mind keeping the formatting consistent with the rest of the code?

tests/unit/data/data_pipeline/test_sample.py
```diff
@@ -253,3 +254,23 @@ def test_op_saves_and_restores_state_when_allow_repeats_false_is_specified(
             assert list(pipeline) == [3, 4, 7, 8, 9, 10, 11, 12]
 
             pipeline.reset(reset_rng=True)
+
+    def test_op_works_with_many_pipelines_when_allow_repeats_false(self) -> None:
+        """Test edge case with many pipelines where only lowest-weight ones remain."""
+
+        # See https://github.com/fairinternal/fairseq2-ext/issues/181
```
cbalioglu: Please remove any internal references in OSS code.

tests/unit/data/data_pipeline/test_sample.py
```diff
@@ -253,3 +254,23 @@ def test_op_saves_and_restores_state_when_allow_repeats_false_is_specified(
...
+            pipelines.append(read_sequence(list(range(s))).and_return())
+
+        builder = DataPipeline.sample(pipelines, weights, seed=123, allow_repeats=False)
+        pipeline = builder.and_return()
+
+        expected_total_size = np.sum(sizes)
+        actual_sampled_size = len([_ for _ in iter(pipeline)])
+        assert (
+            actual_sampled_size == expected_total_size
+        ), f"Expected {expected_total_size}, got {actual_sampled_size}"
```
cbalioglu: The assertion text is not really needed since this is a pytest test.

tests/unit/data/data_pipeline/test_sample.py
```diff
@@ -253,3 +254,23 @@ def test_op_saves_and_restores_state_when_allow_repeats_false_is_specified(
...
+        sizes = np.random.RandomState(0).randint(0, 100_000, nb)
+        pipelines = []
+        for s in sizes:
+            pipelines.append(read_sequence(list(range(s))).and_return())
+
+        builder = DataPipeline.sample(pipelines, weights, seed=123, allow_repeats=False)
+        pipeline = builder.and_return()
+
+        expected_total_size = np.sum(sizes)
+        actual_sampled_size = len([_ for _ in iter(pipeline)])
```
cbalioglu: Nit: `len(list(pipeline))`

native/src/fairseq2n/data/sample_data_source.cc
```diff
@@ -170,9 +175,16 @@ sample_data_source::random_pipeline_index()
             rptr = mptr;
     }
 
+    // The binary search returns the index where cumsum[idx] >= sample and cumsum[idx-1] < sample
+    // If this pipeline is blocked, it means cumsum[idx] == cumsum[idx-1], which is a
+    // degenerate interval. We need to search forward to find the pipeline that actually
+    // owns this probability mass (the next unblocked one with cumsum[idx] > cumsum[idx-1])
+    while (lptr < weight_cumsums_.size() && is_blocked_[lptr]) {
```
cbalioglu: As discussed offline, this can be reduced to:

```python

if (!allow_repeats_) {
  while (is_epoch_done_[lptr])
    lptr++;
}
```

I would also put `lptr < weight_cumsums_.size()` optionally as an assert statement since due to `are_all_done()` check, `lptr` should never move beyond the range of `is_epoch_done_`.

native/src/fairseq2n/data/sample_data_source.cc
```diff
@@ -170,6 +170,17 @@ sample_data_source::random_pipeline_index()
...
+    // Binary search finds an index where cumsum[idx] >= sample.
+    // When allow_repeats_ is false, exhausted pipelines have cumsum[idx] == cumsum[idx-1].
+    // Skip forward to find the next active pipeline that "owns" this probability mass.
+    if (!allow_repeats_) { 
+        while (lptr < is_epoch_done_.size() && is_epoch_done_[lptr]) lptr++;
+    }
+
+    // Should never happen since at least one pipeline is active 
+    // (guaranteed by are_all_done() check in next())
+    assert(lptr < weight_cumsums_.size());
```
cbalioglu: I suggest moving this within the loop and removing the `lptr < ..` check from the loop condition. We likely won't reach this assert anyways since `is_epoch_done_[lptr]` will cause a failure (e.g. SEGFAULT) while looping.

doc/source/reference/fairseq2.data_type.rst
```diff
@@ -8,5 +8,4 @@ fairseq2.data_type
 Functions
 =========
 
-.. autofunction:: current_dtype
```
cbalioglu: This is a "bug" on my end. Let me fix it in another PR and rebase this PR on latest main afterwards.

native/src/fairseq2n/data/sample_data_source.cc
caciolai: like so?
```

        while (is_epoch_done_[lptr]) {
            lptr++;
            // Should never happen since at least one pipeline is active 
            // (guaranteed by are_all_done() check in next())
            assert(lptr < is_epoch_done_.size());
        }
```
just for my curiosity, what's the difference? is it "cosmetic" or does it have practical ramifications?

cbalioglu: It has a semantic difference. If you include the `lptr < ..` check within the loop condition, it implies that this is an expected condition. When you put it as an assert statement within the loop, you make it clear that this is an invariant and should never happen unless the implementation has a bug in it. So instead of silently moving out of the loop when that condition is met, we stop the process with an assertion error.

caciolai: Thanks for explaining!

cbalioglu: No problem!

PR #1479, opened 2025-12-10
HG - Hardware-specific sharding test and instructions with torchrun

rsyue:
When more than 1 GPU present, model sharding can be tested with `torchrun --nproc-per-node 8 test_hg_factory.py`

None aware of

**Check list:**

### DIFF
--- a/src/fairseq2/models/hg/__init__.py
+++ b/src/fairseq2/models/hg/__init__.py
@@ -20,7 +20,6 @@ from fairseq2.models.hg.api import load_seq2seq_lm as load_seq2seq_lm
 from fairseq2.models.hg.config import HG_FAMILY as HG_FAMILY
 from fairseq2.models.hg.config import HuggingFaceModelConfig as HuggingFaceModelConfig
 from fairseq2.models.hg.config import register_hg_configs as register_hg_configs
-
 from fairseq2.models.hg.converter import HuggingFaceConfig as HuggingFaceConfig
 from fairseq2.models.hg.converter import HuggingFaceConverter as HuggingFaceConverter
 from fairseq2.models.hg.converter import (


src/fairseq2/data/tokenizers/hg.py
```diff
@@ -78,20 +78,25 @@ def load_hg_token_model(
...
 
         vocab_info = VocabularyInfo(
             vocab_size,
-            unk_idx=maybe_index(unk_token),
-            bos_idx=maybe_index(bos_token),
-            eos_idx=maybe_index(eos_token),
-            pad_idx=maybe_index(pad_token),
-            boh_idx=maybe_index(boh_token),
-            eoh_idx=maybe_index(eoh_token),
+            unk_idx=maybe_index(unk_token, token_type="unk"),
```
artemru: why this change ?

rsyue: This stems from @zyaoj's draft hg model integration 240440c3

tldr; we are specifically mapping tokens to ids to avoid missing key errors in the tokenizer.

As far as I can tell, it's a bit of a fallback mechanism for special token indices when they're not explicitly provided. We want to make sure that special tokens that we don't know in advance can be recognized and won't throw errors for hg models (of which there are many). If someone's special token is `spk1=<|speaker_1_start|>`, then we won't get errors when loading the tokenizer.

Before:                                          
```
def maybe_index(token: str | None) -> int | None:
    if token:
        return tok.convert_tokens_to_ids(token)
    return None
# Called as:
pad_idx=maybe_index(pad_token)
boh_idx=maybe_index(boh_token)
eoh_idx=maybe_index(eoh_token)
```

After:
```
def maybe_index(token: str | None, token_type: str | None) -> int | None:
    """Get token index from explicit token or tokenizer's default."""
    if token:
        return tok.convert_tokens_to_ids(token)
    if token_type:
        token_id_attr = f"{token_type}_token_id"
        return getattr(tok, token_id_attr, None)
    return None
# Called as:
unk_idx=maybe_index(unk_token, token_type="unk")
pad_idx=maybe_index(pad_token, token_type="pad")
boh_idx=maybe_index(boh_token, token_type="boh")
eoh_idx=maybe_index(eoh_token, token_type="eoh")
```

According to Claude:

The new token_type parameter enables automatic fallback to the tokenizer's built-in special tokens:

  1. If an explicit token is provided (e.g., pad_token="<pad>"): Converts it to its ID
  2. If no explicit token is provided but token_type is specified: Falls back to the HuggingFace tokenizer's default special token by accessing attributes
 like tok.unk_token_id, tok.pad_token_id, tok.bos_token_id, etc.
  3. If neither is provided: Returns None

This makes the code more flexible and robust when working with different HuggingFace tokenizers that may have pre-configured special tokens. Instead of requiring users to explicitly specify every special token, it can use the tokenizer's defaults automatically.

rsyue:
I cleaned this up a bit. Things had been moved to models/utils/ to avoid name conflicts with 'hg'. Everything now in hg alongside module rename. Safe serialization removed. Tested working with gpt2, Qwen2.5 Omni, and gemma2.

examples/README.md
cirquit, line 1: Looks like this file should live in the docs somewhere?

examples/simple_gemma_sft.py
cirquit, line 1: Dont think this should be included in the PR, is it possible to reproduce HG-based training with the basic `lm` recipe?

hg_convergence_tests/$HOME/fairseq2/hg_hardware_test/slurm_logs/comparison/comparison-2663617.err
cirquit, line 1: Should not be commited.

hg_convergence_tests/$HOME/fairseq2/hg_hardware_test/slurm_logs/comparison/comparison-2663617.out
cirquit, line 1: Should not be commited.

hg_convergence_tests/compare_checkpoints_simple.py
cirquit, line 1: Verifying artifact's integrity is not on our plate, if hg gives us wrong weights we can't do anything about it. I'd not include this as part of the PR (goes for all of the other verification scripts as well - it would be optimal to have useful unit/integration tests that are extremely fast and have minimal external dependencies).

`tests/integration` are gated by default so maybe they might be a good place, but testability of this PR is not a trivial task, so it would be good to have a good solution to find out when stuff breaks (e.g. hg API/dependencies).  Open for suggestions, but having a `hg_converged_tests` directory is not clean enough.

recipes/lm/sft/configs/gemma_3_1b_it_gsm8k.yaml
cirquit, line 1: This is a good entry point that should be referenced in the documentation if possible.

recipes/lm/common.py
```diff
@@ -27,10 +27,20 @@ def check_model_vocabulary(context: RecipeContext) -> None:
...
         raise RecipeError(
             f"Number of embeddings in the model ({embed.num_embeddings}) is less than the vocabulary size of the tokenizer ({vocab_info.size})."
         )
+    elif vocab_diff == 1:
+        # Off-by-one is acceptable (common with HF models due to power-of-2 padding)
+        log.info(
+            "Model has {} embeddings, tokenizer has {} tokens (off-by-one, likely due to padding).",
+            embed.num_embeddings,
+            vocab_info.size,
+        )
```
cirquit: So whats the resolution here? My guess is that we simply redefine the vocab_size as the embedding_dim has actual weights behind it, but how is the initialization working then?

src/fairseq2/data/parquet/__init__.py
```diff
@@ -21,7 +21,7 @@
     NamedColumns,
     ParquetFragmentLoader,
 )
-from fairseq2.data.parquet.fragment_streaming import (
+from fairseq2.data.parquet.fragment_streaming import (  # noqa: F401
```
cirquit: Can you check if you installed the stubs requirements for the `# noqa: F401` to be still necessary? `pip install -r requirements-devel.txt`

src/fairseq2/data/tokenizers/hg.py
cirquit: My understanding here is that if we want to interface with huggingface, we will need to have a lot of fallbacks and defensive programming inside fairseq2 as not all models are defined similar enough/correct. As we can't force updates on their side, we will need to make some assumption, and this function simply tries to access the class member based on this name if it exists.

```python
token_id_attr = f"{token_type}_token_id"
return getattr(tok, token_id_attr, None)
```

src/fairseq2/generation/beam_search/generator.py
```diff
@@ -632,7 +632,10 @@ def _step(self) -> bool:
 
         # We split the batch by `beam_sizes` and treat each beam separately.
         for beam_idx, (beam_lprobs, beam_step_scores) in enumerate(
-            zip(lprobs.split(self._beam_sizes), self._step_scores.split(self._beam_sizes))  # fmt: skip
+            zip(
+                lprobs.split(self._beam_sizes),
+                self._step_scores.split(self._beam_sizes),
+            )  # fmt: skip
```
cirquit: This fmt skip does not work, revert back please.

src/fairseq2/models/hg_qwen_omni/adapter.py
```diff
@@ -0,0 +1,352 @@
...
+        """
+        # Try common locations for embedding layers
+        for attr_path in [
+            "embed_tokens",  # Llama, Mistral, etc.
+            "model.embed_tokens",  # Some wrapped models
+            "wte",  # GPT-2
+            "transformer.wte",  # GPT-2 variants
+            "embeddings.word_embeddings",  # BERT-style
+            "model.embeddings.word_embeddings",
+        ]:
```
cirquit: Oh wow this is fun

src/fairseq2/models/hg_qwen_omni/config.py
```diff
@@ -0,0 +1,140 @@
...
+    """Custom model class name for special cases."""
+
+    custom_processor_class: str | None = None
+    """Custom processor class name for special cases."""
+
+    trust_remote_code: bool = False
+    """Whether to trust remote code for custom architectures."""
+
+    dtype: str = "auto"
+    """PyTorch dtype to use ('auto', 'float16', 'bfloat16', etc.)."""
```
cirquit: Why not use `torch.dtype`? This auto-serializes to the string representation.

src/fairseq2/models/hg/factory.py
```diff
@@ -0,0 +1,589 @@
...
+    :raises: HuggingFaceModelError: If model loading fails
+    :raises: NotSupportedError: If transformers library is not available
+    """
+
+    if torch.cuda.is_available():
+        default_gangs = get_default_gangs()
+        with default_gangs:
+            gangs = get_current_gangs()
+
+        gangs = get_current_gangs()
```
cirquit, line 150: Not 100% here, but I think this can be resolved by just using `get_current_gangs`, this should handle cuda and everything else under the hood.

cirquit: Same pattern in the test (test_hg_factory.py)

src/fairseq2/models/hg_qwen_omni/factory.py
```diff
@@ -0,0 +1,589 @@
...
+    """
+    Factory for creating HuggingFace models.
+
+    This class handles the logic of loading HuggingFace models,
+    including device placement, dtype conversion, custom classes, and
+    error handling.
+
+    Args:
+        config: The HuggingFace model configuration
+    """
```
cirquit: Mixing docstring styles in the same file. This is one of the correct ones, please refactor the PR to only use sphinx-style docstrings: `:param name:, :returns:, :raises:` OR just a single string, not `Args:`

src/fairseq2/models/jepa/interop.py
cirquit, line 1: Revert formatting, fix your linter to respect `# fmt: off / on`. Same for all interop.py's please.

src/fairseq2/models/wav2vec2/position_encoder.py
cirquit, line 1: Redo formatting.

Please check all files, this happens multiple times.

src/fairseq2/recipe/base.py
```diff
@@ -49,7 +62,6 @@
 from fairseq2.recipe.tokenizer import RecipeTokenizer
 from fairseq2.runtime.dependency import DependencyContainer, DependencyResolver
 from fairseq2.task import Task
-from fairseq2.trainer import Trainer, TrainUnit
```
cirquit, line 52: Removing public exports will fail, if this was due to circular importing, this needs to be solved elsewhere, but not in the base.py class.

src/fairseq2/recipe/base.py
```diff
@@ -66,6 +78,10 @@
 
 @final
 class RecipeContext:
+    """
+    Represents the underlying framework of a machine learning workflow
+    """
+
```
cirquit:
```suggestion
```

Too nondescript of a comment.

src/fairseq2/__init__.py
```diff
@@ -8,7 +8,7 @@
 
 from collections.abc import Callable
 
-import fairseq2n  # Report any fairseq2n initialization error eagerly.
+import fairseq2n  # noqa: F401  # Report any fairseq2n initialization error eagerly.
```
cirquit: Same as before, I think there's a linting mismatch between your local vs CI.

src/fairseq2/validator.py
```diff
@@ -9,19 +9,21 @@
...
+from typing import TYPE_CHECKING, Any, Final, TypeVar, final
 
 import torch
 from torch import Tensor
 from torch.profiler import record_function
 from typing_extensions import override
 
+if TYPE_CHECKING:
+    from fairseq2.evaluator import EvalUnit
+
```
cirquit: Same error as in base.py propagated here.

.coveragerc
cirquit, line 1: What is this file?

HG_ADAPTER_README.md
cirquit, line 1: Should not be commited

src/fairseq2/models/hg_qwen_omni/factory.py
```diff
@@ -0,0 +1,589 @@
...
+    # Import and load the model class
+    model_class_name = model_info["model_class"]
+    try:
+        model_class = _import_class_from_transformers(model_class_name)
+        try:
+            model = model_class.from_pretrained(
+                **load_kwargs, attn_implementation="flash_attention_2"
+            )
+            print("Using flash_attention_2.")
+        except Exception:
```
cirquit: No `print` statements in library code, we use a logger. Won't mark all manually, please find them yourself.

src/fairseq2/models/hg_qwen_omni/factory.py
```diff
@@ -0,0 +1,589 @@
...
+    config: HuggingFaceModelConfig,
+    model_info: Dict[str, str],
+    gangs: Gangs | None = None,
+) -> Any:
+    """Load a model using special/custom classes."""
+
+    log.info(f"Loading special model '{name}' using custom classes")
+
+    local_rank = int(os.environ.get("LOCAL_RANK", 0))
+    torch.cuda.set_device(local_rank)
```
cirquit: Why are we not using the gang primitive here? There should be no calls to `os.environ.get` in model implementations.

recipes/lm/common.py
cirquit: I think this should be configurable, but fail clearly if this does not fit. These one-off errors are a pain to debug and find out that library code does something this unexpected.

src/fairseq2/models/hg_qwen_omni/__init__.py
cirquit, line 1: Why is this directory named after a model while it should be used generically to load, e.g. gpt-2? Maybe rename to `hg` or `hg_adapter` fits better?

src/fairseq2/models/hg_qwen_omni/factory.py
```diff
@@ -0,0 +1,589 @@
...
+
+    :param gangs: The gangs to use when sharding
+
+    :returns: The sharded model with replaced layers
+    """
+
+    qkv_pattern_c = re.compile("[.]q_|[.]k_|[.]v_|_[qkv]$|[.][qkv]$")
+    out_pattern_c = re.compile("out|[.]o_|_o$|[.]o$|[.]proj$")
+
+    fs_model = copy.copy(model)
```
cirquit: This looks like a shallow copy, is this ok?

src/fairseq2/models/hg/tokenizer.py
```diff
@@ -0,0 +1,248 @@
...
+    """The end-of-sequence token."""
+
+    pad_token: str | None = None
+    """The padding token."""
+
+    boh_token: str | None = None
+    """The beginning-of-head token."""
+
+    eoh_token: str | None = None
+    """The end-of-head token."""
```
cirquit, line 53: Are you sure these are strings? Typically, you use indices in the vocabulary to represent tokens. This also holds for the properties below (should be int instead of str afaik).

src/fairseq2/models/hg_qwen_omni/config.py
```diff
@@ -0,0 +1,140 @@
...
+    """Register predefined HuggingFace model configurations.
+
+    This function registers several predefined configurations for common
+    HuggingFace models, making them available through the fairseq2 model
+    hub system.
+
+    :param container: The dependency container to register configurations with
+
+    Registered Configurations:
+        - `Stack Overflow home <https://stackoverflow.com/>`_
```
cirquit: What is stack overflow doing in source code of a sequence modeling framework 😸

src/fairseq2/models/hg_qwen_omni/__init__.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+
+Modules:
+    api: High-level API functions for model and tokenizer loading
+    config: Configuration classes for HuggingFace models
+    factory: Factory classes for model creation and loading
+    hub: Integration with fairseq2's model hub system
+    tokenizer: HuggingFace tokenizer integration
+
+For detailed documentation, see: :doc:`/reference/fairseq2.models.hg_qwen_omni
```
cirquit:
```suggestion
For detailed documentation, see: :doc:`/reference/fairseq2.models.hg_qwen_omni`
```

src/fairseq2/models/hg_qwen_omni/tokenizer.py
```diff
@@ -0,0 +1,248 @@
...
+    def decode(
+        self, token_indices: Tensor, *, skip_special_tokens: bool = False
+    ) -> str:
+        decoder = self.create_decoder(skip_special_tokens=skip_special_tokens)
+        return decoder(token_indices)
+
+    # TODO: maybe a better way is just fall back to properties of
+    # self._model._tok if it's not overridden in this fs2 Tokenizer class..?
+    def convert_tokens_to_ids(self, tokens: list[str] | str) -> int | list[int]:
+        return self._model._tok.convert_tokens_to_ids(tokens)
```
cirquit: Is this a followup issue from keeping the tokens as str?

src/fairseq2/models/hg_qwen_omni/adapter.py
```diff
@@ -0,0 +1,352 @@
...
+            try:
+                parts = attr_path.split(".")
+                obj = hf_model
+                for part in parts:
+                    obj = getattr(obj, part)
+                # Found it!
```
cirquit:
```suggestion
```

src/fairseq2/models/hg_qwen_omni/adapter.py
```diff
@@ -0,0 +1,352 @@
...
+                return obj
+            except AttributeError:
+                continue
+
+        # Fallback: return the model itself as a placeholder
+        # This will likely fail vocabulary checks, but at least won't crash
```
cirquit: I think this should fail loudly. If the model is unusable, I want to know it.

src/fairseq2/composition/models.py
```diff
@@ -172,7 +179,7 @@ def register_model_family(
 ) -> None:
     if hg_exporter is not None:
         _warn_deprecated(
-            "`hg_exporter` is deprecated and will be removed in v0.9. Use `fairseq2.models.hg.HuggingFaceConverter` instead."
+            "`hg_exporter` is deprecated and will be removed in v0.9. Use `fairseq2.models.hg_qwen_omni.HuggingFaceConverter` instead."
```
cirquit: Update this to the correct import path after refactoring

.gitignore
```diff
@@ -5,6 +5,16 @@ build/
...
+# Testing files and cleanup scripts
+doc/rebuild.sh
+src/all_layers.txt
+src/equiv_fs2.py
+src/github_private_tests/
+src/qwen_omni_arch.txt
+src/replace_mha.py
+src/hg_load_test.py
+src/launch.bash
+
```
cirquit: Private files, please remove.

.gitignore
```diff
@@ -21,3 +31,9 @@ __pycache__/
 
 # Other
 .DS_Store
+
+# coverage
+.coverage
+
+# mypy
+.mypy_cache/
```
cirquit: Same for this please.

recipes/lm/common.py
cirquit: After thinking on this, this is really bad. If the sentencepiece tokenizer has a larger vocab_size during counting compared to its init means that it has additional tokens that literally may not have a representation in the embedding dimension. We must fail here and request flexible handling (your current pathway) and report information that helps determine the gravity (total vocab size, all key token ids, etc).

rsyue: This was actually really simple. It was checking len(tokenizer) instead of tokenizer.vocab_size.

src/fairseq2/models/hg_qwen_omni/__init__.py
rsyue: This was left to avoid conflicts with hg.py, that included legacy converters. But that file was small so I looped it into the larger commit

src/fairseq2/models/hg_qwen_omni/factory.py
rsyue: It did work correctly because we discard original model, but updated to cleaner implementation iterating through layer list

src/fairseq2/models/hg/tokenizer.py
rsyue: These are tokens that are being set to IDs, so here we would enter <UNK>

However, the point is valid, because the _id functions did return strings whereas they should be int idx

src/fairseq2/models/hg_qwen_omni/config.py
rsyue: Was reverted. This must have been hallucinated when applying the linting.

src/fairseq2/models/hg_qwen_omni/tokenizer.py
rsyue: Comment not needed. After applying getattr and fixing ID int, no longer relevant

examples/README.md
rsyue: Done

examples/simple_gemma_sft.py
rsyue: Examples removed

hg_convergence_tests/$HOME/fairseq2/hg_hardware_test/slurm_logs/comparison/comparison-2663617.err
rsyue: err and out files are removed from the commit

hg_convergence_tests/$HOME/fairseq2/hg_hardware_test/slurm_logs/comparison/comparison-2663617.out
rsyue: Same as above

hg_convergence_tests/compare_checkpoints_simple.py
rsyue: Convergence tests dir removed

recipes/lm/sft/configs/gemma_3_1b_it_gsm8k.yaml
rsyue: Comprehensive explanation added.

src/fairseq2/data/parquet/__init__.py
rsyue: Removed with linting revert

src/fairseq2/generation/beam_search/generator.py
rsyue: When reverting, this matches main (all on one line). Is that what needed to change? The # fmt: skip was already there.

src/fairseq2/models/hg_qwen_omni/adapter.py
rsyue: Replaced with cleaner refactor

src/fairseq2/models/hg_qwen_omni/config.py
rsyue: Done

src/fairseq2/models/hg/factory.py
rsyue: Done

src/fairseq2/models/hg_qwen_omni/factory.py
rsyue: Done

src/fairseq2/models/jepa/interop.py
rsyue: Done

src/fairseq2/models/wav2vec2/position_encoder.py
rsyue: All clear

src/fairseq2/recipe/base.py
rsyue: done

src/fairseq2/recipe/base.py
rsyue: Done

src/fairseq2/__init__.py
rsyue: Done

src/fairseq2/validator.py
rsyue: Done

.coveragerc
rsyue: Autogenerated and not necessary. Removed.

HG_ADAPTER_README.md
rsyue: All md files except README were removed

src/fairseq2/models/hg_qwen_omni/factory.py
rsyue: Addressed

src/fairseq2/models/hg_qwen_omni/factory.py
rsyue: Done

src/fairseq2/models/hg_qwen_omni/adapter.py
rsyue: Done

src/fairseq2/composition/models.py
rsyue: Done

.gitignore
rsyue: Removed

.gitignore
rsyue: Done

recipes/lm/sft/configs/gemma_3_1b_it_gsm8k_long.yaml
```diff
@@ -0,0 +1,50 @@
...
+  name: null  # Disable default model to use family/arch approach
+  family: "hg"
+  arch: "causal_lm"
+  config_overrides:
+    hf_name: "google/gemma-3-1b-it"
+    model_type: "causal_lm"
+    trust_remote_code: true
+
+tokenizer:
+  # Don't override name - let path take precedence
```
cirquit:
```suggestion
```
Overly verbose.

recipes/lm/sft/configs/gemma_3_1b_it_gsm8k_long.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+regime:
+  num_steps: 100000
+  checkpoint_every_n_steps: 100
+  validate_every_n_steps: 100
+  checkpoint_every_n_data_epochs: 100
+  keep_last_n_checkpoints: 10
+  publish_metrics_every_n_steps: 10
+  save_model_only: false
+  export_hugging_face: false  # Disable HF export for HG models
```
cirquit, line 49: Wait, why?

recipes/lm/sft/README.md
```diff
@@ -6,7 +6,7 @@ This directory contains the implementation for supervised fine-tuning of languag
 
 The SFT recipe is designed to finetune language models to follow instructions by learning from prompt-response pairs.
 
-This recipe supports both Llama and Qwen model families.
+This recipe supports Llama, Qwen, and HuggingFace (`hg`) model families.
```
cirquit:
```suggestion
This recipe supports a variety of autoregressive models, like Llama, Qwen and supported HuggingFace (`hg`) model families.
```

recipes/lm/sft/README.md
```diff
@@ -69,11 +69,28 @@ regime:
...
+```bash
+# Quick test run
+python -m fairseq2.recipes.lm.sft --config-file sft/configs/gemma_3_1b_it_gsm8k.yaml
+
+# Full training
+python -m fairseq2.recipes.lm.sft --config-file sft/configs/gemma_3_1b_it_gsm8k_long.yaml
+```
+
+HuggingFace models use `family: "hg"` and `arch: "causal_lm"` in the config, with the HF model name specified via `config_overrides.hf_name`.
+
```
cirquit:
```suggestion
```
Overly verbose.

src/fairseq2/models/hg/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+from fairseq2.models.hg.api import load_causal_lm as load_causal_lm
+from fairseq2.models.hg.api import load_hg_model_simple as load_hg_model_simple
+from fairseq2.models.hg.api import load_hg_tokenizer_simple as load_hg_tokenizer_simple
+from fairseq2.models.hg.api import load_multimodal_model as load_multimodal_model
+from fairseq2.models.hg.api import load_seq2seq_lm as load_seq2seq_lm
+from fairseq2.models.hg.config import HG_FAMILY as HG_FAMILY
+from fairseq2.models.hg.config import HuggingFaceModelConfig as HuggingFaceModelConfig
+from fairseq2.models.hg.config import register_hg_configs as register_hg_configs
+
+# Re-export converter API (formerly in fairseq2/models/hg.py)
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/adapter.py
```diff
@@ -0,0 +1,341 @@
...
+    def __init__(
+        self,
+        hf_model: Module,
+        max_seq_len: int = 8192,
+        enable_gradient_checkpointing: bool = False,
+    ) -> None:
+        super().__init__(max_seq_len=max_seq_len)
+
+        # Register the HF model as a proper submodule
+        # Use a name that won't conflict with parent class attributes
```
cirquit:
```suggestion
```
Overly verbose, you already decided on this parameter name to never conflict.

src/fairseq2/models/hg/adapter.py
```diff
@@ -0,0 +1,341 @@
...
+        self.add_module("_wrapped_hf_model", hf_model)
+
+        # Store reference to HF model's config
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/adapter.py
```diff
@@ -0,0 +1,341 @@
...
+        self._hf_config: Any = getattr(hf_model, "config", None)
+
+        # Enable gradient checkpointing if requested
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/adapter.py
```diff
@@ -0,0 +1,341 @@
...
+        - BERT: model.embeddings.word_embeddings
+        """
+        # Try common locations for embedding layers
+        for attr_path in self._EMBEDDING_PATHS:
+            try:
+                parts = attr_path.split(".")
+                obj = hf_model
+                for part in parts:
+                    obj = getattr(obj, part)
+                # Found it!
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/adapter.py
```diff
@@ -0,0 +1,341 @@
...
+        raise LookupError(
+            f"Cannot find embedding layer in {type(hf_model).__name__}. "
+            f"Tried: {', '.join(self._EMBEDDING_PATHS)}. "
+            f"Register the model with a known embedding path or add support for this architecture."
+        )
+
+    @property
+    def _hf_model(self) -> Module:
+        """Access the wrapped HuggingFace model."""
+        # Access via _modules to ensure it works after FSDP wrapping
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/adapter.py
```diff
@@ -0,0 +1,341 @@
...
+        :param seqs_layout: Layout information (padding, etc.).
+        :param targets: Target token IDs for training [batch_size, seq_len].
+        :param state_bag: Incremental state (not used for HF models).
+        :param label_smoothing: Label smoothing factor (not implemented).
+        :param target_mask: Mask for targets [batch_size, seq_len].
+        :param reduction: How to reduce the loss (``"sum"`` or ``"mean"``).
+        :param return_logits: Whether to return logits along with loss.
+        :returns: Loss tensor, or tuple of (loss, logits) if ``return_logits`` is ``True``.
+        """
+        # Create attention mask from seqs_layout
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/adapter.py
```diff
@@ -0,0 +1,341 @@
...
+        attention_mask = self._create_attention_mask(seqs, seqs_layout)
+
+        # Prepare kwargs for HuggingFace model
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/adapter.py
```diff
@@ -0,0 +1,341 @@
...
+                    num_targets = (labels != -100).sum()
+                loss = loss * num_targets
+            # If reduction == "mean", loss is already in the right format
+
+            if return_logits:
+                return loss, cast(Tensor, outputs.logits)
+            else:
+                return loss
+        else:
+            # Inference mode - just get logits
```
cirquit:
```suggestion
            # Inference mode returns logits only
```

src/fairseq2/models/hg/adapter.py
```diff
@@ -0,0 +1,341 @@
...
+            # Handle reduction
+            # HF models typically return mean loss, we need to convert based on reduction
+            if reduction == "sum":
+                # Convert from mean to sum
+                if target_mask is not None:
+                    num_targets = target_mask.sum()
+                else:
+                    num_targets = (labels != -100).sum()
+                loss = loss * num_targets
+            # If reduction == "mean", loss is already in the right format
```
cirquit:
```suggestion
            # HF models return mean loss, need to optionally convert to "sum"
            if reduction == "sum":
                if target_mask is not None:
                    num_targets = target_mask.sum()
                else:
                    num_targets = (labels != -100).sum()
                loss = loss * num_targets
```
Slightly less verbose.

src/fairseq2/models/hg/adapter.py
```diff
@@ -0,0 +1,341 @@
...
+            config, "enable_gradient_checkpointing", False
+        )
+
+        return HgCausalLMAdapter(
+            hf_model,
+            max_seq_len=max_seq_len,
+            enable_gradient_checkpointing=enable_gradient_checkpointing,
+        )
+
+    # Not a causal LM, return as-is
```
cirquit:
```suggestion
    # Not a causal LM
```

src/fairseq2/models/hg/adapter.py
```diff
@@ -0,0 +1,341 @@
...
+
+def wrap_hg_model_if_causal_lm(hf_model: Module, config: Any) -> Module:
+    """
+    Wrap a HuggingFace model in HgCausalLMAdapter if it's a causal language model.
+
+    :param hf_model: The HuggingFace model to potentially wrap.
+    :param config: The HuggingFaceModelConfig used to create the model.
+    :returns: Wrapped model if it's a causal LM, otherwise the original model.
+    """
+    # Check if this is a causal LM that should be wrapped
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/api.py
```diff
@@ -0,0 +1,179 @@
...
+        bos_token=bos_token,
+        eos_token=eos_token,
+        pad_token=pad_token,
+        boh_token=boh_token,
+        eoh_token=eoh_token,
+    )
+    return load_hg_tokenizer(Path(name), config)
+
+
+# Convenience aliases for common use cases
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/config.py
```diff
@@ -0,0 +1,119 @@
...
+
+from dataclasses import dataclass
+from typing import Any, Final
+
+from fairseq2.data_type import DataType
+from fairseq2.runtime.config_registry import ConfigRegistrar
+from fairseq2.runtime.dependency import DependencyContainer
+
+HG_FAMILY: Final = "hg"
+"""The family identifier for HuggingFace models in fairseq2."""
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/factory.py
```diff
@@ -0,0 +1,552 @@
...
+        return True
+    return False
+
+
+def _load_auto_model(name: str, config: HuggingFaceModelConfig, hf_config: Any) -> Any:
+    """Load a model using Auto classes."""
+
+    log.info(f"Loading model '{name}' using Auto classes")
+
+    # Prepare kwargs for from_pretrained
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/factory.py
```diff
@@ -0,0 +1,552 @@
...
+    load_kwargs = _prepare_load_kwargs(config)
+
+    # Determine which AutoModel class to use
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/factory.py
```diff
@@ -0,0 +1,552 @@
...
+    except Exception as ex:
+        if _is_multimodal_config(hf_config):
+            raise HuggingFaceModelError(
+                name,
+                f"Model '{name}' appears to be a multimodal model that requires "
+                f"a processor and cannot be loaded via Auto classes. "
+                f"Register it with register_hg_model_class() or use "
+                f"model_type='custom' with the appropriate model class.",
+            ) from ex
+        # Check if this might be an unsupported model
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/factory.py
```diff
@@ -0,0 +1,552 @@
...
+                f"Model '{name}' is not supported by HuggingFace AutoModel.\nError: {str(ex)}.\nPlease refer to https://huggingface.co/docs/transformers/model_doc/auto\nfor supported architectures or register a custom class.",  # fmt: skip
+            ) from ex
+        else:
+            raise
+
+    # Load tokenizer/processor
+    if config.use_processor:
+        try:
+            processor = AutoProcessor.from_pretrained(name)
+            # Set processor to model
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/factory.py
```diff
@@ -0,0 +1,552 @@
...
+        # Check if this might be an unsupported model
+        if "does not appear to have a file named config.json" not in str(ex):
+            raise HuggingFaceModelError(
+                name,
+                f"Model '{name}' is not supported by HuggingFace AutoModel.\nError: {str(ex)}.\nPlease refer to https://huggingface.co/docs/transformers/model_doc/auto\nfor supported architectures or register a custom class.",  # fmt: skip
+            ) from ex
+        else:
+            raise
+
+    # Load tokenizer/processor
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/factory.py
```diff
@@ -0,0 +1,552 @@
...
+    """Determine which AutoModel class to use."""
+
+    if config.model_type == "causal_lm":
+        return AutoModelForCausalLM
+    elif config.model_type == "seq2seq_lm":
+        return AutoModelForSeq2SeqLM
+    elif config.model_type == "auto":
+        return AutoModel
+    else:
+        # Auto-detect based on config
```
cirquit:
```suggestion
        # Try auto-detection based on config
```

src/fairseq2/models/hg/factory.py
```diff
@@ -0,0 +1,552 @@
...
+def _import_class_from_transformers(class_name: str) -> Any:
+    """Import a class from the transformers library."""
+
+    try:
+        # Try importing from transformers directly
+        transformers_module = importlib.import_module("transformers")
+        return getattr(transformers_module, class_name)
+    except AttributeError:
+        # If not found in main module, it might be in a submodule
+        # This is a simplified approach
```
cirquit:
```suggestion
```

src/fairseq2/models/hg/fsdp.py
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# From huggingface/accelerate: src/accelerate/utils/dataclasses.py
```
cirquit, line 7: Minimally more context on this file/the reference to the linked python file. Also not really easy to follow given that I have no idea where accelerate lives. If useful, please add the full link with a short explanation.

src/fairseq2/models/hg/fsdp.py
```diff
@@ -0,0 +1,51 @@
...
+            setattr(model, name, wrapper(module))
+
+
+def apply_fsdp_to_hg_transformer_lm(
+    model: Module, granularity: str, wrapper: FSDPWrapper
+) -> Module:
+    transformer_cls_names_to_wrap = get_transformer_cls_names_to_wrap(
+        model,
+        # NOTE: to override this, need to register and override this
+        # apply_fsdp in the new model family
```
cirquit:
```suggestion
```
Not the place, this fits better in a docstring imo.

src/fairseq2/models/hg/fsdp.py
```diff
@@ -0,0 +1,51 @@
...
+        if len(list(module.children())) > 0:
+            replace_layers(module, transformer_cls_to_wrap, wrapper)
+        if module.__class__.__name__ in transformer_cls_to_wrap:
+            setattr(model, name, wrapper(module))
+
+
+def apply_fsdp_to_hg_transformer_lm(
+    model: Module, granularity: str, wrapper: FSDPWrapper
+) -> Module:
+    transformer_cls_names_to_wrap = get_transformer_cls_names_to_wrap(
```
cirquit, line 47:
```suggestion
) -> Module:
    """If a custom FSDP wrapper is required, register and override
    by setting `apply_fsdp` in the new model family.
    """
    transformer_cls_names_to_wrap = get_transformer_cls_names_to_wrap(
```

src/fairseq2/models/hg/tokenizer.py
```diff
@@ -0,0 +1,206 @@
...
+
+def load_hg_tokenizer(path: Path, config: HgTokenizerConfig) -> HgTokenizer:
+    """
+    Load a HuggingFace tokenizer.
+
+    :param config: Tokenizer configuration
+    :returns: HgTokenizer instance
+    """
+
+    # Load the HuggingFace token model
```
cirquit:
```suggestion
```

src/fairseq2/models/llama/sharder.py
cirquit, line 1: This file was still not reverted due to formatting.

src/fairseq2/models/llama4/sharder.py
cirquit, line 1: Again formatting revert.

src/fairseq2/models/qwen/sharder.py
cirquit, line 1: Formatting revert still needed.

src/fairseq2/nn/fsdp/fsdp1.py
cirquit, line 1: Why does this file require changes for this PR? Especially FSDP1 seems extremely strange.

src/fairseq2/utils/structured.py
cirquit, line 1: Formatting revert still needed.

tests/unit/data/data_pipeline/test_map.py
cirquit, line 1: This seems extremely suspicious, why this change?

recipes/lm/sft/configs/gemma_3_1b_it_gsm8k_long.yaml
rsyue: export_hugging_face is to convert fs2 native models to hg format. It is not needed here

src/fairseq2/models/llama/sharder.py
rsyue: done

src/fairseq2/models/llama4/sharder.py
rsyue: done

src/fairseq2/models/qwen/sharder.py
rsyue: done

src/fairseq2/utils/structured.py
rsyue: done

src/fairseq2/nn/fsdp/fsdp1.py
rsyue: As mentioned below, this was principally formatting. Other prev changes were done by another author in previous commit for aligning with hg adapter.

tests/unit/data/data_pipeline/test_map.py
rsyue: Reverted, attempt at fixing a flaky test

PR #1489, opened 2026-02-03
fix wrong importance sampling ratio logging.

lydiadli:
fix wrong importance sampling ratio metric - current calculation include positions supposed to be masked out including padding and prompt tokens. metric is abnormal even for quite on-policy setup.

Fixes #{issue number}

**Check list:**
- [X] Did you read the [contributor guideline](https://github.com/facebookresearch/fairseq2/blob/main/CONTRIBUTING.md)?
- [X] Did you write any **new necessary tests**?

### DIFF
--- a/src/fairseq2/recipes/lm/_online_finetune/_common.py
+++ b/src/fairseq2/recipes/lm/_online_finetune/_common.py
@@ -746,3 +746,10 @@ def get_parameter_converter(model_config):
         raise RuntimeError(f"{model_config} not supported in online recipe")
 
     return _convert_parameter
+
+
+def masked_mean(x, mask, eps=1e-8):
+    # mask: boolean tensor with same shape as x
+    masked_x = x * mask
+    denom = mask.sum().clamp_min(eps)
+    return masked_x.sum() / denom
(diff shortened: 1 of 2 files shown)


src/fairseq2/recipes/lm/_online_finetune/_common.py
```diff
@@ -746,3 +746,10 @@ def get_parameter_converter(model_config):
         raise RuntimeError(f"{model_config} not supported in online recipe")
 
     return _convert_parameter
+
+
+def masked_mean(x, mask, eps=1e-8):
```
uralik, line 751: this eps value wont work with fp16 and will lead to funny numbers
what you can do is to use x dtype smallest value that pytorch can generate for you

lydiadli: thanks for the thoughtfulness! this case should be ok actually i think. because whenever denominator is indeed clamped, mask only has zeros. so the masked_x (numerator) is zero as well.  it would be torch.tensor([0], dtype=torch.float16) / 1e-8 or similar for other datetypes :)

PR #1490, opened 2026-02-04
using pyarrow.dataset interface instead of pq.ParquetDataset

artemru:
using `pyarrow.dataset.dataset` instead of `pq.ParquetDataset` can be much faster for deep nested partitioned dataset 
since the later creates global partition expressions which is costly.
it also makes the fragments serialization size much smaller.

Simple illustrative example
```python
import pyarrow as pa
import numpy as np
import pyarrow.parquet as pq
import pyarrow.dataset as ds
parquet_path = "/tmp/part_test/"
n_rows = 100_000
table = pa.table({
    "id": np.arange(n_rows),
    "value": np.random.randn(n_rows),
    "year": np.random.choice(["2020", "2021", "2022"], n_rows),
    "lang": np.random.choice([f"lang_{i:02d}" for i in range(20)], n_rows),
    "domain": np.random.choice([f"domain_{i:02d}" for i in range(10)], n_rows),
})

pq.write_to_dataset(table, root_path=parquet_path, partition_cols=["year", "lang", "domain"])

%time pq_dataset = pq.ParquetDataset(parquet_path)
%time pa_dataset = ds.dataset(parquet_path, partitioning="hive")

frag_old = list(pq_dataset._dataset.get_fragments())[0]
new_old = list(pq_dataset._dataset.get_fragments())[0]
print(frag_old.partition_expression)
print(new_old.partition_expression)
```

Good question. 
It should remain compatible at highlevel API.
I also added a backward compatible casting... but it maybe not all

**Check list:**

### DIFF
--- a/src/fairseq2/data/parquet/fragment_streaming/primitives.py
+++ b/src/fairseq2/data/parquet/fragment_streaming/primitives.py
@@ -13,6 +13,7 @@ from typing import Any, Dict, Iterator, List, Optional
 import numpy as np
 import pyarrow as pa
 import pyarrow.compute as pc  # noqa: F401
+import pyarrow.dataset as ds
 import pyarrow.parquet as pq
 import torch
 from pyarrow.dataset import get_partition_keys
@@ -25,7 +26,6 @@ from fairseq2.data.data_pipeline import (
 from fairseq2.data.parquet.fragment_streaming.config import ParquetDatasetLimitOptions
 from fairseq2.data.parquet.utils import (
     circular_shift_left,
-    get_dataset_fragments,
     split_fragment_in_row_groups,
 )
 from fairseq2.logging import log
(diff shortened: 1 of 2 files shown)


src/fairseq2/data/parquet/fragment_streaming/primitives.py
```diff
@@ -67,26 +67,79 @@ def process_filter(
...
+        dataset=parquet_ds._dataset,
+        partition_filters=parquet_ds._filter_expression,
+    )
+
+
 def init_parquet_dataset(
     parquet_path: str | List[str],
-    partition_filters: Optional[pa.dataset.Expression] = None,
+    partition_filters: Optional[pa.compute.Expression] = None,
     filesystem=None,
```
caciolai, line 121:
```suggestion
    filesystem=None,
    partitioning: str = "hive",
```

src/fairseq2/data/parquet/fragment_streaming/primitives.py
```diff
@@ -67,26 +67,79 @@ def process_filter(
...
     Returns:
-        pq.ParquetDataset: The initialized Parquet dataset.
+        ParquetDatasetWrapper: A wrapper containing the dataset and partition filters.
     """
-    return pq.ParquetDataset(
-        parquet_path, filters=partition_filters, filesystem=filesystem
+    dataset = ds.dataset(
+        parquet_path,
+        format="parquet",
+        partitioning="hive",
```
caciolai, line 139:
```suggestion
        partitioning=partitioning,
```

src/fairseq2/data/parquet/fragment_streaming/primitives.py
caciolai: actually param can be more complex: https://arrow.apache.org/docs/python/generated/pyarrow.dataset.dataset.html

artemru: yes, I know but now i want to keep it simple for BC, next we can add other args we there's a real usecase.

PR #1492, opened 2026-02-09
Add an optional cache for ParquetFragmentStreamer

caciolai:
- When instantiating N data pipelines across different partitions of the same Parquet Dataset, currently `ParquetFragmentStreamer` will initialize N separate [`Dataset`](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.Dataset.html) instances, with much duplicated work in filesystem scanning metadata parsing etc

- The cache mechanism is controlled by a flag set to False by default
- Previously existing code path(s) are not affected

**Check list:**

### DIFF
--- a/src/fairseq2/data/parquet/fragment_streaming/primitives.py
+++ b/src/fairseq2/data/parquet/fragment_streaming/primitives.py
@@ -5,11 +5,10 @@
 # LICENSE file in the root directory of this source tree.
 
 import concurrent.futures
-import os
 from dataclasses import dataclass
 from functools import _CacheInfo as CacheInfo
 from functools import lru_cache, reduce
-from typing import Any, Dict, Iterator, List, NamedTuple, Optional
+from typing import Any, Dict, Iterator, List, NamedTuple, Optional, Tuple
 
 import cloudpickle
 import numpy as np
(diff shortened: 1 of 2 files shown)


src/fairseq2/data/parquet/fragment_streaming/primitives.py
```diff
@@ -44,6 +45,99 @@ def batched(iterable, n):
...
+
+        filesystem_serialized = cloudpickle.dumps(filesystem)
+
+        return cls(
+            path=normalized_path,
+            filesystem=filesystem_serialized,
+        )
+
+
+@lru_cache(maxsize=1)  # TODO: decide on reasonable upper bound
```
artemru, line 97: i would pass max_cache_size as a param here instead of use_cache

artemru, line 97: i would pass max_cache_size as a param here instead of use_cache

caciolai: yes but how can you set dynamically (tied to an object instance) a singleton component that you want to exist across instances?

artemru: maybe as some global context...

caciolai: what if different `ParquetFragmentStreamer` instances ask for different cache sizes? or were you thinking of something like AssetStore i.e. a singleton handled by fairseq2 and controlled via the recipe? that would be a deeper change though I think, as I would need to touch recipe composition

PR #1510, opened 2026-05-24
fix(data): align dynamic_bucket kwarg between Python stub and pybind

### LINKED ISSUE #1103: `dynamic_bucket` Python signature seems to be wrong
**Describe the bug:**

When calling `dynamic_bucket` on a data pipeline, I am getting an "incompatible function arguments" error.
```
TypeError: dynamic_bucket(): incompatible function arguments. The following argument types are supported:
    1. (self: fairseq2n.bindings.data.data_pipeline.DataPipelineBuilder, threshold: float, fn: Callable[[Any], float], bucket_creation_fn: Optional[Callable[[List[Any]], Tuple[List[Any], List[Any]]]] = None, min_num_examples: Optional[int] = None, max_num_examples: Optional[int] = None, drop_remainder: bool = False) -> fairseq2n.bindings.data.data_pipeline.DataPipelineBuilder

Invoked with: <fairseq2n.bindings.data.data_pipeline.DataPipelineBuilder object at 0x7f76c41daaf0>; kwargs: threshold=5.0, cost_fn=<function my_cost_fn at 0x7f76a7f703a0>
```

**Describe how to reproduce:**

```Python
from typing import Any
from fairseq2.data import read_sequence

def my_cost_fn(item: Any) -> float:
    return float(item)

items = list(range(10))

builder = read_sequence(items)
builder.dynamic_bucket(threshold=5.0, cost_fn=my_cost_fn)
pipeline = builder.and_return()
bucketed = list(pipeline)
print(bucketed)

assert bucketed == [[0, 1, 2, 3], [4, 5], [6], [7], [8], [9]]
```

**Describe the expected behavior:**
I expected no error to happen and assertions to pass.

**Environment:**
I am using `fairseq2==0.4.4`, but the error seems to persist even in the main.

**Additional Context:**
What is confusing is that I am using exactly the signature described in the Python interface (https://github.com/facebookresearch/fairseq2/blob/19aeff12c16229dfa55e9c4688a585e0ece9e548/src/fairseq2/data/_data_pipeline.py#L238) and in the documentation (https://facebookresearch.github.io/fairseq2/nightly/basics/data_pipeline.html#bucketing). But the underlying C++ code (https://github.com/facebookresearch/fairseq2/blob/19aeff12c16229dfa55e9c4688a585e0ece9e548/native/src/fairseq2n/data/dynamic_bucket_data_source.h#L24) expects the argument to be called just `fn`. 

Please reconcile these interfaces. Having to consult the native code for writing simple stuff is annoying.

Ace3Z:
Closes #1103.

Calling `pipeline.dynamic_bucket(threshold=..., cost_fn=...)` (the form the Python stub at `src/fairseq2/data/data_pipeline.py:238` declares, and what Sphinx generates docs from) fails with `TypeError: incompatible function arguments`. The C++ pybind binding at `native/python/src/fairseq2n/bindings/data/data_pipeline.cc:503` exposed the arg as `py::arg("fn")`.

Renamed the pybind kwarg from `fn` to `cost_fn` so it matches the documented Python signature. `cost_fn` is also the C++ type alias name in `native/src/fairseq2n/data/data_pipeline.h`. Every internal caller (parquet table bucketing, existing tests) passes the cost function positionally, so this is source compatible.

Added a regression test in `tests/unit/data/data_pipeline/test_dynamic_bucket.py` that calls `dynamic_bucket` with `cost_fn=` as a kwarg. It fails on main with the exact `TypeError` from the issue and passes on this branch.

I didn't rebuild fairseq2n from source locally (CUDA toolchain isn't on this box), so CI will give it its real exercise across the matrix. The change itself is a one character substring swap in an existing `py::arg()` macro call, mechanically equivalent to its neighbors.

Side note: `filter` has the same bug (`predicate` in the stub vs `py::arg("fn")` in the binding, around line 516). Sending that as a separate PR.

### DIFF
--- a/native/python/src/fairseq2n/bindings/data/data_pipeline.cc
+++ b/native/python/src/fairseq2n/bindings/data/data_pipeline.cc
@@ -500,7 +500,7 @@ def_data_pipeline(py::module_ &data_module)
                 return self;
             },
             py::arg("threshold"),
-            py::arg("fn"),
+            py::arg("cost_fn"),
             py::arg("bucket_creation_fn") = std::nullopt,
             py::arg("min_num_examples") = std::nullopt,
             py::arg("max_num_examples") = std::nullopt,
--- a/tests/unit/data/data_pipeline/test_dynamic_bucket.py
+++ b/tests/unit/data/data_pipeline/test_dynamic_bucket.py
@@ -320,3 +320,22 @@ class TestDynamicBucketOp:
 
         with pytest.raises(StopIteration):
             next(iter(pipeline))
+
+    def test_op_works_when_cost_fn_is_passed_as_kwarg(self) -> None:
+        # Regression for #1103: the binding used to expose this kwarg as `fn`.
+        seq = list(range(1, 7))
+
+        pipeline = (
+            read_sequence(seq)
+            .dynamic_bucket(threshold=6, cost_fn=lambda x: x)
+            .and_return()
+        )
+
+        it = iter(pipeline)
+
+        assert next(it) == [1, 2, 3]
+        assert next(it) == [4, 5]
+        assert next(it) == [6]
+
+        with pytest.raises(StopIteration):
+            next(it)


Ace3Z:
Friendly ping on this one. It's a small binding fix for #1103 with a regression test. @cirquit @cbalioglu whenever you have a moment, would appreciate a look.

Ace3Z:
Friendly ping. @cbalioglu, would you have a moment to look at this? Tiny pybind/stub alignment fix. Happy to address any feedback.

PR #1511, opened 2026-05-25
fix(data): align filter() kwarg between Python stub and pybind

Ace3Z:
Follow up to #1510 for the same bug class.

The Python stub at `src/fairseq2/data/data_pipeline.py:275` declares `def filter(self, predicate: ...)`, but the pybind binding at `native/python/src/fairseq2n/bindings/data/data_pipeline.cc:516` exposed the arg as `py::arg("fn")`. So calling `pipeline.filter(predicate=...)` blows up:

```
TypeError: filter(): incompatible function arguments. The following argument types are supported:
    1. (..., fn: Callable[[Any], bool]) -> ...
Invoked with: ...; kwargs: predicate=<lambda>
```

Renamed the pybind kwarg from `fn` to `predicate` to match the documented Python signature. I grep'd every `.filter(` call across `src/`, `tests/`, `examples/`, and `recipes/`: there are 17 callers, all of them pass the predicate positionally, none use `fn=`. So no internal breakage.

Added a regression test in `tests/unit/data/data_pipeline/test_filter.py` that calls `filter` with `predicate=` as a kwarg. Built fairseq2n from source on Linux + Python 3.12 + Torch 2.12 and ran the A/B cycle: test fails against the unpatched binding with the exact `TypeError` above, passes against the patched one. The other two tests in `TestFilterOp` are unchanged and still pass.

As promised in the description of #1510.

### DIFF
--- a/native/python/src/fairseq2n/bindings/data/data_pipeline.cc
+++ b/native/python/src/fairseq2n/bindings/data/data_pipeline.cc
@@ -513,7 +513,7 @@ def_data_pipeline(py::module_ &data_module)
 
                 return self;
             },
-            py::arg("fn"))
+            py::arg("predicate"))
         .def(
             "map",
             [](
--- a/tests/unit/data/data_pipeline/test_filter.py
+++ b/tests/unit/data/data_pipeline/test_filter.py
@@ -39,3 +39,13 @@ class TestFilterOp:
                 pass
 
         assert str(exc_info.value) == "filter error"
+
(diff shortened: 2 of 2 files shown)


Ace3Z:
Friendly ping. This is the sibling of #1510, the same one line kwarg fix but applied to filter(). @cirquit @cbalioglu a quick look when you get a chance would be great.

Ace3Z:
Friendly ping. @cbalioglu, would you have a moment to look at this? Tiny pybind/stub alignment fix. Happy to address any feedback.

PR #1519, opened 2026-06-08
Add Gemma4 12b unified model support in Fairseq2

YunchaoYang:

Fixes #{issue number}

**Check list:**

### DIFF
--- a/src/fairseq2/assets/cards/models/gemma4.yaml
+++ b/src/fairseq2/assets/cards/models/gemma4.yaml
@@ -79,8 +79,8 @@ tokenizer_family: gemma4
 name: gemma4_12b
 model_family: gemma4
 model_arch: 12b
-checkpoint: "file:///checkpoint/smallomnillm/shared/models/gemma-4-12B"
-tokenizer: "file:///checkpoint/smallomnillm/shared/models/gemma-4-12B"
+checkpoint: "file:///checkpoint/fairseq2/shared/models/gemma-4-12B"
+tokenizer: "file:///checkpoint/fairseq2/shared/models/gemma-4-12B"
 tokenizer_family: gemma4
 
 ---
--- a/src/fairseq2/assets/cards/models/gemma4.yaml
+++ b/src/fairseq2/assets/cards/models/gemma4.yaml
@@ -79,8 +79,8 @@ tokenizer_family: gemma4
 name: gemma4_12b
 model_family: gemma4
 model_arch: 12b
-checkpoint: "hg://google/gemma-4-12B"
-tokenizer: "hg://google/gemma-4-12B"
+checkpoint: "file:///checkpoint/smallomnillm/shared/models/gemma-4-12B"
+tokenizer: "file:///checkpoint/smallomnillm/shared/models/gemma-4-12B"
 tokenizer_family: gemma4
 
 ---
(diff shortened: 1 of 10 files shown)


src/fairseq2/assets/cards/models/gemma4.yaml
```diff
@@ -73,3 +73,21 @@ model_arch: e2b_it
 checkpoint: "hg://google/gemma-4-E2B-it"
 tokenizer: "hg://google/gemma-4-E2B-it"
 tokenizer_family: gemma4
+
+---
+
+name: gemma4_12b
+model_family: gemma4
+model_arch: 12b
+checkpoint: "file:///checkpoint/smallomnillm/shared/models/gemma-4-12B"
```
zyaoj: Let's put this under ext instead of exposing in the OS. Use hg://* model cards.

src/fairseq2/models/gemma4/factory.py
```diff
@@ -527,14 +527,23 @@ def create_final_projection(self, embed: Embedding) -> Projection:
...
+        multimodal embedder (see :meth:`create_audio_embedder` and
+        ``Gemma4Model.forward``).
         """
         config = self._config
 
         if config.audio_config is None:
             return None
 
+        # Unified family: no tower; embedder consumes raw waveform frames.
+        if getattr(config.audio_config, "audio_mode", "conformer") == "linear":
```
zyaoj: why `getattr`? is `audio_mode` a proper dataclass or there's a BC requirement for this field?

src/fairseq2/models/gemma4/audio/config.py
```diff
@@ -11,19 +11,38 @@
...
+
+    * ``"linear"`` (used by Gemma 4 Unified family — 12B+): raw waveform frames
+      of ``audio_samples_per_token`` (640) samples each are fed directly through
+      the multimodal embedder (RMSNorm + Linear) — no tower, no mel, no convs.
+      The ``hidden_size``, ``num_hidden_layers``, conv/attention etc. fields
+      are ignored in this mode; only ``output_proj_dims`` (= 640 for unified)
+      and ``rms_norm_eps`` are read.
     """
 
+    audio_mode: str = "conformer"
```
zyaoj: do we want to use `Literal["conformer", "linear"]`?

zyaoj (changes requested):

src/fairseq2/models/gemma4/factory.py
YunchaoYang: Good catch — no BC requirement. Gemma4AudioConfig is a proper @dataclass(kw_only=True) and audio_mode is a typed field with a default, so it's always present. The getattr was leftover defensive code from an earlier iteration where the unified path didn't yet have its own config field. Removed in the latest commit — both call sites in factory.py now use config.audio_config.audio_mode directly, and the field is now typed as Literal["conformer", "linear"] per your other comment.

src/fairseq2/models/gemma4/audio/config.py
YunchaoYang: Yes — done. Switched audio_mode to Literal["conformer", "linear"] in src/fairseq2/models/gemma4/audio/config.py. 

Only those two pipelines are supported (conformer for E*B/31B/26B, linear for the Unified 12B), so the closed set matches reality and gives us static checking + clearer config errors than a free-form str.

src/fairseq2/assets/cards/models/gemma4.yaml
YunchaoYang: Fixed. Thanks for catching that.

zyaoj (approved):
thanks for addressing the comments

